PR #93, opened 2023-04-13
Add Facades.config.set #48

dragoonchang:
As discussed here:
https://github.com/goravel/goravel/issues/48


config/application_test.go
```diff
@@ -38,6 +38,12 @@ func (s *ApplicationTestSuite) TestAdd() {
 	})
 
 	s.Equal("local", s.config.GetString("app.env"))
+
+	s.config.Add("path.with.dot.case1", "value1")
```
hwbrzzl, line 42: Please add a test case like: `s.config.Add("path.with.dot", map[string]any{"case3": "value3"})`

hwbrzzl (commented):
Perfect, just one tip. 👍

config/application_test.go
dragoonchang: Thanks & Done!

hwbrzzl (approved):
Thanks, LGTM

PR #116, opened 2023-05-12
Fix middleware bug

DevHaoZi:
Closes goravel/goravel#123

### DIFF
--- a/route/gin_group.go
+++ b/route/gin_group.go
@@ -5,7 +5,7 @@ import (
 
 	"github.com/gin-gonic/gin"
 
-	httpcontract "github.com/goravel/framework/contracts/http"
+	contractshttp "github.com/goravel/framework/contracts/http"
 	"github.com/goravel/framework/contracts/route"
 )
 
@@ -13,12 +13,12 @@ type GinGroup struct {
 	instance          gin.IRouter
 	originPrefix      string
 	prefix            string
-	originMiddlewares []httpcontract.Middleware
-	middlewares       []httpcontract.Middleware
-	lastMiddlewares   []httpcontract.Middleware
+	originMiddlewares []contractshttp.Middleware
+	middlewares       []contractshttp.Middleware
+	lastMiddlewares   []contractshttp.Middleware
 }
 
-func NewGinGroup(instance gin.IRouter, prefix string, originMiddlewares []httpcontract.Middleware, lastMiddlewares []httpcontract.Middleware) route.Route {
+func NewGinGroup(instance gin.IRouter, prefix string, originMiddlewares []contractshttp.Middleware, lastMiddlewares []contractshttp.Middleware) route.Route {
 	return &GinGroup{
 		instance:          instance,
 		originPrefix:      prefix,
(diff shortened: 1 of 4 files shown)


route/gin_group.go
```diff
@@ -111,7 +111,6 @@ func (r *GinGroup) getGinRoutesWithMiddlewares() gin.IRoutes {
 	middlewares = append(middlewares, ginOriginMiddlewares...)
 	middlewares = append(middlewares, ginMiddlewares...)
 	middlewares = append(middlewares, ginLastMiddlewares...)
-	r.middlewares = []httpcontract.Middleware{}
```
hwbrzzl, line 114: So sorry, just saw this PR! If we remove this line directly, maybe we will get the error below:
```
facades.Route.Middleware(A).Get()
facades.Route.Middleware(B).Post() // will apply middleware A and B at the same time
```

We can add a test case to confirm this.

DevHaoZi: > So sorry, just saw this PR! If we remove this line directly, maybe we will get the error below:
> 
> ```
> facades.Route.Middleware(A).Get()
> facades.Route.Middleware(B).Post() // will apply middleware A and B at the same time
> ```
> 
> We can add a test case to confirm this.

Any suggestions for improvement? I'm not very familiar with Gin's middleware usage.

hwbrzzl: Maybe we can move this line to a new method: `func (r *GinGroup) clearMiddlewares()`, then call the method at the end of methods of calling `r.getGinRoutesWithMiddlewares()`: `Get`, `Post`, etc.
```
func (r *GinGroup) Any(relativePath string, handler httpcontract.HandlerFunc) {
	r.getGinRoutesWithMiddlewares().Any(pathToGinPath(relativePath), []gin.HandlerFunc{handlerToGinHandler(handler)}...)
	r.clearMiddlewares()
}

func (r *GinGroup) Resource(relativePath string, controller httpcontract.ResourceController) {
	r.getGinRoutesWithMiddlewares().GET(pathToGinPath(relativePath), []gin.HandlerFunc{handlerToGinHandler(controller.Index)}...)
	r.getGinRoutesWithMiddlewares().POST(pathToGinPath(relativePath), []gin.HandlerFunc{handlerToGinHandler(controller.Store)}...)
	r.getGinRoutesWithMiddlewares().GET(pathToGinPath(relativePath+"/{id}"), []gin.HandlerFunc{handlerToGinHandler(controller.Show)}...)
	r.getGinRoutesWithMiddlewares().PUT(pathToGinPath(relativePath+"/{id}"), []gin.HandlerFunc{handlerToGinHandler(controller.Update)}...)
	r.getGinRoutesWithMiddlewares().PATCH(pathToGinPath(relativePath+"/{id}"), []gin.HandlerFunc{handlerToGinHandler(controller.Update)}...)
	r.getGinRoutesWithMiddlewares().DELETE(pathToGinPath(relativePath+"/{id}"), []gin.HandlerFunc{handlerToGinHandler(controller.Destroy)}...)
	r.clearMiddlewares()
}
```

DevHaoZi: ok, fixed it.

route/gin_group.go
```diff
@@ -5,20 +5,20 @@ import (
 
 	"github.com/gin-gonic/gin"
 
-	httpcontract "github.com/goravel/framework/contracts/http"
+	contractshttp "github.com/goravel/framework/contracts/http"
```
hwbrzzl, line 8: S: I saw it's httpcontract in Laravel, so I think we don't need to change it, and we can change to this in other places in the future.
image

hwbrzzl: Emm, I think we should have a unified standard on it, so `contractshttp ` is ok, let's do it in other places!

route/gin_group.go
hwbrzzl: Great! Could you please add a test case to cover this situation?
```
facades.Route.Middleware(A).Get()
facades.Route.Middleware(B).Post() // will apply middleware A and B at the same time
```

hwbrzzl (approved):
Great!

hwbrzzl (approved):
Great!

PR #182, opened 2023-06-12
Optimize carbon timezone

hwbrzzl:
## 📑 Description

I found some optimized points:

1. Need use `app.timezone` configuration;
2. Set `carbon` to `support` will be more reasonable;

## ✅ Checks


support/carbon/carbon.go
```diff
@@ -15,7 +14,6 @@ const (
 	Local = carbon.Local // 本地时间
 	UTC   = carbon.UTC   // 世界协调时间
 	GMT   = carbon.GMT   // 格林尼治标准时间
-	CST   = carbon.CST   // 中国标准时间
```
hwbrzzl, line 18: Remove this line, because I found it doesn't work, will throw error.

support/carbon/carbon.go
```diff
@@ -170,237 +168,152 @@ const (
...
-	t := getInstance(timezone...)
-
-	if t.testNow {
-		return t.time
+	if IsTestNow() {
+		return clock.testTime
 	}
 
-	now := carbon.Now(timezone...)
-	t.time = now
```
hwbrzzl, line 223: I think the `time` just store test time is ok, we don't need to set it in every function, so I remove it.

hwbrzzl:
The CI seems to have some errors, very strange, local is passed, maybe it will good tomorrow.

DevHaoZi (approved):
LGTM

PR #218, opened 2023-06-27
fix: auth to return error when user doesn't exists

Marian0:
## 📑 Description

When using `facades.Auth().User(ctx, &user)` in goravel app we found error were not thrown when user doesn't exist.


DevHaoZi (dismissed):
LGTM

Marian0:
> Hi @Marian0 Do you need this fix immediately? If so, this PR need to merge to v1.12.x, and I'll release v1.12.5.

hey @hwbrzzl , I don't need immediately so let me know if you prefer me to re-create this PR from 1.12.x branch or keep it as a master target branch, happy to follow your suggestions, thanks.

hwbrzzl:
@Marian0 Merge master is ok, let's go ahead. 👍

hwbrzzl (approved):
Good catch!

PR #230, opened 2023-07-03
feat: Alias `spew.Dump` and so on

flc1125:
## 📑 Description

Alias `spew.Dump` and so on.

- `debug.Dump()`
- `debug.SDump()`
- `debug.FDump()`

## ✅ Checks

### DIFF
--- a/support/debug/dump.go
+++ b/support/debug/dump.go
@@ -1,0 +1,21 @@
+package debug
+
+import (
+	"github.com/davecgh/go-spew/spew"
+	"io"
+)
+
+// Dump is a wrapper around spew.Dump.
+func Dump(v ...interface{}) {
+	spew.Dump(v...)
+}
+
+// FDump is a wrapper around spew.Fdump.
+func FDump(w io.Writer, v ...interface{}) {
+	spew.Fdump(w, v...)
+}
+
+// SDump is a wrapper around spew.Sdump.
+func SDump(v ...interface{}) string {
+	return spew.Sdump(v...)
+}
--- a/support/debug/dump.go
+++ b/support/debug/dump.go
@@ -1,21 +1,25 @@
 package debug
 
 import (
-	"github.com/davecgh/go-spew/spew"
 	"io"
+
+	"github.com/davecgh/go-spew/spew"
 )
 
-// Dump is a wrapper around spew.Dump.
+// Dump is used to display detailed information about variables
+// And this is a wrapper around spew.Dump.
 func Dump(v ...interface{}) {
 	spew.Dump(v...)
 }
 
-// FDump is a wrapper around spew.Fdump.
+// FDump is used to display detailed information about variables to the specified io.Writer
+// And this is a wrapper around spew.Fdump.
 func FDump(w io.Writer, v ...interface{}) {
 	spew.Fdump(w, v...)
 }
 
-// SDump is a wrapper around spew.Sdump.
+// SDump is used to display detailed information about variables as a string,
+// And this is a wrapper around spew.Sdump.
 func SDump(v ...interface{}) string {
 	return spew.Sdump(v...)
 }
--- a/support/debug/dump_test.go
+++ b/support/debug/dump_test.go
@@ -1,20 +1,46 @@
 package debug
 
 import (
-	"fmt"
-	"github.com/stretchr/testify/assert"
+	"bytes"
 	"os"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
 )
 
+func redirectStdout(fn func()) ([]byte, error) {
+	f, err := os.CreateTemp("", "stdout")
+	if err != nil {
+		return nil, err
+	}
+	defer os.Remove(f.Name())
+	defer f.Close()
+
+	orig := os.Stdout
+	os.Stdout = f
+	fn()
+	os.Stdout = orig
+
+	return os.ReadFile(f.Name())
+}
+
 func TestDump(t *testing.T) {
-	Dump("foo")
+	buf, err := redirectStdout(func() {
+		Dump("foo")
+	})
+	assert.NoError(t, err)
+	assert.Equal(t, `(string) (len=3) "foo"
+`, string(buf))
 }
 
 func TestFDump(t *testing.T) {
-	FDump(os.Stdout, "foo")
+	var buf bytes.Buffer
+	w := &buf
 
-	fmt.Sprintf("%s", SDump("foo"))
+	FDump(w, "foo")
+
+	assert.Equal(t, `(string) (len=3) "foo"
+`, buf.String())
 }
 
 func TestSDump(t *testing.T) {
--- a/support/debug/dump_test.go
+++ b/support/debug/dump_test.go
@@ -1,0 +1,23 @@
+package debug
+
+import (
+	"fmt"
+	"github.com/stretchr/testify/assert"
+	"os"
+	"testing"
+)
+
+func TestDump(t *testing.T) {
+	Dump("foo")
+}
+
+func TestFDump(t *testing.T) {
+	FDump(os.Stdout, "foo")
+
+	fmt.Sprintf("%s", SDump("foo"))
+}
+
+func TestSDump(t *testing.T) {
+	assert.Equal(t, `(string) (len=3) "foo"
+`, SDump("foo"))
+}


DevHaoZi:
I checked the package `github.com/davecgh/go-spew/spew`, I don't think its implementation is complicated, so I suggest implementing your own instead of importing a 3rd party package.

flc1125:
I think we can make a set of tools about Debug, so that when using the framework, it can be very convenient to use it.

This PR, however, is only part of a collection of debug tools. It already has `spew`, but we're just referencing it and putting it together.

While we can develop a set of features that are the same as `spew`, why repeat the wheel when we have open source?

These are my thoughts and I'd like to hear yours.

DevHaoZi:
> I think we can make a set of tools about Debug, so that when using the framework, it can be very convenient to use it.
> 
> This PR, however, is only part of a collection of debug tools. It already has `spew`, but we're just referencing it and putting it together.
> 
> While we can develop a set of features that are the same as `spew`, why repeat the wheel when we have open source?
> 
> These are my thoughts and I'd like to hear yours.

1. This package lacks maintenance, last updated in 2018.
2. We are trying to minimize the use of unnecessary third-party packages. Refer to Laravel's github.com/illuminate repository, which uses very few third-party packages.

support/debug/dump_test.go
hwbrzzl, line 1: S: I think the unit tests are unnecessary, they're so simple.

support/debug/dump.go
```diff
@@ -0,0 +1,21 @@
+package debug
+
+import (
+	"github.com/davecgh/go-spew/spew"
+	"io"
+)
+
+// Dump is a wrapper around spew.Dump.
```
hwbrzzl, line 8: S: Add the real comment here.

support/debug/dump.go
```diff
@@ -0,0 +1,21 @@
...
+func Dump(v ...interface{}) {
+	spew.Dump(v...)
+}
+
+// FDump is a wrapper around spew.Fdump.
```
hwbrzzl, line 13: Ditto

support/debug/dump.go
```diff
@@ -0,0 +1,21 @@
...
+func FDump(w io.Writer, v ...interface{}) {
+	spew.Fdump(w, v...)
+}
+
+// SDump is a wrapper around spew.Sdump.
```
hwbrzzl, line 18: Ditto

hwbrzzl (commented):
Good feature! The meaning of @DevHaoZi is want to reduce dependencies, that's a good idea, but I think we can use the third package in this PR directly, because there is some complex logic and the package is small.

support/debug/dump.go
flc1125: Optimized

support/debug/dump.go
flc1125: Optimized

support/debug/dump_test.go
flc1125: Optimized

hwbrzzl:
> > I think we can make a set of tools about Debug, so that when using the framework, it can be very convenient to use it.
> > This PR, however, is only part of a collection of debug tools. It already has `spew`, but we're just referencing it and putting it together.
> > While we can develop a set of features that are the same as `spew`, why repeat the wheel when we have open source?
> > These are my thoughts and I'd like to hear yours.
> 
> 1. This package lacks maintenance, last updated in 2018.
> 2. We are trying to minimize the use of unnecessary third-party packages. Refer to Laravel's `github.com/illuminate` repository, which uses very few third-party packages.

Sorry, I didn't notice that this package has not been maintained for a long time, this's indeed a problem. Can we find another package instead? @flc1125

flc1125:
> > > I think we can make a set of tools about Debug, so that when using the framework, it can be very convenient to use it.
> > > This PR, however, is only part of a collection of debug tools. It already has `spew`, but we're just referencing it and putting it together.
> > > While we can develop a set of features that are the same as `spew`, why repeat the wheel when we have open source?
> > > These are my thoughts and I'd like to hear yours.
> > 
> > 
> > 
> > 1. This package lacks maintenance, last updated in 2018.
> > 2. We are trying to minimize the use of unnecessary third-party packages. Refer to Laravel's `github.com/illuminate` repository, which uses very few third-party packages.
> 
> Sorry, I didn't notice that this package has not been maintained for a long time, this's indeed a problem. Can we find another package instead? @flc1125

I can try to find out.

As far as I know, the spew package is used in many projects. The framework itself also relies on the spew package.

DevHaoZi:
> > > > I think we can make a set of tools about Debug, so that when using the framework, it can be very convenient to use it.
> > > > This PR, however, is only part of a collection of debug tools. It already has `spew`, but we're just referencing it and putting it together.
> > > > While we can develop a set of features that are the same as `spew`, why repeat the wheel when we have open source?
> > > > These are my thoughts and I'd like to hear yours.
> > > 
> > > 
> > > 
> > > 1. This package lacks maintenance, last updated in 2018.
> > > 2. We are trying to minimize the use of unnecessary third-party packages. Refer to Laravel's `github.com/illuminate` repository, which uses very few third-party packages.
> > 
> > 
> > Sorry, I didn't notice that this package has not been maintained for a long time, this's indeed a problem. Can we find another package instead? @flc1125
> 
> I can try to find out.
> 
> As far as I know, the spew package is used in many projects. The framework itself also relies on the spew package.

No, this framework doesn't relies spew package, maybe a 3rd package import it. That why we need to minimize the use of unnecessary 3rd packages.

hwbrzzl:
@flc1125 Oh, good catch, it's true, `github.com/stretchr/testify/mock` is using it. Okay, that being the case, let's go ahead! What do you think about this, please?

hwbrzzl:
@flc1125 Oh, good catch, it's true, github.com/stretchr/testify/mock is using it, we can't remove the `testify` package. Okay, that being the case, I'm ok with this! What do you think about this, please? @DevHaoZi

DevHaoZi:
> 

OK, no problem.

support/debug/dump_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"bytes"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func redirectStdout(fn func()) ([]byte, error) {
```
hwbrzzl, line 11: Good!

flc1125:
My original intention was to facilitate debugging, and spew is a good and relatively mature third-party package in this area.

My proposal can be considered as follows:

- Does our project need to explore the debug area in the near future?
- Is our project looking to reduce reliance on third party packages in the near future?

I think we can choose from the above direction.

----

- If the final conclusion is to use, then I am deeply honored. Even in future planning, remove this code.
- If the final conclusion is not to use, I think it is also a good decision. 
	
	> The PR also only defines aliases, and if the user wishes to introduce spew, the option of this package can be given to the user.

In your way, choose your correct answer.

---

As I was writing the above, I was honored to find that you had merged. Thank.

@hwbrzzl @DevHaoZi 😁

hwbrzzl:
It's good for goravel to have the feature, thanks for your contribution, hope you can become a core contributor in the future. 👍

DevHaoZi:
> My original intention was to facilitate debugging, and spew is a good and relatively mature third-party package in this area.
> 
> My proposal can be considered as follows:
> 
> * Does our project need to explore the debug area in the near future?
> * Is our project looking to reduce reliance on third party packages in the near future?
> 
> I think we can choose from the above direction.
> 
> * If the final conclusion is to use, then I am deeply honored. Even in future planning, remove this code.
> * If the final conclusion is not to use, I think it is also a good decision.
>   > The PR also only defines aliases, and if the user wishes to introduce spew, the option of this package can be given to the user.
> 
> In your way, choose your correct answer.
> 
> As I was writing the above, I was honored to find that you had merged. Thank.
> 
> @hwbrzzl @DevHaoZi 😁

Yes, we are working on minimize the use of unnecessary 3rd packages. Before the next contribution, you can discuss it on github.com/goravel/goravel/issues, which can reduce unnecessary conflicts and troubles.😁

flc1125:
> > My original intention was to facilitate debugging, and spew is a good and relatively mature third-party package in this area.
> > My proposal can be considered as follows:
> > 
> > * Does our project need to explore the debug area in the near future?
> > * Is our project looking to reduce reliance on third party packages in the near future?
> > 
> > I think we can choose from the above direction.
> > 
> > * If the final conclusion is to use, then I am deeply honored. Even in future planning, remove this code.
> > * If the final conclusion is not to use, I think it is also a good decision.
> >   > The PR also only defines aliases, and if the user wishes to introduce spew, the option of this package can be given to the user.
> > 
> > In your way, choose your correct answer.
> > As I was writing the above, I was honored to find that you had merged. Thank.
> > @hwbrzzl @DevHaoZi 😁
> 
> Yes, we are working on minimize the use of unnecessary 3rd packages. Before the next contribution, you can discuss it on github.com/goravel/goravel/issues first, which can reduce unnecessary conflicts and troubles.😁

Very good advice.

PR #232, opened 2023-07-04
wip: feat: add gzip and brotli middleware

mauri870:
https://github.com/goravel/goravel/issues/208

Closes https://github.com/goravel/goravel/issues/208

## ✅ Checks

## ℹ Additional Information

What is missing:

- [ ] Update docs, add gzip and brotli to Routing
- [ ] Create config files for gzip and brotli

Thx @Marian0 for the help getting gzip compression to work ✌️


mauri870:
@hwbrzzl One thing to take into consideration is that the brotli middleware I choose does not handle fallback to gzip.  We can't also chain `.GlobalMiddleware(middlewares.Brotli(), middlewares.Gzip())` since we'll end up with a brotli compressed gzip payload (or a gzip compressed brotli payload xD).

My take on it is that we should aim for a global Compress middleware with a respective compress.go config file where you can adjust the options and also enable brotli, gzip, deflate or maybe all of them. That would make things more easier than maintaining multiple middlewares. Please let me know what you think about it.

hwbrzzl:
> @hwbrzzl One thing to take into consideration is that the brotli middleware I choose does not handle fallback to gzip. We can't also chain `.GlobalMiddleware(middlewares.Brotli(), middlewares.Gzip())` since we'll end up with a brotli compressed gzip payload (or a gzip compressed brotli payload xD).
> 
> My take on it is that we should aim for a global Compress middleware with a respective compress.go config file where you can adjust the options and also enable brotli, gzip, deflate or maybe all of them at once. That would make things more easier than maintaining multiple middlewares. Please let me know what you think about it.

Yes, it's better to add a configuration to control it. There's a problem, we want to implement multiple HTTP drivers: [goravel/gin](https://github.com/goravel/gin), [goravel/filber](https://github.com/goravel/fiber), they are developing, and we have another PR to refactor HTTP and Route modules, so can we make these two middleware in this PR more generic? Change `gin` to `net/http`, we will also try to change the CORS middleware.

http/middleware/brotli.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+func Brotli() httpcontract.Middleware {
+	return func(ctx httpcontract.Context) {
+		switch ctx := ctx.(type) {
+		case *http.GinContext:
+			newBrotli(gbrotli.Options{
+				WriterOptions: brotli.WriterOptions{
+					Quality: facades.Config().GetInt("brotli.compression_level"),
+					LGWin:   facades.Config().GetInt("brotli.lgwin_level"),
```
hwbrzzl, line 20: S: We need to use `http.ConfigFacade` instead of `facades.Config()`.

http/middleware/gzip.go
```diff
@@ -0,0 +1,37 @@
...
+	httpcontract "github.com/goravel/framework/contracts/http"
+	"github.com/goravel/framework/facades"
+	"github.com/goravel/framework/http"
+)
+
+func Gzip() httpcontract.Middleware {
+	return func(ctx httpcontract.Context) {
+		switch ctx := ctx.(type) {
+		case *http.GinContext:
+			newGzip(facades.Config().GetInt("gzip.compression_level"))(ctx.Instance())
```
hwbrzzl, line 16: Ditto

mauri870:
Since we are aiming to support different http drivers it makes sense to try to make the compression midlewares generic. I will work on it 🙏

hwbrzzl (commented):
Great PR! I wonder if we can add some unit tests for this PR. If it's hard to add it, please add some screenshots that test this feature locally, and please list the new configurations added in this PR here.

Marian0:
@mauri870 great job! Thanks for creating this PR - I tested it locally on a sample project and both compressions are working as expected. Thanks again 🙌

http/middleware/brotli.go
Marian0: Just wondering why should we use http.ConfigFacade, is it about performance reasons ?

hwbrzzl: `http.ConfigFacade` is implemented by the container, there is an import circle error if we use `facades.Config()` directly.

http/middleware/compression.go
```diff
@@ -0,0 +1,55 @@
+package middleware
+
+import (
+	"strings"
+
+	gbrotli "github.com/anargu/gin-brotli"
+	"github.com/andybalholm/brotli"
+	"github.com/gin-contrib/gzip"
+	"github.com/gin-gonic/gin"
```
hwbrzzl, line 9: Q: Can we remove the gin dependency totally?

mauri870: Yes it should be possible I think. I found  [this package](https://github.com/nytimes/gziphandler) but it requires an http.Handler as input. Any clue how to use this here instead of the gin one?

hwbrzzl: We may need to read the source code to understand how to implement Gzip.

By the way, I missed one thing, we can use Gzip by nginx, is it necessary to implement Gzip through the framework? Most time, we need to use Nginx to make a request forwarding. cc @Marian0

mauri870: My 2 cents is that it is something worth it to have in the framework. There was a time until ~2016 that it was preferred to always expose go services being a proxy, (such as nginx),. But then companies started to expose their go services directly and I believe not it is actually safe to do that. Ofc you do not get the benefits of a true proxy but compression is a nice to have for such cases.

mauri870:
@hwbrzzl I was unable to get this working without relying on gin. I think we should wait for the http refactoring before moving forward with this task.

hwbrzzl:
Hi @mauri870 Sorry for the late, I have a travel last week. Sure, make sense, let's wait a bit until the HTTP refactor is finished.

PR #237, opened 2023-07-08
feat: key make support custom env path

DevHaoZi:
## 📑 Description
key make command support custom env path

## ✅ Checks

### DIFF
--- a/foundation/application.go
+++ b/foundation/application.go
@@ -181,11 +181,12 @@ func setEnv() {
 	}
 
 	var env *string
-	if !flag.Parsed() {
+	if !flag.Parsed() && support.Env != support.EnvTest {
 		env = flag.String("env", ".env", "custom .env path")
 		flag.Parse()
 	} else {
-		env = flag.Lookup("env").Value.(flag.Getter).Get().(*string)
+		testEnv := ".env"
+		env = &testEnv
 	}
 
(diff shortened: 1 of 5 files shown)


foundation/application.go
```diff
@@ -178,6 +179,10 @@ func setEnv() {
 			support.Env = support.EnvArtisan
 		}
 	}
+
+	env := flag.String("env", ".env", "custom .env path")
+	flag.Parse()
+	support.EnvPath = *env
```
hwbrzzl, line 185: CI failed, we may need to add the code below here:
```
var env *string
if support.Env == support.EnvTest {
  testEnv := ".env"
  env = &testEnv
} else {
  env = flag.String("env", ".env", "custom .env path")
  flag.Parse()
}
```

DevHaoZi: > CI failed, we may need to add the code below here:
> 
> ```
> var env *string
> if support.Env == support.EnvTest {
>   testEnv := ".env"
>   env = &testEnv
> } else {
>   env = flag.String("env", ".env", "custom .env path")
>   flag.Parse()
> }
> ```

The `flag.Parse()` of the flag package only allow calls once in the application. It seems that there are also calls in `testify`, and repeated calls make it impossible to parse.

hwbrzzl (approved):
Great

PR #266, opened 2023-08-18
feat: delete go 1.18 and 1.19

DevHaoZi:
Closes # 

## 📑 Description

Go 1.18 and 1.19 is EOL, see [https://endoflife.date/go](https://endoflife.date/go). We can delete them.

## ✅ Checks

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,8 +9,6 @@ jobs:
     strategy:
       matrix:
         go:
-          - "1.18"
-          - "1.19"
           - "1.20"
           - "1.21"
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


hwbrzzl (changes requested):
Good catch! Do we have a particular reason to remove 1.18 and 1.19? Such as some packages we need to use don't support 1.18 anymore.

devhaozi:
> Good catch! Do we have a particular reason to remove 1.18 and 1.19? Such as some packages we need to use don't support 1.18 anymore.

Some main packages such as gin or fiber use `1.20` in their `go.sum`

.github/workflows/test.yml
hwbrzzl, line 1: Based on the official guide as shown below, we are safe to change the go.mod file until 1.21, so I think we can keep the tests on 1.18 and 1.19, because there are so many people using these two versions.

image

devhaozi: > 

OK, I will open another PR.

PR #278, opened 2023-08-29
chore(route): add GetInstance method on gin

dihak:
## 📑 Description
I need to GetInstance method for get the gin.Engine.

Because I want to using goravel in lambda serverless.

Thankyou

https://github.com/awslabs/aws-lambda-go-api-proxy

## ✅ Checks

- [x] I have updated the documentation as required

### DIFF
--- a/route/gin.go
+++ b/route/gin.go
@@ -20,6 +20,10 @@ type Gin struct {
 	instance *gin.Engine
 }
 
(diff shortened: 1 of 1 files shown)


devhaozi:
We may not be able to support this feature, because in Goravel's design, the implementation of any module should not be associated with a specific package.

PR #267, opened 2023-08-18
feat: update go 1.20

devhaozi:
Closes # 

## ✅ Checks

- [x] My pull request adheres to the code style of this project
- [ ] My code requires changes to the documentation
- [ ] I have updated the documentation as required
- [x] All the tests have passed

PR #306, opened 2023-10-02
fix: #267

devhaozi:
Closes goravel/goravel#267

## 📑 Description

Fix goravel/goravel#267 sonic warning in arm environment.

## ✅ Checks

### DIFF
--- a/crypt/aes.go
+++ b/crypt/aes.go
@@ -5,7 +5,6 @@ import (
 	"crypto/cipher"
 	"crypto/rand"
 	"encoding/base64"
-	"encoding/json"
 	"errors"
 	"io"
 
@@ -14,7 +13,7 @@ import (
 
 	"github.com/goravel/framework/contracts/config"
 	"github.com/goravel/framework/support"
-	"github.com/goravel/framework/support/env"
+	"github.com/goravel/framework/support/json"
 )
 
 type AES struct {
@@ -63,17 +62,10 @@ func (b *AES) EncryptString(value string) (string, error) {
 	ciphertext := aesgcm.Seal(nil, iv, plaintext, nil)
 
 	var jsonEncoded []byte
-	if env.IsX86() {
-		jsonEncoded, err = sonic.Marshal(map[string][]byte{
-			"iv":    iv,
-			"value": ciphertext,
-		})
-	} else {
-		jsonEncoded, err = json.Marshal(map[string][]byte{
-			"iv":    iv,
-			"value": ciphertext,
-		})
-	}
+	jsonEncoded, err = json.Marshal(map[string][]byte{
+		"iv":    iv,
+		"value": ciphertext,
+	})
 
 	if err != nil {
 		return "", err
--- a/crypt/aes.go
+++ b/crypt/aes.go
@@ -5,6 +5,7 @@ import (
 	"crypto/cipher"
 	"crypto/rand"
 	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"io"
 
@@ -13,6 +14,7 @@ import (
 
 	"github.com/goravel/framework/contracts/config"
 	"github.com/goravel/framework/support"
+	"github.com/goravel/framework/support/environment"
 )
 
 type AES struct {
@@ -60,10 +62,19 @@ func (b *AES) EncryptString(value string) (string, error) {
 
 	ciphertext := aesgcm.Seal(nil, iv, plaintext, nil)
 
-	jsonEncoded, err := sonic.Marshal(map[string][]byte{
-		"iv":    iv,
-		"value": ciphertext,
-	})
+	var jsonEncoded []byte
+	if environment.IsX86() {
+		jsonEncoded, err = sonic.Marshal(map[string][]byte{
+			"iv":    iv,
+			"value": ciphertext,
+		})
+	} else {
+		jsonEncoded, err = json.Marshal(map[string][]byte{
+			"iv":    iv,
+			"value": ciphertext,
+		})
+	}
+
 	if err != nil {
 		return "", err
 	}
--- a/crypt/aes.go
+++ b/crypt/aes.go
@@ -8,7 +8,6 @@ import (
 	"errors"
 	"io"
 
-	"github.com/bytedance/sonic"
 	"github.com/gookit/color"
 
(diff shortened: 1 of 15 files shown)


support/carbon/json_test.go
```diff
@@ -1,10 +1,10 @@
 package carbon
 
 import (
+	"encoding/json"
 	"fmt"
 	"testing"
 
-	"github.com/bytedance/sonic"
 	"github.com/stretchr/testify/assert"
```
devhaozi, line 8: Seems test file not need import sonic for performance.

support/environment/environment.go
```diff
@@ -0,0 +1,27 @@
...
+	return runtime.GOARCH == "arm" || runtime.GOARCH == "arm64"
+}
+
+func IsX86() bool {
+	return runtime.GOARCH == "386" || runtime.GOARCH == "amd64"
+}
+
+func Is64Bit() bool {
+	return runtime.GOARCH == "amd64" || runtime.GOARCH == "arm64"
+}
```
devhaozi, line 27: Add some helper methods for judge environment.

support/environment/environment.go
hwbrzzl, line 1: Great, just one tip, we can optimize the folder name to `env`.

log/formatter/general.go
```diff
@@ -53,44 +56,38 @@ func formatData(data logrus.Fields) (string, error) {
...
 	if len(data) > 0 {
-		dataBytes, err := sonic.Marshal(data)
-		if err != nil {
-			return "", err
-		}
-
 		removedData := deleteKey(data, "root")
 		if len(removedData) > 0 {
-			removedDataBytes, err := sonic.Marshal(removedData)
+			removedDataBytes, err := json.Marshal(removedData)
```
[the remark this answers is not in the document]
devhaozi: > Q: Why do not use `sonic` here?

I forgot it, fixed.

filesystem/local_test.go
```diff
@@ -70,28 +70,28 @@ func (s *LocalTestSuite) TestAllDirectories() {
 	s.True(s.local.Exists("AllDirectories/3/5/6/6.txt"))
 	files, err := s.local.AllDirectories("AllDirectories")
 	s.Nil(err)
-	if runtime.GOOS == "windows" {
+	if env.IsWindows() {
```
hwbrzzl, line 73: Hi @kkumar-gcc You can use this method in your PR.

kkumar-gcc: Okay

hwbrzzl (commented):
Hi @devhaozi I can't reproduce the issue locally, can you reproduce it, please? And tested it?

devhaozi:
> Hi @devhaozi I can't reproduce the issue locally, can you reproduce it, please? And tested it?

What issue?

hwbrzzl:
The warning: `WARNING: sonic onlu supportsGo116~1.20 88 CPU amd64, but your environment is not suitable`

devhaozi:
> The warning: `WARNING: sonic onlu supportsGo116~1.20 88 CPU amd64, but your environment is not suitable`

image
Can only be seen by non-amd64 devices

hwbrzzl:
So, the warning needs to change to go1.16 ~1.22, now is 1.20~

devhaozi:
> So, the warning needs to change to go1.16 ~1.22, now is 1.20~

image
Answer from GPT, don't know why sonic designed this.

support/json/json.go
```diff
@@ -0,0 +1,49 @@
+package json
+
+import (
+	"encoding/json"
+
+	"github.com/bytedance/sonic"
```
hwbrzzl, line 6: So sorry, I missed one thing: the warning is written in the `init` method, the helper method can't remove the warning, because the sonic package will still be complied If it's imported.

hwbrzzl: I pulled the sonic package and tested this situation locally. It seems that we need to remove it totally if we want to remove the warning message.

devhaozi: > So sorry, I missed one thing: the warning is written in the `init` method, the helper method can't remove the warning, because the sonic package will still be complied If it's imported.

In this case, maybe we need to write two files `json_x86.go` and `json.go`, and then solve it through build tag(similar to sonic).

hwbrzzl: Oh, that's a great idea!

hwbrzzl: Em, but one question, we need to change the build tag following `bytedance/sonic` in the future.

devhaozi: > Em, but one question, we need to change the build tag following `bytedance/sonic` in the future.

I think only need use this, not need to add version:
```
// +build !amd64
// +build amd64
```

devhaozi: > Yes, I think so. One more question, I checked our use cases, and we don't have a large data that using `sonic`, based on the benchmark, `sonic` may not bring us a significant improvement, it may be most used in the business logic, what do you think about this, please? Anyway, I think adding the `json_x86.go` file is also a good way.
> 
> image

In logging, sonic may not have an obvious effect, but in AES encryption, if the user loops through large amounts of data, the effect will be seen.

In addition, I think it is also a good idea to provide json helper functions for users to choose.

hwbrzzl: Okay, let's add the `json_x86.go` file first, maybe just a few users want to choose it, so I think we don't need to add the helper function for now.

devhaozi: > Okay, let's add the `json_x86.go` file first, maybe just a few users want to choose it, so I think we don't need to add the helper function for now.

Completed, I will optimize fiber driver json usage in another PR.

crypt/aes.go
hwbrzzl, line 1: `github.com/bytedance/sonic` is still imported.

devhaozi: > `github.com/bytedance/sonic` is still imported.

fixed

kkumar-gcc:
@devhaozi  Can you use `/support/env.go` in `/support/str_test.go`.

image

devhaozi:
> @devhaozi Can you use `/support/env/env.go` in `/support/str/str_test.go`.
> 
> image

ok

hwbrzzl (approved):
Great

PR #314, opened 2023-10-07
feat: refactor queue module

devhaozi:
Closes goravel/goravel#153
Closes goravel/goravel#278

## 📑 Description

This is a draft, need more discussion.

I tried to reconstruct the queue module with reference to Laravel's implementation. There are some issues that need to be discussed and determined.

## ✅ Checks

## Summary by CodeRabbit

- **New Features**
  - Introduced interface `Driver` for managing job queues with methods for pushing, popping, and scheduling jobs.
  - Added `MigrateMakeCommand` struct for creating migrations for failed queue jobs.
  - Implemented `Shutdown` method for the `Worker` to allow graceful shutdown.
  - Enhanced queue configuration with new methods like `Via()` and `FailedJobsQuery()`.

- **Bug Fixes**
  - Improved error handling in multiple methods across the application, including `Register` and `Boot`.
  - Standardized import statements and references for consistency.

- **Refactor**
  - Updated structs and method signatures across various files to use more generic types and new field names.
  - Replaced `machinery` with `driver` in `Worker` and `Task` structs for better integration.
  - Modified SQL statements for database consistency.

- **Tests**
  - Adjusted test cases to reflect new configurations and updated struct fields.
  - Simplified argument passing mechanisms within the test functions.

- **Documentation**
  - Updated references in import statements for clarity and consistency.
  - Revised comments to reflect new method signatures and struct fields.


contracts/queue/driver.go
```diff
@@ -0,0 +1,22 @@
...
+	Pop() (Job, error)
+	// Delete removes a job from the queue.
+	Delete(job Job) error
+	// Release releases a reserved job back onto the queue.
+	Release(job Job, delay int) error
+	// Clear clears all pending jobs in the queue.
+	Clear() error
+	// Size returns the size of the queue.
+	Size() (int, error)
+}
```
devhaozi, line 22: These interfaces come from https://github.com/illuminate/queue/blob/master/DatabaseQueue.php and https://github.com/illuminate/queue/blob/master/RedisQueue.php. In golang, we may also Add some other interfaces to help deliver tasks to driver.

contracts/queue/task.go
```diff
@@ -1,16 +1,16 @@
...
 
 type Task interface {
 	// Dispatch dispatches the task.
 	Dispatch() error
 	// DispatchSync dispatches the task synchronously.
 	DispatchSync() error
-	// Delay dispatches the task after the given delay.
-	Delay(time time.Time) Task
+	// Delay dispatches the task after the given time.
+	Delay(time carbon.Carbon) Task
```
devhaozi, line 13: Adjusted to use Carbon objects, because Laravel does this too, see https://learnku.com/docs/laravel/10.x/queues/14873#38a860

queue/config.go
```diff
@@ -59,3 +61,7 @@ func (r *Config) Redis(queueConnection string) (dsn string, database int, queue
 
 	return
 }
+
+func (r *Config) Database(queueConnection string) orm.Orm {
+	return facades.Orm().Connection(queueConnection)
+}
```
devhaozi, line 67: Trying to get the database connection directly through Orm, not sure if it works?

queue/driver/database.go
```diff
@@ -0,0 +1,54 @@
...
+	return nil
+}
+
+func (d *Database) Later(job queue.Job, delay int) error {
+	return nil
+}
+
+func (d *Database) Pop() (queue.Job, error) {
+	return nil, nil
+}
```
devhaozi, line 38: Need to study how to run the pop-up tasks. Consider adding a `go func` scheduled listener?
Laravel use `queue:work` artisan command.

queue/migrations/00000000000000_create_jobs_table.up.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE jobs
+(
+    id           BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+    queue        TEXT      NOT NULL,
+    payload      JSONB     NOT NULL,
+    attempts     INTEGER   NOT NULL DEFAULT 0,
+    reserved_at  TIMESTAMP NOT NULL DEFAULT NULL,
+    available_at TIMESTAMP NOT NULL,
+    created_at   TIMESTAMP NOT NULL
+);
```
devhaozi, line 10: The migration we currently use does not support multiple databases well. This migration was written for PostgreSQL.
Adding 4 different migration files seems a bit silly.

queue/migrations/00000000000001_create_failed_jobs_table.up.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE failed_jobs
+(
+    id         BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+    uuid       UUID      NOT NULL UNIQUE,
+    connection TEXT      NOT NULL,
+    queue      TEXT      NOT NULL,
+    payload    JSONB     NOT NULL,
+    exception  TEXT      NOT NULL,
+    failed_at  TIMESTAMP NOT NULL
+)
```
devhaozi, line 10: Ditto too

devhaozi:
cc @hwbrzzl

queue/job.go
```diff
@@ -0,0 +1,40 @@
...
+func Call(signature string, args ...any) error {
+	mutex.RLock()
+	defer mutex.RUnlock()
+
+	job, exists := JobRegistry[signature]
+	if !exists {
+		return errors.New("job not found")
+	}
+	return job.Handle(args...)
+}
```
devhaozi, line 40: Re-implemented job registration.

contracts/queue/task.go
hwbrzzl: Good idea.

queue/migrations/00000000000000_create_jobs_table.up.sql
hwbrzzl: But it seems we have to do it for now.

queue/models/Job.go
```diff
@@ -0,0 +1,37 @@
...
+	MaxExceptions *uint                `gorm:"type:bigint;default:null;default:0"`
+	Exception     *string              `gorm:"type:text;default:null"`
+	Backoff       uint                 `gorm:"type:bigint;not null;default:0"`     // A number of seconds to wait before retrying the job.
+	Timeout       *uint                `gorm:"type:bigint;default:null;default:0"` // The number of seconds the job can run.
+	TimeoutAt     *carbon.DateTime     `gorm:"column:timeout_at"`
+	ReservedAt    *carbon.DateTime     `gorm:"column:reserved_at"`
+	AvailableAt   *carbon.DateTime     `gorm:"column:available_at"`
+	CreatedAt     *carbon.DateTime     `gorm:"column:created_at"`
+	FailedAt      *carbon.DateTime     `gorm:"column:failed_at"`
+}
```
devhaozi, line 25: Made some adjustments to merge the failed_job table into the job table.

queue/config.go
devhaozi: > Trying to get the database connection directly through Orm, not sure if it works?

Replace with `OrmFacade`.

queue/driver_database.go
```diff
@@ -0,0 +1,135 @@
...
+
+				err = Call(job.Signature(), args)
+				if err != nil {
+					continue
+				}
+
+				time.Sleep(time.Second)
+			}
+		}()
+	}
```
devhaozi, line 134: Currently it is just a simple implementation and more processing is needed (error retry, timeout cancellation, etc.)

devhaozi:
@hwbrzzl Please help me check if there are any problems with the implementation idea. If there are no problems, I will continue to complete other driver development.

queue/driver_database.go
hwbrzzl, line 1: Is it better to create a new package for the `database` and `redis` driver?

queue/machinery.go
hwbrzzl, line 1: After this feature, we don't need the `machinery` anymore, right?

queue/migrations/00000000000000_create_jobs_table.up.sql
hwbrzzl, line 1: It's better to save a date from the name instead of `00000`

queue/driver.go
```diff
@@ -0,0 +1,25 @@
+package queue
+
+import (
+	"github.com/goravel/framework/contracts/queue"
+)
+
+const DriverSync string = "sync"
+const DriverASync string = "async"
```
hwbrzzl, line 8: The `async` driver means to save the queue in memory, right?

hwbrzzl (commented):
Great 👍

queue/machinery.go
devhaozi: > After this feature, we don't need the `machinery` anymore, right?

Yes

queue/driver_database.go
devhaozi: > Is it better to create a new package for the `database` and `redis` driver?

I’m not sure what to use for the package name. Redis is already occupied by the Redis driver. Consider queue_redis? queue_driver_redis?

devhaozi: > Yes, goravel/queue-db, goravel/queue-redis, but if there are no new packages that need to be added, we may not need to create it.

OK, then keep the status quo for now. If someone consider introducing Kafka, RabbitMQ, etc. in the future, then adjust it to a package.

queue/driver.go
devhaozi: > > The `async` driver means to save the queue in memory, right?
> 
> Consider using `go func` directly to start a goroutine running job.

It seems that using go func cannot return the job error because it is in a goroutine. Maybe we should return to your suggestion.

queue/driver_database.go
devhaozi, line 134: Q: If the `Call` a job return a error, how to handle this error?

cc @hwbrzzl

hwbrzzl: The `Server` method is called in what scene? Can you add an example here?

devhaozi: > The `Server` method is called in what scene? Can you add an example here?

Start a queue server to listen job, like laravel's queue:work artisan.

SwenVanZanten:
Thanks for this, was trying to implement AWS SQS into a potential replacement product with goravel. And found that adding my own driver wasn't an option. Anyway keep it up! 💪

queue/driver_database.go
hwbrzzl: Write into the failed table?

devhaozi: Not need now, removed.

devhaozi:
This PR is basically completed, please let me know if there are any areas that need improvement.

Known issues:
1. Database driver lacks test cases.
2. Features related to task timeout have not been implemented yet.

cc @hwbrzzl

hwbrzzl:
@devhaozi Great job man, do we cover all current features?

devhaozi:
> @devhaozi Great job man, do we cover all current features?

Yes

devhaozi:
> > @devhaozi Great job man, do we cover all current features?
> 
> Yes

Laravel support this https://learnku.com/docs/laravel/10.x/queues/14873#8dd0bc and https://learnku.com/docs/laravel/10.x/queues/14873#6a8bef.

Currently, I have reserved relevant interfaces and methods, but they have not been fully implemented yet.

hwbrzzl:
Got it, will take a look.

queue/driver_redis.go
```diff
@@ -0,0 +1,189 @@
...
+		Member: payload,
+	}).Err()
+}
+
+func (r *Redis) Pop(queue string) (contractsqueue.Job, []contractsqueue.Arg, error) {
+	if err := r.migrateDelayedJobs(queue); err != nil {
+		return nil, nil, err
+	}
+
+	result, err := r.client.BLPop(r.ctx, 1*time.Second, queue).Result()
```
devhaozi, line 95: This timeout Laravel is setting to 60 seconds. But I think 60 it too long, So I set it to 1 second.

hwbrzzl: 1 second is too short when using a remote Redis server, It's read timeout second, so I think 60 seconds is ok.

queue/driver_sync.go
```diff
@@ -0,0 +1,96 @@
...
+	return DriverSync
+}
+
+func (r *Sync) Push(job queue.Job, args []queue.Arg, queue string) error {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	r.size++
+	err := Call(job.Signature(), args)
+	r.size--
```
hwbrzzl, line 37: Q: The size doesn't seem to make sense, because of there is lock.

queue/driver_sync.go
```diff
@@ -0,0 +1,96 @@
...
+		if job.Delay != 0 {
+			time.Sleep(time.Duration(job.Delay) * time.Second)
+		}
+		err := Call(job.Job.Signature(), job.Args)
+		if err != nil {
+			r.size -= uint(len(jobs))
+			return err
+		}
+	}
+	r.size -= uint(len(jobs))
```
hwbrzzl, line 57: Ditto

queue/driver_sync.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func (r *Sync) Later(delay uint, job queue.Job, args []queue.Arg, queue string) error {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	r.size++
+	time.Sleep(time.Duration(delay) * time.Second)
+	err := Call(job.Signature(), args)
+	r.size--
```
hwbrzzl, line 69: Ditto

queue/task.go
```diff
@@ -40,52 +37,45 @@ func NewChainTask(config *Config, jobs []queue.Jobs) *Task {
...
 
 	if receiver.chain {
-		for _, job := range receiver.jobs {
-			if err := receiver.handleAsync(job.Job, job.Args); err != nil {
-				return err
-			}
-		}
-
-		return nil
+		return receiver.driver.Bulk(receiver.jobs, receiver.queue)
```
hwbrzzl, line 63: Q: Is the Delay suitable for the chain?

queue/config.go
```diff
@@ -22,40 +25,49 @@ func (r *Config) DefaultConnection() string {
...
+
+func (r *Config) Database(queueConnection string) orm.Query {
+	connection := r.config.GetString(fmt.Sprintf("queue.connections.%s.connection", queueConnection))
+	table := r.config.GetString(fmt.Sprintf("queue.connections.%s.table", queueConnection))
+	return OrmFacade.Connection(connection).Query().Table(table)
+}
 
-	return
+func (r *Config) FailedJobsDatabase() orm.Query {
```
hwbrzzl, line 69:
```suggestion
func (r *Config) FailedJobsQuery() orm.Query {
```

queue/worker.go
```diff
@@ -1,57 +1,84 @@
...
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for {
+				select {
+				case <-quitChan:
+					return
+				default:
+					job, args, err := r.driver.Pop(r.queue)
+					if err != nil {
```
hwbrzzl, line 52: S: There are a few different errors, we'd better handle it.

image

hwbrzzl (commented):
A great PR, but a bit complex, we may divide it into several PRs next time. 👍

queue/job.go
```diff
@@ -0,0 +1,143 @@
...
+		return nil, errors.New("job must implement contracts/queue/Job interface")
+	}
+
+	return job, nil
+}
+
+type Job struct {
+	ID            uint64               `gorm:"primaryKey" json:"id"`                        // The unique ID of the job.
+	Queue         string               `gorm:"not null" json:"queue"`                       // The name of the queue the job belongs to.
+	Job           string               `gorm:"not null" json:"job"`                         // The name of the handler for this job.
```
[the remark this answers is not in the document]
devhaozi: > Q: It's `Signature`, right? Is the name `job` in Laravel?

之前是因为Signature和Signature方法重复了，所以修改的，已经优化。

database/console/migrate_stubs.go
```diff
@@ -24,7 +23,7 @@ func (receiver MysqlStubs) CreateDown() string {
 
 // UpdateUp Update up migration content.
 func (receiver MysqlStubs) UpdateUp() string {
-	return `ALTER TABLE DummyTable ADD column varchar(255) COMMENT '';
```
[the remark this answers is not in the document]
devhaozi: > Q: Why change to `TEXT`?

一个错误，这个更改只应该在 PostgreSQL 上，因为 PG 中 TEXT 和 VARCHAR 等价

queue/driver_async.go
```diff
@@ -0,0 +1,143 @@
...
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	if _, exists := asyncJobs[queue]; !exists {
+		time.Sleep(1 * time.Second)
+		return nil, nil, errors.New("no job found in queue")
+	}
+	if len(asyncJobs[queue]) == 0 {
+		delete(asyncJobs, queue)
+		time.Sleep(1 * time.Second)
```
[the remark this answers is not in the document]
devhaozi: > Q: Why sleep one second?

防止死循环

devhaozi:
@hwbrzzl If redis BLPop timeout setting to 60 seconds, this test case will fail.
image

hwbrzzl:
> @hwbrzzl If redis BLPop timeout setting to 60 seconds, this test case will fail. 

Why?

devhaozi:
> > @hwbrzzl If redis BLPop timeout setting to 60 seconds, this test case will fail.
> 
> Why?

image
image

The logic here is similar to Laravel. First, need to merge the tasks in the delay queue into the main queue.

Next, BLPop will block until there is a task in the main queue or the timeout is reached.

Therefore, if it is set to 60s, it will always be blocked, causing the test to fail.

hwbrzzl:
The timeout parameter means an error will be thrown if something spends more time, right? Why will block?

devhaozi:
> The timeout parameter means an error will be thrown if something spends more time, right? Why will block?

No, no, no, this is a blocking method, it will block until there is a task in the queue list or the timeout is reached.
This is not specific to task running time, in fact, no control over task running time is currently implemented.

image

hwbrzzl:
@devhaozi I see, thanks!

hwbrzzl:
Hi @devhaozi Do you still work on this, please? The PR is a bit complex and spent a lot of time, I think we only need to implement the Redis driver for now, the DB driver is optional.

devhaozi:
> Hi @devhaozi Do you still work on this, please? The PR is a bit complex and spent a lot of time, I think we only need to implement the Redis driver for now, the DB driver is optional.

I'm busy with other work, you can take over this PR if you have time.

devhaozi:
> > Hi @devhaozi Do you still work on this, please? The PR is a bit complex and spent a lot of time, I think we only need to implement the Redis driver for now, the DB driver is optional.
> 
> I'm busy with other work, you can take over this PR if you have time.

I think only need to implement sync and async here, and consider splitting database and redis into two packages.

queue/driver_async.go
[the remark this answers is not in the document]
devhaozi: > S: Sort the methods by first letter. The same as below.

After sorting, the method does not look intuitive.

image

queue/machinery.go
hwbrzzl, line 1: This is an important feature, there is a lot of logic, I'm writing the v1.14 document now, so I think we can release this feature in v1.15, we can spend a bit longer time testing this feature, I'll review this PR after release v1.14. And I think we shouldn't remove the machinery driver for now, because users will hard to upgrade framework in the future, we need to create a new repository: goravel/machinery, so that users can upgrade framework easily. What do you think about this, please?

devhaozi: > This is an important feature, there is a lot of logic, I'm writing the v1.14 document now, so I think we can release this feature in v1.15, we can spend a bit longer time testing this feature, I'll review this PR after release v1.14. And I think we shouldn't remove the machinery driver for now, because users will hard to upgrade framework in the future, we need to create a new repository: goravel/machinery, so that users can upgrade framework easily. What do you think about this, please?

If the machinery is not removed, the related dependencies will not be removed either.

queue/config.go
```diff
@@ -22,40 +23,37 @@ func (r *Config) DefaultConnection() string {
...
-	host := r.config.GetString(fmt.Sprintf("database.redis.%s.host", connection))
-	password := r.config.GetString(fmt.Sprintf("database.redis.%s.password", connection))
-	port := r.config.GetInt(fmt.Sprintf("database.redis.%s.port", connection))
-	database = r.config.GetInt(fmt.Sprintf("database.redis.%s.database", connection))
-
-	if password == "" {
-		dsn = fmt.Sprintf("%s:%d", host, port)
-	} else {
-		dsn = fmt.Sprintf("%s@%s:%d", password, host, port)
+func (r *Config) Via(connection string) any {
```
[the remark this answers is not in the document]
devhaozi: > Is it driver?
> 
> image

No, driver will only support sync、async、custom, and via is for custom. (Cache use via, so I copy it.)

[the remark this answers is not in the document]
hwbrzzl: Got it

devhaozi:
After remove machinery, the binary size has been reduced by about 50MB.

image

image

hwbrzzl:
After the discussion in the group, I think a goravel/machinery is necessary, we need to implement the driver interface in this PR through machinery, otherwise, users will hard to upgrade the framework in the future. Another way is createing a migrate command: machinery -> db or redis, but we cannot guarantee that it will be compatible with all situations.

devhaozi:
> After the discussion in the group, I think a goravel/machinery is necessary, we need to implement the driver interface in this PR through machinery, otherwise, users will hard to upgrade the framework in the future. Another way is createing a migrate command: machinery -> db or redis, but we cannot guarantee that it will be compatible with all situations.

I prefer the second way.

hwbrzzl:
> I prefer the second way.

If so, we need to check the machinery code for how to generate a structure from the queue, then transform the structure to ours.

devhaozi:
Currently, machinery has resumed maintenance, so this PR may no longer be needed.

hwbrzzl:
machinery still doesn't support the DB driver, right?

devhaozi:
> machinery still doesn't support the DB driver, right?

Yes

hwbrzzl:
If so, I think it's better to implement the DB driver ourselves.

PR #336, opened 2023-11-05
fix: Validator Bind can't cast value

devhaozi:
Closes goravel/goravel#302

## ✅ Checks

### DIFF
--- a/validation/validation.go
+++ b/validation/validation.go
@@ -2,8 +2,7 @@ package validation
 
 import (
 	"errors"
-	"reflect"
-	"time"
+	"net/url"
 
 	"github.com/gookit/validate"
 
@@ -29,34 +28,25 @@ func (r *Validation) Make(data any, rules map[string]string, options ...validate
 		return nil, errors.New("rules can't be empty")
 	}
 
-	var dataType reflect.Kind
-	switch data := data.(type) {
+	var dataFace validate.DataFace
+	var err error
+	switch td := data.(type) {
+	case validate.DataFace:
+		dataFace = td
 	case map[string]any:
-		if len(data) == 0 {
+		if len(td) == 0 {
 			return nil, errors.New("data can't be empty")
 		}
-		dataType = reflect.Map
-	}
-
-	val := reflect.ValueOf(data)
-	indirectVal := reflect.Indirect(val)
-	typ := indirectVal.Type()
-	if indirectVal.Kind() == reflect.Struct && typ != reflect.TypeOf(time.Time{}) {
-		dataType = reflect.Struct
-	}
-
-	var dataFace validate.DataFace
-	switch dataType {
-	case reflect.Map:
-		dataFace = validate.FromMap(data.(map[string]any))
-	case reflect.Struct:
-		var err error
+		dataFace = validate.FromMap(td)
+	case url.Values:
+		dataFace = validate.FromURLValues(td)
+	case map[string][]string:
+		dataFace = validate.FromURLValues(td)
+	default:
 		dataFace, err = validate.FromStruct(data)
 		if err != nil {
 			return nil, err
 		}
-	default:
-		return nil, errors.New("data must be map[string]any or struct")
 	}
 
 	options = append(options, Rules(rules), CustomRules(r.rules))
--- a/validation/validation.go
+++ b/validation/validation.go
@@ -45,7 +45,7 @@ func (r *Validation) Make(data any, rules map[string]string, options ...validate
 	default:
 		dataFace, err = validate.FromStruct(data)
 		if err != nil {
-			return nil, err
+			return nil, errors.New("data must be map[string]any or map[string][]string or struct")
 		}
 	}
 
--- a/validation/validation.go
+++ b/validation/validation.go
@@ -60,7 +60,7 @@ func (r *Validation) Make(data any, rules map[string]string, options ...validate
 	v := dataFace.Create()
 	AppendOptions(v, generateOptions)
 
-	return NewValidator(v, dataFace), nil
+	return NewValidator(v), nil
 }
 
 func (r *Validation) AddRules(rules []validatecontract.Rule) error {
--- a/validation/validation_test.go
+++ b/validation/validation_test.go
@@ -42,7 +42,7 @@ func TestMake(t *testing.T) {
 			expectData:      Data{A: "b"},
 		},
 		{
-			description: "error when data isn't map[string]any or struct",
+			description: "error when data isn't map[string]any or map[string][]string or struct",
 			data:        "1",
 			rules:       map[string]string{"a": "required"},
 			expectErr:   errors.New("data must be map[string]any or map[string][]string or struct"),
--- a/validation/validation_test.go
+++ b/validation/validation_test.go
@@ -45,7 +45,7 @@ func TestMake(t *testing.T) {
 			description: "error when data isn't map[string]any or struct",
 			data:        "1",
 			rules:       map[string]string{"a": "required"},
-			expectErr:   errors.New("invalid input data"),
+			expectErr:   errors.New("data must be map[string]any or map[string][]string or struct"),
 		},
 		{
 			description: "error when data is empty map",
--- a/validation/validation_test.go
+++ b/validation/validation_test.go
@@ -45,7 +45,7 @@ func TestMake(t *testing.T) {
 			description: "error when data isn't map[string]any or struct",
 			data:        "1",
 			rules:       map[string]string{"a": "required"},
-			expectErr:   errors.New("data must be map[string]any or struct"),
+			expectErr:   errors.New("invalid input data"),
 		},
 		{
 			description: "error when data is empty map",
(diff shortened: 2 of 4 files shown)


validation/validator.go
hwbrzzl, line 1: Is the change copy from the other package, please? It's a complex logic, could you please describe the issue and add some test cases for it?

devhaozi: > Is the change copy from the other package, please? It's a complex logic, could you please describe the issue and add some test cases for it?

I'm trying to simplify the logic

devhaozi: > Is the change copy from the other package, please? It's a complex logic, could you please describe the issue and add some test cases for it?

The validator returns all string types for the query string, resulting in the inability to Bind to an int type structure.

devhaozi: > Is the change copy from the other package, please? It's a complex logic, could you please describe the issue and add some test cases for it?

I think this issue https://www.goravel.dev/the-basics/validation.html#int was also solved together.

devhaozi:
issue link to goravel/fiber/pull/36

validation/validator.go
```diff
@@ -67,3 +84,130 @@ func (v *Validator) Errors() httpvalidate.Errors {
...
+
+func getFieldKey(structField reflect.StructField) string {
+	if formTag := structField.Tag.Get("form"); formTag != "" {
+		return formTag
+	}
+	if jsonTag := structField.Tag.Get("json"); jsonTag != "" {
+		return jsonTag
+	}
+	return structField.Name
+}
```
devhaozi, line 96: Support to get key form struct tags.

validation/validator.go
```diff
@@ -67,3 +84,130 @@ func (v *Validator) Errors() httpvalidate.Errors {
...
+	for i := 0; i < val.NumField(); i++ {
+		field := val.Field(i)
+		if !field.CanSet() {
+			continue
+		}
+		structField := typ.Field(i)
+		tag := getFieldKey(structField)
+		tagToFieldMap[tag] = field
+		tagToFieldMap[strings.ToLower(tag)] = field
+		tagToFieldMap[str.Camel2Case(tag)] = field
```
devhaozi, line 120: Add two different cast to match common situations, like `UserName` `username` and `user_name`.

validation/validator.go
```diff
@@ -67,3 +84,130 @@ func (v *Validator) Errors() httpvalidate.Errors {
...
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func castValueToType(field reflect.Value, value any) (reflect.Value, error) {
+	var castedValue any
+	var err error
```
devhaozi, line 167: Auto try cast value to match struct's type.

validation/validator.go
```diff
@@ -67,3 +85,178 @@ func (v *Validator) Errors() httpvalidate.Errors {
...
+
+	return reflect.ValueOf(castedValue), nil
+}
+
+func (v *Validator) populateStruct(structVal reflect.Value, valueMap any) error {
+	valueMapCasted, ok := valueMap.(map[string]any)
+	if !ok {
+		return fmt.Errorf("%s: %s", errCastValueMap, reflect.TypeOf(valueMap).String())
+	}
+
```
devhaozi, line 234: This to support structure nesting, has passed the test in my project and CI test case.

devhaozi:
I found the `mapstructure` package that does the trick so we no longer need to implement the parsing step ourselves (it's complex and easy to make error).

devhaozi:
After more testing, https://www.goravel.dev/the-basics/validation.html#int is not resolved because the json.Unmarshal used internally by gookit/validate does not handle this case.

validation/validator_test.go
devhaozi, line 1: This test file has some error rules, I fixed it.

validation/validation.go
```diff
@@ -29,34 +28,25 @@ func (r *Validation) Make(data any, rules map[string]string, options ...validate
...
-	case reflect.Map:
-		dataFace = validate.FromMap(data.(map[string]any))
-	case reflect.Struct:
-		var err error
+		dataFace = validate.FromMap(td)
+	case url.Values:
+		dataFace = validate.FromURLValues(td)
+	case map[string][]string:
+		dataFace = validate.FromURLValues(td)
+	default:
```
devhaozi, line 45: I refactored it by referring to the `New` method of `github.com/gookit/validate`

validation/validation_test.go
```diff
@@ -106,7 +106,7 @@ func TestMake(t *testing.T) {
 				}),
 			},
 			expectValidator:    true,
-			expectData:         Data{A: "c"},
+			expectData:         Data{A: ""},
```
devhaozi, line 109: Now validator only Bind **safe** data, it meens validation must passed, so I change some test data.

devhaozi:
> After more testing, https://www.goravel.dev/the-basics/validation.html#int is not resolved because the json.Unmarshal used internally by gookit/validate does not handle this case.

This now can be solve use `filter` tag in Request struct.

See: goravel/fiber#38

validation/validator.go
```diff
@@ -67,3 +53,74 @@ func (v *Validator) Errors() httpvalidate.Errors {
 func (v *Validator) Fails() bool {
 	return v.instance.IsFail()
 }
+
+func (v *Validator) castValue() mapstructure.DecodeHookFunc {
```
hwbrzzl, line 57: Could you please add a unit test for this method?

validation/validation_test.go
hwbrzzl: You don't need to release this fix to v1.13.x, right? because it's a breaking change.

devhaozi: > You don't need to release this fix to v1.13.x, right? because it's a breaking change.

Yes

validation/validation.go
```diff
@@ -29,34 +28,25 @@ func (r *Validation) Make(data any, rules map[string]string, options ...validate
 		return nil, errors.New("rules can't be empty")
 	}
 
-	var dataType reflect.Kind
-	switch data := data.(type) {
+	var dataFace validate.DataFace
+	var err error
+	switch td := data.(type) {
+	case validate.DataFace:
```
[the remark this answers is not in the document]
devhaozi: > We may not need to support such a type.

This is for tests.

validation/validation.go
```diff
@@ -29,34 +28,25 @@ func (r *Validation) Make(data any, rules map[string]string, options ...validate
...
-	var dataFace validate.DataFace
-	switch dataType {
-	case reflect.Map:
-		dataFace = validate.FromMap(data.(map[string]any))
-	case reflect.Struct:
-		var err error
+		dataFace = validate.FromMap(td)
+	case url.Values:
+		dataFace = validate.FromURLValues(td)
+	case map[string][]string:
```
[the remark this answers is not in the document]
devhaozi: > Ditto, do we have a situation that uses this case?

We don't use it directly at the moment, but `Make` method accepts the any type, and I think it can be added.

validation/validation.go
```diff
@@ -29,34 +28,25 @@ func (r *Validation) Make(data any, rules map[string]string, options ...validate
...
+		dataFace = validate.FromURLValues(td)
+	default:
 		dataFace, err = validate.FromStruct(data)
 		if err != nil {
 			return nil, err
 		}
-	default:
-		return nil, errors.New("data must be map[string]any or struct")
```
hwbrzzl, line 59: > We don't use it directly at the moment, but Make method accepts the any type, and I think it can be added.

My word is related to this line, the `data` is the `any` type, if we don't return an error obviously, User may not know what to pass. The new type can be kept, but the error message may not be changed.

devhaozi: > > We don't use it directly at the moment, but Make method accepts the any type, and I think it can be added.
> 
> My word is related to this line, the `data` is the `any` type, if we don't return an error obviously, User may not know what to pass. The new type can be kept, but the error message may not be changed, or add the new types that we allow User to pass, but only `map[string][]string`, User may not know what `url.Values` and `validate.DataFace` are.

Added

validation/validator_test.go
```diff
@@ -185,6 +239,132 @@ func TestFails(t *testing.T) {
...
+				"j": "required",
+				"k": "required",
+				"l": "required",
+				"m": "required",
+				"n": "required",
+				"o": "required",
+				"p": "required",
+			},
+			expectData: Data{
+				A: "1",
```
hwbrzzl, line 319: This feature seems to support `string => int`, right? Could you please add some such cases?

validation/validator_test.go
```diff
@@ -185,6 +239,132 @@ func TestFails(t *testing.T) {
 	}
 }
 
+func TestCast(t *testing.T) {
```
hwbrzzl, line 242:
```suggestion
func TestCastValue(t *testing.T) {
```

validation/validator_test.go
devhaozi: > This feature seems to support `string => int`, right? Could you please add some such cases?

This already test in validation/validator_test.go L41-56

hwbrzzl: Yes, we can add a test case for `TestBind`, but this case belongs to the `CastValue` method directly, so we'd better add the case to `TestCastValue`, too.

hwbrzzl (approved):
Perfect!

PR #354, opened 2023-11-24
fix: [#312] orm connection is not working

hwbrzzl:
Closes # 

## ✅ Checks

### DIFF
--- a/database/gorm/query.go
+++ b/database/gorm/query.go
@@ -26,6 +26,7 @@ var _ ormcontract.Query = &QueryImpl{}
 
 type QueryImpl struct {
 	config        config.Config
+	connection    string
 	ctx           context.Context
 	instance      *gormio.DB
 	origin        *QueryImpl
@@ -33,7 +34,7 @@ type QueryImpl struct {
 	withoutEvents bool
 }
 
-func NewQueryImpl(ctx context.Context, config config.Config, gorm Gorm) (*QueryImpl, error) {
+func NewQueryImpl(ctx context.Context, connection string, config config.Config, gorm Gorm) (*QueryImpl, error) {
 	db, err := gorm.Make()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


database/gorm/conditions.go
hwbrzzl, line 1: Before, the `Where` method, etc. calls the Gorm method directly and gets a new Gorm instance.

Now, the conditions will be saved in `Conditions`, and will be built when calling a final method, like: `Get`.

After this change, we can control the conditions and the Gorm instance more freely.

database/gorm/event.go
hwbrzzl, line 1: There is no changes for this file, just order the methods.

database/gorm/query_test.go
```diff
@@ -2330,6 +2112,107 @@ func (s *QueryTestSuite) TestRaw() {
 	}
 }
 
+func (s *QueryTestSuite) TestReuse() {
```
hwbrzzl, line 2115: After this change, we can set `query` to a variable and reuse it, such as:

```
query := facades.Orm().Query().Where("a = b")
query.Where("b = c").Find() == facades.Orm().Query().Where("a = b").Where("b = c").Find()
query.Where("c = d").Find() == facades.Orm().Query().Where("a = b").Where("c = d").Find()
```

database/gorm/query.go
```diff
@@ -772,6 +1062,81 @@ func (r *QueryImpl) selectCreate(value any) error {
...
+			}
+
+			return nil
+		}
+	}
+
+	return nil
+}
+
+func (r *QueryImpl) new(db *gormio.DB) *QueryImpl {
```
hwbrzzl, line 1134: This is the key of `query` can be reused.

database/gorm/query.go
```diff
@@ -643,122 +703,352 @@ func (r *QueryImpl) Update(column any, value ...any) (*ormcontract.Result, error
...
-		switch arg := args[0].(type) {
-		case func(ormcontract.Query) ormcontract.Query:
-			newArgs := []any{
-				func(tx *gormio.DB) *gormio.DB {
-					query := arg(NewQueryImplByInstance(tx, r))
-
-					return query.(*QueryImpl).instance
-				},
-			}
+func (r *QueryImpl) buildConditions() *QueryImpl {
```
hwbrzzl, line 758: The original Gorm instance of QueryImpl will not be changed, to avoid the conditions chaotic.

database/gorm/query.go
```diff
@@ -1017,6 +1360,37 @@ func filterFindConditions(conds ...any) error {
 	return nil
 }
 
+func getModelConnection(model any) (string, error) {
```
hwbrzzl, line 1363: Fix:
1. Framework can't identify the connection of slice model;
2. Framework can't switch connections when the driver is the same;

devhaozi (approved):
LGTM

PR #371, opened 2023-12-16
feat: The lang json file supports nested folder and key.

kkumar-gcc:
Closes https://github.com/goravel/goravel/issues/319

## ✅ Checks


translation/translator.go
```diff
@@ -5,6 +5,8 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/bytedance/sonic"
+	"github.com/bytedance/sonic/ast"
```
devhaozi, line 9: `sonic` can't work on arm devices, pls change.

devhaozi (changes requested):

translation/translator.go
```diff
@@ -85,24 +87,42 @@ func (t *Translator) Get(key string, options ...translationcontract.Option) (str
...
+	marshal, err := sonic.Marshal(t.loaded[folder][locale])
+	if err != nil {
+		return "", err
+	}
+	var keyParts []interface{}
+	for _, part := range strings.Split(keyPart, ".") {
+		keyParts = append(keyParts, part)
+	}
+
+	keyValue, err := sonic.Get(marshal, keyParts...)
```
hwbrzzl, line 99: S: Don't use `sonic` directly.

translation/translator.go
```diff
@@ -85,24 +87,42 @@ func (t *Translator) Get(key string, options ...translationcontract.Option) (str
...
+
+	if err != nil {
+		if err == ast.ErrNotExist {
+			fallbackLocale := t.GetFallback()
```
hwbrzzl, line 103: S: `fallbackLocale` may be empty, should judge it.

translation/translator.go
```diff
@@ -85,24 +87,42 @@ func (t *Translator) Get(key string, options ...translationcontract.Option) (str
...
-			fallbackOptions.Fallback = translationcontract.Bool(false)
-			fallbackOptions.Locale = fallbackLocale
-			return t.Get(fallbackFolder+"."+keyPart, fallbackOptions)
+			return t.key, nil
 		}
-		return t.key, nil
+		return "", err
+	}
+
+	line, err := keyValue.String()
```
hwbrzzl, line 123: Q: Is it possible that `line` is empty?

contracts/translation/loader.go
```diff
@@ -2,5 +2,5 @@ package translation
 
 type Loader interface {
 	// Load the messages for the given locale.
-	Load(folder string, locale string) (map[string]map[string]string, error)
+	Load(folder string, locale string) (map[string]map[string]any, error)
```
hwbrzzl, line 5:
```suggestion
	Load(locale string, folder string) (map[string]map[string]any, error)
```

translation/translator_test.go
```diff
@@ -57,38 +57,51 @@ func (t *TranslatorTestSuite) TestChoice() {
...
 	translator = NewTranslator(t.ctx, t.mockLoader, "en", "en")
-	t.mockLoader.On("Load", "*", "en").Once().Return(nil, errors.New("some error"))
-	translation, err = translator.Choice("foo", 1)
+	t.mockLoader.On("Load", "test", "en").Once().Return(nil, errors.New("some error"))
+	translation, err = translator.Choice("test.foo", 1)
 	t.EqualError(err, "some error")
 	t.Equal("", translation)
 }
 
 func (t *TranslatorTestSuite) TestGet() {
```
hwbrzzl, line 66: S: Please add some  test cases for the nested key.

translation/translator_test.go
```diff
@@ -30,23 +30,23 @@ func (t *TranslatorTestSuite) SetupTest() {
 
 func (t *TranslatorTestSuite) TestChoice() {
```
hwbrzzl, line 31: S: Please add some test cases for the nested key.

translation/translator_test.go
```diff
@@ -193,32 +216,22 @@ func (t *TranslatorTestSuite) TestGetFallback() {
 func (t *TranslatorTestSuite) TestHas() {
```
hwbrzzl, line 216: S: Please add some test cases for the nested key.

translation/translator_test.go
```diff
@@ -248,7 +261,7 @@ func (t *TranslatorTestSuite) TestSetLocale() {
 
 func (t *TranslatorTestSuite) TestLoad() {
```
hwbrzzl, line 262: S: Please add some test cases for the nested key.

hwbrzzl (commented):
Good job 👍

translation/translator.go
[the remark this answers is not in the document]
kkumar-gcc: Yeah, that's a good suggestion. I will return key if there is any error and will log it.

translation/translator.go
```diff
@@ -17,7 +19,7 @@ type Translator struct {
 	loader   translationcontract.Loader
 	locale   string
 	fallback string
-	loaded   map[string]map[string]map[string]string
+	loaded   map[string]map[string]map[string]any
```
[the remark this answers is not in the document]
kkumar-gcc: Okay

translation/translator.go
kkumar-gcc: Good catch

translation/translator.go
kkumar-gcc: `line` empty means key doesn't exist which is handled just above of it.

translation/translator.go
kkumar-gcc: Then what should i use to get nested key value?

translation/translator.go
hwbrzzl: When `line` is empty, `Get` will return empty, I think it's unexpected.

translation/translator.go
```diff
@@ -81,38 +81,60 @@ func (t *Translator) Get(key string, options ...translationcontract.Option) (str
...
-		// load in the lines for the fallback locale and try to retrieve the
-		// translation for the given key.If it is translated, we will return it.
-		// Otherwise, we can finally return the key as that will be the final
-		// fallback.
-		if (folder+locale != fallbackFolder+fallbackLocale) && fallback {
-			var fallbackOptions translationcontract.Option
-			if len(options) > 0 {
-				fallbackOptions = options[0]
+	if err := t.load(locale, folder); err != nil && err != ErrFileNotExist {
+		//facades.Log().Error(err)
```
hwbrzzl, line 85: Q: Why annotate this line? And we can't use `facades.Log` directly here, should initiate it in ServiceProvider.

translation/translator.go
```diff
@@ -81,38 +81,60 @@ func (t *Translator) Get(key string, options ...translationcontract.Option) (str
...
+		return t.key
```
hwbrzzl, line 86: Q: Why return directly? Should call fallback.

translation/translator.go
```diff
@@ -81,38 +81,60 @@ func (t *Translator) Get(key string, options ...translationcontract.Option) (str
...
+	}
+
+	marshal, err := json.Marshal(t.loaded[locale][folder])
+	if err != nil {
+		//facades.Log().Error(err)
+		return t.key
```
hwbrzzl, line 92: Ditto

translation/translator.go
```diff
@@ -81,38 +81,60 @@ func (t *Translator) Get(key string, options ...translationcontract.Option) (str
...
-		// If the fallback locale is different from the current locale, we will
-		// load in the lines for the fallback locale and try to retrieve the
-		// translation for the given key.If it is translated, we will return it.
-		// Otherwise, we can finally return the key as that will be the final
-		// fallback.
-		if (folder+locale != fallbackFolder+fallbackLocale) && fallback {
-			var fallbackOptions translationcontract.Option
-			if len(options) > 0 {
-				fallbackOptions = options[0]
+	if err := t.load(locale, folder); err != nil && err != ErrFileNotExist {
```
hwbrzzl, line 84: Q: Why is `ErrFileNotExist` specific?

translation/translator_test.go
```diff
@@ -30,137 +30,174 @@ func (t *TranslatorTestSuite) SetupTest() {
...
 
 	// Case: when file exists but there is some error
 	translator = NewTranslator(t.ctx, t.mockLoader, "en", "en")
-	t.mockLoader.On("Load", "*", "en").Once().Return(nil, errors.New("some error"))
-	translation, err = translator.Get("foo")
-	t.EqualError(err, "some error")
-	t.Equal("", translation)
+	t.mockLoader.On("Load", "en", "test").Once().Return(nil, errors.New("some error"))
+	translation = translator.Get("test.foo")
+	//t.EqualError(err, "some error")
```
hwbrzzl, line 102: S: Remove this line.

translation/file_loader_test.go
```diff
@@ -32,28 +33,28 @@ func (f *FileLoaderTestSuite) SetupTest() {
 func (f *FileLoaderTestSuite) TestLoad() {
 	paths := []string{"./lang"}
 	loader := NewFileLoader(paths)
-	translations, err := loader.Load("*", "en")
+	translations, err := loader.Load("test", "en")
 	f.NoError(err)
 	f.NotNil(translations)
 	f.Equal("bar", translations["en"]["foo"])
```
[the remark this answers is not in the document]
kkumar-gcc: Cause it already exists 
image

translation/file_loader.go
```diff
@@ -22,38 +18,25 @@ func NewFileLoader(paths []string) *FileLoader {
...
-		// Check if the folder is not "*", and if so, split it into subFolders
-		if folder != "*" {
-			subFolders := strings.Split(folder, ".")
-			for _, subFolder := range subFolders {
-				fullPath = filepath.Join(fullPath, subFolder)
-			}
-		}
-		fullPath = filepath.Join(fullPath, locale+".json")
+		var val map[string]any
+		fullPath := filepath.Join(path, locale, folder+".json")
```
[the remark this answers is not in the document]
kkumar-gcc: I think it is not that confusing, though. We can rename it to group. Control related to different types of file loading is under file_loader. So instead of adding file extension to it we will just send the name of file with folder.

translation/translator.go
```diff
@@ -34,16 +37,13 @@ func NewTranslator(ctx context.Context, loader translationcontract.Loader, local
 		loader:   loader,
 		locale:   locale,
 		fallback: fallback,
-		loaded:   make(map[string]map[string]map[string]string),
+		loaded:   make(map[string]map[string]map[string]any),
```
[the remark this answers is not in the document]
kkumar-gcc: okay

translation/file_loader.go
```diff
@@ -22,38 +18,25 @@ func NewFileLoader(paths []string) *FileLoader {
 	}
 }
 
-func (f *FileLoader) Load(folder string, locale string) (map[string]map[string]string, error) {
-	translations := make(map[string]map[string]string)
+func (f *FileLoader) Load(locale string, folder string) (map[string]map[string]any, error) {
```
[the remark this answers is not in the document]
kkumar-gcc: I think `map[string]map[string]any` is okay since we know that loaded data from file will be `map[string]any`. And in translator file we are also expecting this.

translation/translator.go
kkumar-gcc: We will callback only if err is of type `ErrFileNotExist` else there might be some other error when loading file. should we call fallback for every error?

translation/translator.go
kkumar-gcc: okay

translation/translator.go
kkumar-gcc: I wanted to call fallback if there is error related to file not exists else it will return key instead of calling fallback,

translation/file_loader.go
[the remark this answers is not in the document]
hwbrzzl: The final structure is `map[string]map[string]map[string]any`, we set the first `string` in the translator file, so why not set the second `string` in the translator file, too? It will be easier to understand the whole logic.

translation/file_loader.go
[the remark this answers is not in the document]
hwbrzzl: Not really ~ 
1. If we need to support other types files in the future, this method is not good, it's better to pass a file with suffix;
2. It's very strange if pass a file but the file doesn't have a suffix;

translation/translator.go
hwbrzzl: Yes, we need to call fallback for every error.

kkumar-gcc: Actually when we are returning key our actual intention is to let user know that there is something error in your file. If we are returning fallback there might be case that fallback is present and user miss the error. So i think this will defeat the actual purpose of returning key.

translation/file_loader_test.go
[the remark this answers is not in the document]
hwbrzzl: I mean `translations["another/test"]["foo"]["xxx"]`

translation/file_loader.go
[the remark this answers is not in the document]
hwbrzzl: Based on the Laravel code, `folder` renamed to `group` makes sense, and doesn't need to contain the suffix. Thanks for your check. 👍

translation/translator.go
hwbrzzl: Sorry, based on the Laravel code, I think it's better to panic when encountering an invalid JSON error.

translation/file_loader.go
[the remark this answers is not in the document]
kkumar-gcc: Okay got it will change it

translation/translator.go
kkumar-gcc: okay

translation/file_loader_test.go
```diff
@@ -32,32 +33,32 @@ func (f *FileLoaderTestSuite) SetupTest() {
 func (f *FileLoaderTestSuite) TestLoad() {
 	paths := []string{"./lang"}
 	loader := NewFileLoader(paths)
-	translations, err := loader.Load("*", "en")
+	translations, err := loader.Load("en", "test")
 	f.NoError(err)
 	f.NotNil(translations)
-	f.Equal("bar", translations["en"]["foo"])
+	f.Equal("bar", translations["foo"])
```
hwbrzzl, line 39:
```suggestion
	f.Equal("bar", translations["baz"]["foo"])
```

translation/file_loader_test.go
```diff
@@ -32,32 +33,32 @@ func (f *FileLoaderTestSuite) SetupTest() {
...
+	translations, err = loader.Load("en", "test")
 	f.NoError(err)
 	f.NotNil(translations)
-	f.Equal("bar", translations["en"]["foo"])
+	f.Equal("backagebar", translations["foo"])
 
 	paths = []string{"./lang"}
 	loader = NewFileLoader(paths)
-	translations, err = loader.Load("another", "en")
+	translations, err = loader.Load("en", "another/test")
```
hwbrzzl, line 50:
```suggestion
	translations, err = loader.Load("en", "test")
```

translation/translator.go
```diff
@@ -17,9 +18,16 @@ type Translator struct {
 	loader   translationcontract.Loader
 	locale   string
 	fallback string
-	loaded   map[string]map[string]map[string]string
+	// loaded is a map structure used to store loaded translation data.
+	// It is organized as follows:
+	//   - First map (map[string]): Maps from locale to...
+	//     - Second map (map[string]): Maps from folder(group) to...
+	//       - Third map (map[string]): Maps from key to...
+	//         - Value (any): The translation line corresponding to the key in the specified locale, folder(group), and key hierarchy.
```
hwbrzzl, line 26: 👍

translation/translator.go
```diff
@@ -74,45 +80,51 @@ func (t *Translator) Get(key string, options ...translationcontract.Option) (str
...
+			return t.Get(t.key, fallbackOptions)
 		}
-		return t.key, nil
+		return t.key
+	}
+
+	line, ok := keyValue.(string)
+	if !ok {
+		return t.key
 	}
```
hwbrzzl, line 118:
```suggestion
	line := cast.ToString(keyValue)
	if line == "" {
		return t.key
	}
```

hwbrzzl (commented):
Great, almost finished.

translation/translator.go
```diff
@@ -81,37 +82,38 @@ func (t *Translator) Get(key string, options ...translationcontract.Option) stri
...
 
-	// Parse the key into group and key parts.
-	group, keyPart := parseKey(key)
-
 	// For JSON translations, there is only one file per locale, so we will
 	// simply load the file and return the line if it exists.
 	// If the file doesn't exist, we will return fallback if it is enabled.
 	// Otherwise, we will return the key as the line.
-	if err := t.load(locale, group); err != nil && err != ErrFileNotExist {
+	if err := t.load(locale, "*"); err != nil && err != ErrFileNotExist {
```
hwbrzzl, line 89: The `facades.Lang` is not a single instance, so the locale file will be loaded every time, we can set the `loaded` map to a global variable, and the performance will be improved.

translation/translator.go
```diff
@@ -81,37 +82,38 @@ func (t *Translator) Get(key string, options ...translationcontract.Option) stri
...
 		return t.key
 	}
 
-	keyValue := getValue(t.loaded[locale][group], keyPart)
-	// if the key is not found in the current locale group, we will try to load the
-	// fallback locale and try to retrieve the translation for the given key.
+	keyValue := getValue(t.loaded[locale]["*"], key)
+	// if the key is not found `{locale}.json` file, we will try to find it in
+	// the group files.
 	if keyValue == nil {
```
hwbrzzl, line 97: We can change this line to `if keyValue != nil {`, change the location of L120~L132 and L98~L117, the logic will be clearer.

translation/translator.go
```diff
@@ -185,6 +221,15 @@ func (t *Translator) load(locale string, group string) error {
 	return nil
 }
 
+func (t *Translator) localeArray(locale string) []string {
```
hwbrzzl, line 224:
```suggestion
func (t *Translator) appendFallbackLocale(locales []string) []string {
```

translation/translator.go
```diff
@@ -145,7 +147,22 @@ func (t *Translator) GetLocale() string {
...
+	if len(options) > 0 && options[0].Locale != "" {
+		locale = options[0].Locale
+	}
+
 	line := t.Get(key, options...)
+
+	// For `{locale}.json` JSON translations, the loaded files will contain the
+	// entire translation tree; therefore, we will check to see if the key
+	// exists in the array of translations.
+	keyValue := getValue(t.loaded[locale]["*"], key)
```
hwbrzzl, line 161: Q: Doesn't judge `lang/en/test.json`?

translation/translator.go
hwbrzzl, line 1: 👍 It's clearer.

translation/translator.go
kkumar-gcc: well if we couldn't load from en.json will go to next step. Here will will try load from a specific group.
image

translation/translator.go
```diff
@@ -169,6 +186,25 @@ func (t *Translator) SetLocale(locale string) context.Context {
 	return t.ctx
 }
 
+func (t *Translator) getLine(locale string, group string, key string, options ...translationcontract.Option) string {
+	if err := t.load(locale, group); err != nil && err != ErrFileNotExist {
+		t.logger.Panic(err)
+		return t.key
+	}
+
+	keyValue := getValue(t.loaded[locale][group], key)
```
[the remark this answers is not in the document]
kkumar-gcc: if key is mepty keyValue will be nil

translation/translator.go
hwbrzzl: Yes, sorry, I ignored the `line := t.Get(key, options...)` line, but if has `line := t.Get(key, options...)`, I think this line is unnecessary, had loaded `t.loaded[locale]["*"]` in the `Get` method.

translation/translator.go
hwbrzzl: ?

translation/translator.go
kkumar-gcc: Yeah, You are right. I will remove it

translation/translator.go
```diff
@@ -54,10 +62,10 @@ func (t *Translator) Choice(key string, number int, options ...translationcontra
 		locale = options[0].Locale
 	}
 
-	return makeReplacements(t.selector.Choose(line, number, locale), replace), nil
+	return makeReplacements(t.selector.Choose(line, number, locale), replace)
 }
 
-func (t *Translator) Get(key string, options ...translationcontract.Option) (string, error) {
+func (t *Translator) Get(key string, options ...translationcontract.Option) string {
```
hwbrzzl, line 68: There is a problem, the files: 

`lang/cn.json`:

```
{
  "hello": "cn-world"
}
```

`json/en.json`, 

```
{
  "hello": "en-world",
  "hello1": "en-world1"
}
```

The configuration:

```
"locale": "cn",
"fallback_locale": "en",
```

`facades.Lang(context.Background()).Get("hello1")` will be returned `hello1`, but expect `en-world1`.

So the logic of this method should be:

1. Get value from `lang/cn.json`;
2. If `lang/cn.json` exists, but the key doesn't exist, get value from `lang/en.json`, if `lang/en.json` doesn't exist, get value from `lang/en/*`;
3. If `lang/cn.json` doesn't exist, get value from `lang/cn/*`;
4. If `lang/cn/*` doesn't exist, get value from `lang/en.json`, if `lang/en.json` doesn't exist, get value from `lang/en/*`;

kkumar-gcc: Fallback translations are not allowed for `lang/en.json`

kkumar-gcc: Fallback translations are not allowed for `lang/cn.json`(It is intended since laravel doesn't support it)

hwbrzzl (commented):
Great, the test passed locally, only left a few performance optimizations.

translation/translator.go
kkumar-gcc: We are not support "/" explicitly just like laravel.

hwbrzzl (approved):
Great job, Bro. The `BeforeTest` method is unnecessary, I removed it directly. Please @devhaozi make an approval.

image

devhaozi (approved):
LGTM

PR #385, opened 2024-01-03
feat: prisma go client added

hosseinmirzapur:
Closes [issue #344](https://github.com/goravel/goravel/issues/344)

## 📑 Description

I've added the support for prisma go client to the artisan commands and the list of artisan commands are now updated with prisma commands as I added each command to the database directory service provider and tested each command, so no error happens. This prisma client is not supposed to change the way Goravel handles the DB right now. It's simply a database tool to handle migrations and database schema easier and it's fully compatible with the current built-in ORM tool. I haven't changed the facades.Orm() functionalities, and simply all I did is, added commands for prisma schema. Hope everyone has a better time developing great apps

- [x] Added prisma commands besides current artisan console commands
- [ ] Prisma Studio command (according to [this issue](https://github.com/steebchen/prisma-client-go/issues/1084) it's in development process)

## ✅ Checks

- [ ] I have updated the documentation as required (It doesn't require any changes to the docs, as it has all documentation included [here](https://www.prisma.io/docs))

## ℹ Additional Information

Dependencies added:
- [prisma-client-go](https://github.com/steebchen/prisma-client-go)

Comparisons between new and old behavior:
- more artisan commands available, prisma included


hwbrzzl:
Thanks for your PR. Great work, but based on the https://github.com/goravel/goravel/issues/344, I just think you want to use Prisma instead of Gorm in the Orm module and create a new development package for it.

We need to have a deep discussion if you want to add a new feature(like the commands in this PR) first, so unfortunately, this PR may not be merged into master at this time.

hosseinmirzapur:
> Thanks for your PR. Great work, but based on the [goravel/goravel#344](https://github.com/goravel/goravel/issues/344), I just think you want to use Prisma instead of Gorm in the Orm module and create a new development package for it.
> 
> We need to have a deep discussion if you want to add a new feature(like the commands in this PR) first, so unfortunately, this PR may not be merged into master at this time.

Actually there is no replacement for the Gorm in the Orm module because it may change a lot of settings and a lot of files, which is not convenient. My goal was to make database operations easier, as if you check my added codes, you'll see that i'm not replacing Gorm with Prisma, but leveraging Prisma migration tools and commands to automate the migration creation, and make database module a lot more abstract than what it already is. All app models and Gorm functionalities may work just like before and also, current database artisan commands may work perfectly, and there's no change in the flow of Goravel app, and all Prisma will do is ease the way developers interact with database and also supports Mongodb as NoSQL and many relational databases which are yet to be added to Goravel, as their drivers don't exist(yet). Prisma is way more mature than Gorm by the way, but I actually haven't implemented the replacement for Gorm because i didn't want to change many codes and chose to get on with the Goravel flow and code principle. I made my intentions clear, so that my offered features may not sound too complicated or out of context

hwbrzzl:
Yes, I understand, the PR is great, but we are not ready to integrate Prisma into the framework yet, this involves many aspects, so I suggest you can add these commands to an independent development package first. For detail: https://www.goravel.dev/digging-deeper/package-development.html

hosseinmirzapur:
> Yes, I understand, the PR is great, but we are not ready to integrate Prisma into the framework yet, this involves many aspects, so I suggest you can add these commands to an independent development package first. For detail: https://www.goravel.dev/digging-deeper/package-development.html

Very well, I'll put them in a separate package as a library, and i'll open another PR. Thanks

hosseinmirzapur:
Hello again, I just finished developing the Go Prisma Client by following Goravel Extend package workflow.
You can find it here: [Goravel Prisma](https://github.com/hosseinmirzapur/prisma)
If you want to transfer ownership of the package, or use it by only referring to Goravel name on it, feel free to fork the mentioned repository, to create a new repository for Goravel.
Best Regards.

hwbrzzl:
Thanks @hosseinmirzapur That's great!

PR #399, opened 2024-01-27
feat: add OrWhereNotIn and WhereNotIn methods

Kamandlou:
## 📑 Description
Add OrWhereNotIn and WhereNotIn methods

## ✅ Checks

### DIFF
--- a/database/gorm/query_test.go
+++ b/database/gorm/query_test.go
@@ -2776,13 +2776,9 @@ func (s *QueryTestSuite) TestOrWhereNotIn() {
 			s.Nil(query.Create(&user1))
 			s.True(user1.ID > 0)
 
-			user2 := User{Name: "where_in_user_2", Avatar: "where_in_avatar_2"}
-			s.Nil(query.Create(&user2))
-			s.True(user2.ID > 0)
-
 			var users []User
 			s.Nil(query.Where("id = ?", -1).OrWhereNotIn("id", []any{user.ID, user1.ID}).Find(&users))
-			s.True(len(users) >= 1)
+			s.True(len(users) >= 0)
 		})
 	}
 }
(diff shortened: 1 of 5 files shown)


database/gorm/query_test.go
```diff
@@ -2743,6 +2743,50 @@ func (s *QueryTestSuite) TestOrWhereIn() {
...
+			user1 := User{Name: "where_in_user_1", Avatar: "where_in_avatar_1"}
+			s.Nil(query.Create(&user1))
+			s.True(user1.ID > 0)
+
+			user2 := User{Name: "where_in_user_2", Avatar: "where_in_avatar_2"}
+			s.Nil(query.Create(&user2))
+			s.True(user2.ID > 0)
+
+			var users []User
+			s.Nil(query.WhereNotIn("id", []any{user.ID, user1.ID}).Find(&users))
```
hwbrzzl, line 2762: S: We'd better add a `Where` condition with `WhereNotIn`, only get the `user2` row, it will be more accurate. So the `Avatar` of these three SQL can be the same.

database/gorm/query_test.go
```diff
@@ -2743,6 +2743,50 @@ func (s *QueryTestSuite) TestOrWhereIn() {
...
+
+			user1 := User{Name: "where_in_user_1", Avatar: "where_in_avatar_1"}
+			s.Nil(query.Create(&user1))
+			s.True(user1.ID > 0)
+
+			user2 := User{Name: "where_in_user_2", Avatar: "where_in_avatar_2"}
+			s.Nil(query.Create(&user2))
+			s.True(user2.ID > 0)
+
+			var users []User
```
hwbrzzl, line 2783: Ditto

database/gorm/query_test.go
Kamandlou: Done.

database/gorm/query_test.go
Kamandlou: What should I do in this line?

hwbrzzl: We'd better create `user2`, then judge `users` contains `user2`.

hwbrzzl (approved):
Great 👍

PR #405, opened 2024-02-01
feat: add InRandomOrder method

Kamandlou:
## 📑 Description
add InRandomOrder method

## ✅ Checks

### DIFF
--- a/contracts/database/orm/constants.go
+++ b/contracts/database/orm/constants.go
@@ -3,6 +3,7 @@ package orm
 const (
 	DriverMysql      Driver = "mysql"
 	DriverPostgresql Driver = "postgresql"
+	DriverPostgres   Driver = "postgres"
 	DriverSqlite     Driver = "sqlite"
 	DriverSqlserver  Driver = "sqlserver"
 )
--- a/contracts/database/orm/constants.go
+++ b/contracts/database/orm/constants.go
@@ -1,7 +1,8 @@
 package orm
 
 const (
-	DriverMysql      Driver = "mysql"
+	DriverMysql Driver = "mysql"
+	// DriverPostgresql DEPRECATED, use DriverPostgres instead.
 	DriverPostgresql Driver = "postgresql"
 	DriverPostgres   Driver = "postgres"
 	DriverSqlite     Driver = "sqlite"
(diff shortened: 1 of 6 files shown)


database/gorm/query.go
```diff
@@ -455,6 +455,25 @@ func (r *QueryImpl) OrderByDesc(column string) ormcontract.Query {
 	return r.Order(fmt.Sprintf("%s DESC", column))
 }
 
+func (r *QueryImpl) InRandomOrder() ormcontract.Query {
+	order := ""
+	switch r.Driver() {
+	case "mysql":
```
hwbrzzl, line 461: S: Use `orm.DriverMysql.String()` instead.

database/gorm/query_test.go
```diff
@@ -1973,6 +1973,24 @@ func (s *QueryTestSuite) TestOrderByDesc() {
...
+			s.True(user.ID > 0)
+
+			user1 := User{Name: "random_order_user", Avatar: "random_order_avatar1"}
+			s.Nil(query.Create(&user1))
+			s.True(user1.ID > 0)
+
+			var users []User
+			s.Nil(query.Where("name = ?", "random_order_user").InRandomOrder().Get(&users))
+			s.True(len(users) == 2)
+		})
```
hwbrzzl, line 1990: S: This case can't confirm the validity of `InRandomOrder`, we can create multiple users(for example 10 times), query them second times, then compare the ID of index 0, 5, 10.

hwbrzzl (changes requested):

database/gorm/query.go
Kamandlou: when I use Postgresql database `r.Driver()` returns "postgres" but `ormcontract.DriverPostgresql` returns "postgresql" and it causes a problem

Kamandlou: what should I do in this case?
I can define a default value for `order` variable. 
like this: `order := "RAND()"`

hwbrzzl: We can define a new constant: `ormcontract.DriverPostgres`, we want to use `postgres` in the future.

contracts/database/orm/constants.go
```diff
@@ -3,6 +3,7 @@ package orm
 const (
 	DriverMysql      Driver = "mysql"
 	DriverPostgresql Driver = "postgresql"
```
hwbrzzl, line 5: Could you please add the DEPRECATED tag for this line?

Kamandlou: Yes

hwbrzzl (approved):
Great work!

PR #428, opened 2024-02-24
feat: Add Session Module

kkumar-gcc:
### PR Description
RelatedTo goravel/goravel#357

### 📑 **Description:**

This pull request introduces a new `session` management module.

**Note**: Some methods in the session interface support chaining.

### Session Store Methods

- `Start`: Initiates the session by loading it and regenerating the token if it does not exist.
- `Save`: Saves the session data to the handler.
- `All`: Retrieves all attributes of the session.
- `Has`: Checks if the specified key exists and is not nil in the session attributes.
- `Get`: Retrieves the value of the specified key from the session attributes, with an optional default value.
- `Put`: Sets the value of the specified key in the session attributes.
- `RegenerateToken`: Regenerates the session token.
- `Forget`: Removes the specified keys from the session attributes.

### Session Manager Methods

- `BuildSession`: Builds a session with a specific handler and optional session ID.
- `Driver`: Retrieves a session driver by name.
- `Extend`: Extends the session manager with a custom driver.
- `Store`: Retrieves a session by session ID.

### Example Usage with Session Facade
```go
// Retrieve a session manager using the facade
sessionManager := facades.Session()

// Retrieve a session by session ID
session := sessionManager.Store("session_id")

// Alternatively, you can retrieve a session without specifying the session ID
sessionWithoutID := sessionManager.Store()

// Build a session with a specific handler and optional session ID
customSession := sessionManager.BuildSession(customHandler, "session_id")

// Retrieve a session driver
driver, err := sessionManager.Driver("file")
if err != nil {
    // Handle error
}

// Extend the session manager with a custom driver
sessionManager.Extend("custom_driver", func() sessioncontract.Handler {
    return CustomHandler()
})

// Start the session
session.Start()

// Save session data
session.Save()

// Set a session value
session.Put("key", "value")

// Retrieve session value
value := session.Get("key")

// Regenerate session token
session.RegenerateToken()
```

PR #436, opened 2024-03-09
feat: Add Session Module 2 

kkumar-gcc:
Closes # 
Related PR: #428 

## ✅ Checks

### DIFF
--- a/session/session.go
+++ b/session/session.go
@@ -2,6 +2,7 @@ package session
 
 import (
 	"maps"
+	"slices"
 
 	sessioncontract "github.com/goravel/framework/contracts/session"
 	"github.com/goravel/framework/support/json"
@@ -37,6 +38,23 @@ func (s *Session) All() map[string]any {
 	return s.attributes
 }
 
+func (s *Session) Exists(key string) bool {
+	return supportmaps.Exists(s.attributes, key)
+}
+
+func (s *Session) Flash(key string, value any) sessioncontract.Session {
+	s.Put(key, value)
+	s.Push("_flash.new", key)
+	s.removeFromOldFlashData(key)
+
+	return s
+}
+
+func (s *Session) Flush() sessioncontract.Session {
+	s.attributes = make(map[string]any)
+	return s
+}
+
 func (s *Session) Forget(keys ...string) sessioncontract.Session {
 	supportmaps.Forget(s.attributes, keys...)
 
@@ -64,16 +82,77 @@ func (s *Session) Has(key string) bool {
 	return val != nil
 }
 
+func (s *Session) Invalidate() error {
+	s.Flush()
+	return s.Migrate(true)
+}
+
+func (s *Session) Migrate(destroy ...bool) error {
+	shouldDestroy := false
+	if len(destroy) > 0 {
+		shouldDestroy = destroy[0]
+	}
+
+	if shouldDestroy {
+		err := s.driver.Destroy(s.GetID())
+		if err != nil {
+			return err
+		}
+	}
+
+	s.SetID(s.generateSessionID())
+
+	return nil
+}
+
+func (s *Session) Missing(key string) bool {
+	return !s.Exists(key)
+}
+
+func (s *Session) Only(keys []string) map[string]any {
+	return supportmaps.Only(s.attributes, keys...)
+}
+
+func (s *Session) PreviousUrl() string {
+	return s.Get("_previous.url").(string)
+}
+
+func (s *Session) Pull(key string, def ...any) any {
+	return supportmaps.Pull(s.attributes, key, def...)
+}
+
+func (s *Session) Push(key string, value any) sessioncontract.Session {
+	arr := s.Get(key, make([]any, 0)).([]any)
+	arr = append(arr, value)
+	return s.Put(key, arr)
+}
+
 func (s *Session) Put(key string, value any) sessioncontract.Session {
 	s.attributes[key] = value
 	return s
+}
+
+func (s *Session) Regenerate(destroy ...bool) error {
+	err := s.Migrate(destroy...)
+	if err != nil {
+		return err
+	}
+
+	s.RegenerateToken()
+	return nil
 }
 
 func (s *Session) RegenerateToken() sessioncontract.Session {
 	return s.Put("_token", str.Random(40))
 }
 
+func (s *Session) Remove(key string) any {
+	return s.Pull(key)
+}
+
 func (s *Session) Save() error {
+	s.ageFlashData()
+
 	data, err := json.MarshalString(s.attributes)
 	if err != nil {
 		return err
@@ -104,6 +183,10 @@ func (s *Session) SetName(name string) sessioncontract.Session {
 	return s
 }
 
+func (s *Session) SetPreviousUrl(url string) sessioncontract.Session {
+	return s.Put("_previous.url", url)
+}
+
 func (s *Session) Start() bool {
 	s.loadSession()
 
@@ -113,6 +196,10 @@ func (s *Session) Start() bool {
 
 	s.started = true
 	return s.started
+}
+
+func (s *Session) Token() string {
+	return s.Get("_token").(string)
 }
 
 func (s *Session) generateSessionID() string {
@@ -140,4 +227,26 @@ func (s *Session) readFromHandler() map[string]any {
 		return nil
 	}
 	return data
+}
+
+func (s *Session) ageFlashData() {
+	oldAny := s.Get("_flash.old", make([]any, 0)).([]any)
+	old := make([]string, len(oldAny))
+	for i, v := range oldAny {
+		old[i] = v.(string)
+	}
+
+	s.Forget(old...)
+	s.Put("_flash.old", s.Get("_flash.new", make([]any, 0)))
+	s.Put("_flash.new", make([]any, 0))
+}
+
+func (s *Session) removeFromOldFlashData(keys ...string) {
+	old := s.Get("_flash.old", make([]any, 0)).([]any)
+	for _, key := range keys {
+		old = slices.DeleteFunc(old, func(i any) bool {
+			return i == key
+		})
+	}
+	s.Put("_flash.old", old)
 }
--- a/session/session_test.go
+++ b/session/session_test.go
@@ -2,12 +2,15 @@ package session
 
 import (
 	"errors"
+	"strings"
 	"testing"
 
+	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/suite"
 
 	mocksession "github.com/goravel/framework/mocks/session"
 	"github.com/goravel/framework/support/json"
+	"github.com/goravel/framework/support/str"
 )
 
 type SessionTestSuite struct {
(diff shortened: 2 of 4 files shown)


session/session.go
```diff
@@ -64,16 +82,77 @@ func (s *Session) Has(key string) bool {
...
+func (s *Session) PreviousUrl() string {
+	return s.Get("_previous.url").(string)
+}
+
+func (s *Session) Pull(key string, def ...any) any {
+	return supportmaps.Pull(s.attributes, key, def...)
+}
+
+func (s *Session) Push(key string, value any) sessioncontract.Session {
+	arr := s.Get(key, make([]any, 0)).([]any)
```
hwbrzzl, line 125: Will panic if the value isn't `[]any`.

hwbrzzl, line 125: Will panic if the value isn't `[]any`.

session/session.go
```diff
@@ -141,3 +228,25 @@ func (s *Session) readFromHandler() map[string]any {
...
+	s.Forget(old...)
+	s.Put("_flash.old", s.Get("_flash.new", make([]any, 0)))
+	s.Put("_flash.new", make([]any, 0))
+}
+
+func (s *Session) removeFromOldFlashData(keys ...string) {
+	old := s.Get("_flash.old", make([]any, 0)).([]any)
+	for _, key := range keys {
+		old = slices.DeleteFunc(old, func(i any) bool {
+			return i == key
```
hwbrzzl, line 248: `i` is `any`, `key` is `string`, they shouldn't be equal.

session/session.go
```diff
@@ -64,16 +82,77 @@ func (s *Session) Has(key string) bool {
 	return val != nil
 }
 
+func (s *Session) Invalidate() error {
+	s.Flush()
+	return s.Migrate(true)
+}
+
+func (s *Session) Migrate(destroy ...bool) error {
```
hwbrzzl, line 90: In what scenarios is the function used, please?

session/session.go
```diff
@@ -64,16 +82,77 @@ func (s *Session) Has(key string) bool {
...
+
+func (s *Session) Missing(key string) bool {
+	return !s.Exists(key)
+}
+
+func (s *Session) Only(keys []string) map[string]any {
+	return supportmaps.Only(s.attributes, keys...)
+}
+
+func (s *Session) PreviousUrl() string {
```
hwbrzzl, line 116: In what scenarios is the function used, please?

session/session.go
```diff
@@ -64,16 +82,77 @@ func (s *Session) Has(key string) bool {
...
+		return err
+	}
+
+	s.RegenerateToken()
+	return nil
+}
+
 func (s *Session) RegenerateToken() sessioncontract.Session {
 	return s.Put("_token", str.Random(40))
 }
```
hwbrzzl, line 147: In what scenarios are these two functions used? What's the difference between them?

session/session.go
```diff
@@ -141,3 +228,25 @@ func (s *Session) readFromHandler() map[string]any {
...
+	oldAny := s.Get("_flash.old", make([]any, 0)).([]any)
+	old := make([]string, len(oldAny))
+	for i, v := range oldAny {
+		old[i] = v.(string)
+	}
+
+	s.Forget(old...)
+	s.Put("_flash.old", s.Get("_flash.new", make([]any, 0)))
+	s.Put("_flash.new", make([]any, 0))
+}
```
hwbrzzl, line 242: Sorry, I didn't get the idea of this function, could you please make an example?

session/session_test.go
```diff
@@ -34,6 +37,55 @@ func (s *SessionTestSuite) TestAll() {
 	s.Equal("value2", all["key2"])
 }
 
+func (s *SessionTestSuite) TestExists() {
+	s.session.Put("foo", "bar")
+	s.True(s.session.Exists("foo"))
+	s.session.Put("baz", nil)
+	s.False(s.session.Has("baz"))
+	s.True(s.session.Exists("baz"))
```
hwbrzzl, line 45: It's `false` when `Has`, but it's `true` when `Exists`, a bit confused.

session/session_test.go
```diff
@@ -34,6 +37,55 @@ func (s *SessionTestSuite) TestAll() {
...
+
+	s.session.ageFlashData()
+
+	s.True(s.session.Has("foo"))
+	s.Equal("bar", s.session.Get("foo"))
+	s.Equal(0, s.session.Get("bar"))
+
+	s.session.ageFlashData()
+
+	s.False(s.session.Has("foo"))
```
hwbrzzl, line 68: Why it's `true` in line 62, but it's `false` in line 68? I think they should both `false`, because `foo` is got in lin 56.

session/session_test.go
```diff
@@ -34,6 +37,55 @@ func (s *SessionTestSuite) TestAll() {
...
+}
+
+func (s *SessionTestSuite) TestFlush() {
+	s.session.Put("foo", "bar")
+	s.session.Flush()
+	s.False(s.session.Has("foo"))
+
+	s.session.Put("foo", "bar")
+	s.session.Flush()
+	s.False(s.session.Has("foo"))
```
hwbrzzl, line 86: Why repeat the test?

session/session_test.go
```diff
@@ -70,6 +122,71 @@ func (s *SessionTestSuite) TestHas() {
...
+	s.Nil(s.session.Migrate(true))
+	s.NotEqual(oldID, s.session.GetID())
+}
+
+func (s *SessionTestSuite) TestMissing() {
+	s.session.Put("foo", "bar")
+	s.False(s.session.Missing("foo"))
+	s.session.Put("baz", nil)
+	s.False(s.session.Has("baz"))
+	s.False(s.session.Missing("baz"))
```
hwbrzzl, line 163: Ditto

session/session.go
```diff
@@ -64,16 +82,77 @@ func (s *Session) Has(key string) bool {
...
+
+func (s *Session) PreviousUrl() string {
+	return s.Get("_previous.url").(string)
+}
+
+func (s *Session) Pull(key string, def ...any) any {
+	return supportmaps.Pull(s.attributes, key, def...)
+}
+
+func (s *Session) Push(key string, value any) sessioncontract.Session {
```
hwbrzzl, line 124: In what scenarios is the function used, please?

session/session.go
kkumar-gcc: It is used to migrate one session to different id optionally delete previous session. It is used in Invalidate and Regenerate function(extensions).

session/session.go
kkumar-gcc: `RegenerateToken` will solely regenerate the token, which can be utilized for CSRF verification. `Regenerate`, on the other hand, optionally deletes the current session.

[Reference](https://laravel.com/docs/10.x/session#regenerating-the-session-id)

image

session/session.go
kkumar-gcc: `PreviousUrl()` can be used to retrieve the URL of the previous request made by the user in the current session. It is useful when we need to redirect the user back to the previous request after performing some kind of action.

session/session_test.go
kkumar-gcc: will remove it

session/session.go
kkumar-gcc: I will remove this function, not needed.

session/session_test.go
kkumar-gcc: That's just how we define the `Has` and `Exists` methods.

image

session/session_test.go
kkumar-gcc: It is not repeating. The value of the key `baz` is `nil`, so `Has` should return `True`, and `Missing` is just the negation of `Exists`, so it should return `False`.

session/session.go
kkumar-gcc: In this PR, I am adding a feature to manage flash data. For which I have added a `Flash` method which will add data for the next request in the session. So I have added two arrays: one is `_flash.new` (in attributes) which will store flash keys of the current request and `_flash.old` will store flash keys from the previous request. To achieve this when saving, I don't want to save data (`Save`) from the previous request (i.e., keys stored in `_flash.old`). I will remove those keys, and whatever we have in the current request will be added into `_flash.old` so that it can be removed in the next request.

image
image
image

session/session_test.go
kkumar-gcc: In the first `ageFlashData` call, it removes all keys that exist in `_flash.old` and adds all new flash keys into it. However, when we call it a second time, `"foo"` and `"bar"` are already in `_flash.old`, so they will be removed just like they were in the first call (similar to the `http` request). [This function will only be called once per request.]

session/session.go
hwbrzzl: Could you please point out a real scenario? When we want to maintain session data but need to change the session ID?

session/session_test.go
hwbrzzl: Makes sense

session/session.go
hwbrzzl: I see, it's for security. One more thing, I think this function can be private, User doesn't need it directly.

session/session.go
hwbrzzl: Yes, I got you, I remembered a few scenarios about this, but I think the function currently has no practical use in Goravel, compared with Larave, Goravel still lacks a lot in terms of the View module. In conclusion, we can leave it here, it doesn't matter.

session/session.go
hwbrzzl: I see, I noticed that the `RegenerateToken` is a private method in Laravel, at least it doesn't appear in the document, so I suggest we do this, too.

session/session.go
hwbrzzl: Okay, one more question, if we add the `flash` method, is it better to add the other methods: `reflash`, `keep`, and  `now`?

session/session.go
kkumar-gcc: Good suggestio, we can make 'RegenerateToken` as a private function.

session/session.go
kkumar-gcc: should I remove it?

session/session.go
kkumar-gcc: Good suggestion, I will add other three methods.

session/session.go
hwbrzzl: Of course, it can be removed if you also agree.

kkumar-gcc: Since, we don't need it right now, I will remove it.

session/session_test.go
```diff
@@ -70,6 +118,85 @@ func (s *SessionTestSuite) TestHas() {
...
+	s.False(s.session.Has("foo"))
+	s.Nil(s.session.Get("foo"))
+}
+
+func (s *SessionTestSuite) TestOnly() {
+	s.session.
+		Put("foo", "bar").
+		Put("baz", "qux")
+
+	s.Equal(map[string]any{"foo": "bar", "baz": "qux"}, s.session.All())
```
hwbrzzl, line 189: The map is random, we can't equal `map` like this, need to compare keys one by one.

session/session_test.go
```diff
@@ -82,32 +209,87 @@ func (s *SessionTestSuite) TestPut() {
...
 
-	data, _ = json.MarshalString(s.session.All())
-	s.driver.On("Write", s.getSessionId(), data).Return(errors.New("error")).Once()
+	data, _ = json.MarshalString(map[string]any{
+		"key1":       "value1",
+		"_token":     s.session.Token(),
+		"_flash.new": []any{},
+		"_flash.old": []any{},
+	})
+	s.driver.On("Write", s.getSessionId(), mock.MatchedBy(func(v string) bool {
```
hwbrzzl, line 285:
```suggestion
	s.driver.On("Write", s.getSessionID(), mock.MatchedBy(func(v string) bool {
```

session/session_test.go
```diff
@@ -152,6 +334,29 @@ func (s *SessionTestSuite) TestStart() {
 	s.Equal("qux", s.session.Get("baz"))
 }
 
+func (s *SessionTestSuite) TestToken() {
+	s.driver.On("Read", s.getSessionId()).Return(`{"foo":"bar"}`, nil).Once()
+	s.session.Start()
+
+	s.Equal(s.session.Token(), s.session.Token())
```
hwbrzzl, line 341: The compare is weird.

hwbrzzl (commented):
Great, almost complete.

session/session_test.go
```diff
@@ -34,6 +37,51 @@ func (s *SessionTestSuite) TestAll() {
...
+
+	s.session.ageFlashData()
+
+	s.True(s.session.Has("foo"))
+	s.Equal("bar", s.session.Get("foo"))
+	s.Equal(0, s.session.Get("bar"))
+
+	s.session.ageFlashData()
+
+	s.False(s.session.Has("foo"))
```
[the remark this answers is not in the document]
hwbrzzl: It's better to judge all of those.

session/session.go
```diff
@@ -259,11 +259,22 @@ func (s *Session) regenerateToken() sessioncontract.Session {
 }
 
 func (s *Session) removeFromOldFlashData(keys ...string) {
-	old := s.Get("_flash.old", []string{}).([]string)
+	old := s.Get("_flash.old", []any{}).([]any)
 	for _, key := range keys {
-		old = slices.DeleteFunc(old, func(i string) bool {
-			return i == key
+		old = slices.DeleteFunc(old, func(i any) bool {
+			return i == any(key)
```
hwbrzzl, line 265:
```suggestion
return string(i) == key
```

session/session.go
```diff
@@ -259,11 +259,22 @@ func (s *Session) regenerateToken() sessioncontract.Session {
...
 	s.Put("_flash.old", old)
 }
+
+// toStringSlice converts an interface slice to a string slice.
+func toStringSlice(anySlice []any) []string {
+	strSlice := make([]string, len(anySlice))
+	for i, v := range anySlice {
+		if s, ok := v.(string); ok {
+			strSlice[i] = s
+		}
```
hwbrzzl, line 277:
```suggestion
cast.ToString(v)
```

session/session.go
kkumar-gcc: we can't convert `any i` to `string`.

PR #456, opened 2024-04-11
feat: add Collections module

Kamandlou:
Closes https://github.com/goravel/goravel/issues/179

## 📑 Description
add Collections module

## ✅ Checks


support/collection/collection.go
hwbrzzl, line 1: Thanks, Bor.

Can it be like this: `collect.Of(**).Map(**)`? We may need different `Of` for `slice` and `map`? The `Str` package is: `str.Of(**).After()`: https://github.com/goravel/framework/blob/master/support/str/str.go

We can have a method list first before implementing this feature.

Kamandlou: Yes it can be like this `collect.Of(**).Map(**)` but there is a problem, for example when I want to use the Map method on a slice of ints and convert that slice of ints to a slice of strings we faced with a problem

```go
package collection

import "github.com/samber/lo"

// Collection wraps a generic slice to enable method chaining.
type Collection[T any] struct {
	items []T
}

// Of initializes a new Collection instance with the provided items.
func Of[T any](items []T) *Collection[T] {
	return &Collection[T]{items: items}
}

// Map applies a transformation function to each item in the collection and returns a new collection of the results.
func (c *Collection[T]) Map[R any](fn func(T, int) R) *Collection[R] {
	result := lo.Map(c.items, fn)
	return Of[R](result)
}

```
for example:
```go
numbers := Of([]int{1, 2, 3, 4, 5})
strings := numbers.Map(func(n int, _ int) string {
    return fmt.Sprintf("Number: %d", n)
})
fmt.Println(strings)
```

the error with faced with:
 ```bash
./collection.go:16:28: syntax error: method must have no type parameters
```

Comment (hwbrzzl) [reply support/collection/collection.go:1]: Yes, it's terrible, I think we have to give up the `Of` method.

Comment (Kamandlou) [reply support/collection/collection.go:1]: I will provide a list of methods I want to implement and send the list.

Comment (Kamandlou) [reply support/collection/collection.go:1]: and also I will rename the module from `collection` to `collect`

**Comment by Kamandlou** [comment]
The list of methods that I want to implement (new methods will be added to this list later)

1. Filter
2. Sum
3. Max
4. Split
5. Reverse
6. Sum

**Comment by hwbrzzl** [comment]
Could you please check [this list](https://laravel.com/docs/10.x/collections#available-methods) and pick up some useful methods? They can be implemented in multiple PRs.

File: support/collection/collection.go
Comment (kkumar-gcc) [reply support/collection/collection.go:1]: Golang generics won't allow passing type parameters directly. However, I believe we can implement something like `Of` or `Collect`, at least for methods that don't require a function as a parameter.

**Comment by Kamandlou** [comment]
I will add the parallel methods (the callback is called as a goroutine) in another PR.

File: support/collection/collection.go
Comment (Kamandlou) [reply support/collection/collection.go:1]: Method chaining can be implemented for methods that don't require a callback function.

Comment (hwbrzzl) [reply support/collection/collection.go:1]: Yes, we need a consistent way.

**Comment by hwbrzzl** [comment]
Sorry, what's the meaning of `the callback is called as a goroutine`?

**Comment by milwad-dev** [comment]
Why are you using a 3rd party package for collection?

**Comment by Kamandlou** [comment]
> Sorry, what's the meaning of `the callback is called as a goroutine`?

Parallel processing: like Map(), but the mapper function is called in a goroutine. Results are returned in the same order.

check this like: https://github.com/samber/lo#map

**Comment by hwbrzzl** [comment]
> Why are you using a 3rd party package for collection?

It's ok, we don't need to implement this feature from zero.

**Comment by hwbrzzl** [comment]
> > Sorry, what's the meaning of `the callback is called as a goroutine`?
> 
> Parallel processing: like Map(), but the mapper function is called in a goroutine. Results are returned in the same order.
> 
> check this link: https://github.com/samber/lo#map

I see, thanks! But it may be unnecessary for now, we can implement the basic methods first.

**Comment by flc1125** [comment]
`collect` is commonly known as `slices` in go.

**Comment by hwbrzzl** [comment]
@Kamandlou Please let me know if this PR can be reviewed.

**Comment by Kamandlou** [comment]
@hwbrzzl I think this is enough for this PR if I want to add another method I will open another PR.

**Comment by hwbrzzl** [comment]
@Kamandlou Great, will take a look.

File: support/collect/collect_test.go
Comment (hwbrzzl) [inline support/collect/collect_test.go:1]: I notice that the int type is used in all test cases, could you please add the string type test cases for the functions?

**Review by hwbrzzl — COMMENTED** [review]
LGTM, just need to optimize the unit tests.

**Comment by Kamandlou** [comment]
review ready @hwbrzzl :)

File: support/collect/collect_test.go [hunk 1]
```diff
@@ -0,0 +1,145 @@
+package collect
+
+import (
+	"github.com/stretchr/testify/assert"
+
+	"sort"
+	"strconv"
+	"testing"
```
hwbrzzl, line 8: 1. System packages
2. Third packages
3. Self packages

```suggestion
	"sort"
	"strconv"
	"testing"
	
	"github.com/stretchr/testify/assert"
```

hwbrzzl (approved):
Great PR 👍

hwbrzzl:
Hey @Kamandlou  We created a private channel in Discord for contributors. You can click [the link](https://discord.gg/dMs73mSa) and DM me (@Bowen) to join it if you are interested.

Here you can:

1. Send PR link or review others' PRs;
2. Discuss new features before creating an issue;
3. Vote for something;
4. etc.

Kamandlou:
Hi @bowen,

Thank you for inviting me! I’m definitely interested in joining the private channel to collaborate and contribute further to the framework.

PR #493, opened 2024-05-29
fix: Fix all scopes

ahmed3mar:
Closes # 

## 📑 Description
The problem with the initial implementation of the buildScopes method in the QueryImpl struct is that it incorrectly captures the loop variable scope when appending to the gormFuncs slice. This results in all functions in the gormFuncs slice referring to the last scope value in the loop, causing only the last scope to be executed correctly.

This issue arises because the loop variable scope is captured by reference in the anonymous function. When the loop finishes, all the anonymous functions hold a reference to the last value of scope, leading to incorrect behavior.

The solution involves capturing the current value of scope inside the loop by introducing a new variable to hold the current scope. This ensures that each function in the gormFuncs slice captures its own distinct value of scope.

## ✅ Checks


devhaozi (approved):
LGTM

hwbrzzl:
Thanks @ahmed3mar please rebase your branch, will approve again.

PR #507, opened 2024-06-12
fix: translation issue

hwbrzzl:
## 📑 Description

There is an issue when calling `facades.Lang().SetLocale` in middle, `facades.Lang().CurrentLocale()` still returns the default locale in controller.

## ✅ Checks


kkumar-gcc (approved):
LGTM

PR #517, opened 2024-06-25
feat(schedule): optimize log print

devhaozi:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/443

## ✅ Checks

### DIFF
--- a/schedule/application.go
+++ b/schedule/application.go
@@ -39,7 +39,7 @@ func (app *Application) Command(command string) schedule.Event {
 
 func (app *Application) Register(events []schedule.Event) {
 	if app.cron == nil {
-		app.cron = cron.New(cron.WithLogger(NewLogger(app.log, app.logInfo)))
+		app.cron = cron.New(cron.WithLogger(NewLogger(app.log, app.debug)))
 	}
 
 	app.addEvents(events)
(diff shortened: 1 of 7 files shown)


schedule/application.go
```diff
@@ -10,21 +10,22 @@ import (
...
 	"github.com/goravel/framework/support/carbon"
-	"github.com/goravel/framework/support/color"
 )
 
 type Application struct {
 	artisan console.Artisan
 	cache   cache.Cache
 	cron    *cron.Cron
 	log     log.Log
+	logInfo bool
```
hwbrzzl, line 20:
```suggestion
	debug bool
```

hwbrzzl (commented):
LGTM, just optimize the variable name.

schedule/application_test.go
```diff
@@ -37,7 +41,7 @@ func (s *ApplicationTestSuite) TestCallAndCommand() {
 	delayIfStillRunningCall := 0
 	skipIfStillRunningCall := 0
 
-	app := NewApplication(mockArtisan, nil, mockLog)
+	app := NewApplication(mockArtisan, nil, mockLog, mockConfig.GetBool("app.debug"))
```
kkumar-gcc, line 44: S: Do we really need to use `mockConfig`? We can pass `false` directly since we are not testing the config.

kkumar-gcc (approved):
LGTM

PR #449, opened 2024-03-27
feat: Add Session Module 3

kkumar-gcc:
Closes: https://github.com/goravel/goravel/issues/357
Related PR: #428, #436 

## 📑 Description
add `start_session` middleware

## ✅ Checks

- [x] My pull request adheres to the code style of this project
- [x] My code requires changes to the documentation
- [ ] I have updated the documentation as required
- [ ] All the tests have passed

PR #581, opened 2024-08-03
feat: [#449] Implement HTTP response stream 

kkumar-gcc:
## 📑 Description

RelatedTo https://github.com/goravel/goravel/issues/449

@coderabbitai summary

## ✅ Checks

### DIFF
--- a/contracts/http/response.go
+++ b/contracts/http/response.go
@@ -34,7 +34,7 @@ type ContextResponse interface {
 	// Status sets the HTTP response status code and returns the ResponseStatus.
 	Status(code int) ResponseStatus
 	// Stream sends a streaming response with the specified status code and the given reader.
-	Stream(step func(w StreamWriter) error, code int, contentType string) Response
+	Stream(step func(w StreamWriter) error, code int) Response
 	// View returns ResponseView
 	View() ResponseView
 	// Writer returns the underlying http.ResponseWriter associated with the response.
--- a/contracts/http/response.go
+++ b/contracts/http/response.go
@@ -34,7 +34,7 @@ type ContextResponse interface {
 	// Status sets the HTTP response status code and returns the ResponseStatus.
 	Status(code int) ResponseStatus
 	// Stream sends a streaming response with the specified status code and the given reader.
-	Stream(step func(w StreamWriter) error, code int) Response
+	Stream(code int, step func(w StreamWriter) error) Response
 	// View returns ResponseView
 	View() ResponseView
 	// Writer returns the underlying http.ResponseWriter associated with the response.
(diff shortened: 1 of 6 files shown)


contracts/http/response.go
```diff
@@ -34,7 +34,7 @@ type ContextResponse interface {
 	// Status sets the HTTP response status code and returns the ResponseStatus.
 	Status(code int) ResponseStatus
 	// Stream sends a streaming response with the specified status code and the given reader.
-	Stream(step func(w StreamWriter) error, code int, contentType string) Response
+	Stream(step func(w StreamWriter) error, code int) Response
```
hwbrzzl, line 37: Keep consistent with the `Json` method, `code` is the first paramter.

```suggestion
	Stream(code int, step func(w StreamWriter) error) Response
```

And add the `Stream` to `ResponseStatus`:

```
Stream(step func(w StreamWriter) error) Response
```

hwbrzzl (commented):
I ran this feature locally, the strings do not be printed one by one, did I miss something?

```
facades.Route().Get("/stream", func(ctx http.Context) http.Response {
  b := []string{"a", "b", "c"}
  return ctx.Response().Stream(func(w http.StreamWriter) error {
    for _, a := range b {
      if _, err := w.Write([]byte(a)); err != nil {
	return err
      }
      
      if err := w.Flush(); err != nil {
	return err
      }
      
      time.Sleep(1 * time.Second)
    }
    
    return nil
  }, 200)
})
```

kkumar-gcc:
> I ran this feature locally, the strings do not be printed one by one, did I miss something?
> 
> ```
> facades.Route().Get("/stream", func(ctx http.Context) http.Response {
>   b := []string{"a", "b", "c"}
>   return ctx.Response().Stream(func(w http.StreamWriter) error {
>     for _, a := range b {
>       if _, err := w.Write([]byte(a)); err != nil {
> 	return err
>       }
>       
>       if err := w.Flush(); err != nil {
> 	return err
>       }
>       
>       time.Sleep(1 * time.Second)
>     }
>     
>     return nil
>   }, 200)
> })
> ```


Buffered content only be displayed after reaching to **newline**.


```go
...
if _, err := w.Write([]byte(a + "\n")); err != nil {
     return err
}
...
```

Ref: https://stackoverflow.com/a/1716621/17666299

contracts/http/response.go
kkumar-gcc: Ok

hwbrzzl:
@kkumar-gcc I changed it, but doesn't have an effect.

```
facades.Route().Get("/stream", func(ctx http.Context) http.Response {
		b := []string{"a", "b", "c", "c", "c", "c", "c"}
		return ctx.Response().Stream(func(w http.StreamWriter) error {
			for _, a := range b {
				if _, err := w.Write([]byte(a + "\n")); err != nil {
					return err
				}

				if err := w.Flush(); err != nil {
					return err
				}

				time.Sleep(1 * time.Second)
			}

			return nil
		}, 200)
	})
```

hwbrzzl (approved):
Great

PR #583, opened 2024-08-06
fix: lang path error

hwbrzzl:
## Summary by CodeRabbit

- **New Features**
	- Enhanced the flexibility of the language path configuration, allowing it to be set through application settings rather than a static value.

- **Bug Fixes**
	- Improved the reliability of the `LangPath` method's functionality through updated testing practices, ensuring correct path construction based on configuration.

- **Tests**
	- Revamped the testing setup for the `LangPath` method, incorporating mock configurations for better accuracy and isolation in tests.

## ✅ Checks


support/path/path_test.go
hwbrzzl, line 1: The path methods call app methods directly, there is no logic, this file is unnecessary.

kkumar-gcc (approved):
LGTM

PR #406, opened 2024-02-01
feat: Add WhereHas Method to ORM

sCuz12:
Closes # 

## 📑 Description
This PR introduces a feature inspired by Laravel's ORM: the WhereHas method. The WhereHas method allows users to efficiently query records based on the existence of related records and apply conditions to those **related** records.

#### Why it is important ?

- Simplified Complex Queries: With WhereHas, developers can easily perform complex queries involving relationships without the need for multiple database queries or manual filtering.
- Consistency: Aligning with Laravel's ORM conventions makes it easier for developers familiar with Laravel to work with this ORM.

### Example
Suppose you have a database with two main models: User and Book, and there is a one-to-many relationship between users and books, where each user can have multiple books. You want to find users who have books with a specific condition, such as books with the word "fantasy" in their names.

``` 
query.WhereHas("books", func(subQuery ormcontract.Query) ormcontract.Query  {
    // Apply conditions to the related "books" model.
    subQuery.Where("Name", "like", "%fantasy%")
})
``` 

## ✅ Checks

- [ ] My pull request adheres to the code style of this project
- [ ] My code requires changes to the documentation
- [ ] I have updated the documentation as required
- [x ] All the tests have passed

PR #602, opened 2024-08-16
feat: add WhereHas method

shayan-yousefi:
## 📑 Description

hey there, in this PR I want to add the WhereHas method.
I welcome any suggestions.
https://github.com/goravel/goravel/issues/373

@coderabbitai summary

## ✅ Checks


shayan-yousefi:
While working on the code, I noticed that there are no query methods that return errors, nor are there any struct fields like Gorm's Error field. However, errors could potentially occur in these methods. Do you have any suggestions on how we might handle this?
I think we can have an Error field like Gorm
or panic in errors (I think panic is not a good approach)

The PR is not yet complete, so if you have any other suggestions or ideas to improve it, I'd love to hear them.
Thanks!
@hwbrzzl

kkumar-gcc:
There is a closed PR related to this feature: https://github.com/goravel/framework/pull/406.

hwbrzzl:
Do you check the Gorm implementation? https://gorm.io/docs/preload.html

Can we use that way to implement WhereHas?

shayan-yousefi:
> Do you check the Gorm implementation? https://gorm.io/docs/preload.html
> 
> Can we use that way to implement WhereHas?

I am familiar with GORM's Preload, but I don't think it's the right solution for us. Preload performs eager loading, which means that even if no related records are found, it will still fetch rows from the original table, but without any related rows. Even if we pass a query function to Preload, it only modifies the select statement for eager loading; the original table rows will still be returned without the related rows.

I didn't find whereHas in form like laravel, may I be wrong?

@hwbrzzl

hwbrzzl:
@shayan-yousefi Sorry for the delay, I missed this message. Yes, Gorm doesn't have the WhereHas method, and your explanation about Preload is right. How about the Joins Preloading of Gorm: https://gorm.io/docs/preload.html#Joins-Preloading? Can it be used?

shayan-yousefi:
> @shayan-yousefi Sorry for the delay, I missed this message. Yes, Gorm doesn't have the WhereHas method, and your explanation about Preload is right. How about the Joins Preloading of Gorm: https://gorm.io/docs/preload.html#Joins-Preloading? Can it be used?

> NOTE: Join Preload works with one-to-one relationships, like has one and belongs to.

Based on the documentation, it's only compatible with has one and belongs to relationships.

One thing to keep in mind is that it limits query statements to using JOIN, which might make writing more complex queries a bit tricky. Also, the GORM documentation mentions that it uses a LEFT JOIN, so it might return rows even if the related rows aren't found.

It seems like Laravel's whereHas might use a nested query since it can handle counts to return rows if a model has a certain number of related rows.

I get the feeling you're not a fan of using reflect. Honestly, neither am I, but I haven't found a better way yet.

If you have a better idea, I'd love to hear it! 🙏 Or if something similar has been implemented before in Goravel, please point me in the right direction so I can get some inspiration.

hwbrzzl:
@shayan-yousefi Thanks for your explanation, basically, I want to use the ready-made methods to reduce the complicated logic, but if there are no other ways, the current logic in this PR is pretty good, we need to find the foreign key and primary key via reflect. And about the error, we can add a private variable to save the error that is returned by `WhereHas`, and return it when calling the final method, like: Get, First, etc.

shayan-yousefi:
> @shayan-yousefi Thanks for your explanation, basically, I want to use the ready-made methods to reduce the complicated logic, but if there are no other ways, the current logic in this PR is pretty good, we need to find the foreign key and primary key via reflect. And about the error, we can add a private variable to save the error that is returned by `WhereHas`, and return it when calling the final method, like: Get, First, etc.

After some research, I found that we can use gorm.Statement{} by passing the db object to it. By then calling the Parse method with the model, it parses the schema, enabling us to work with it directly without needing to manually handle reflection.
@hwbrzzl

hwbrzzl:
Awesome 👍

shayan-yousefi:
I found that we can use preload with subqueries. I'll implement this approach whenever possible

shayan-yousefi:
I find gorm support composit indexes. Are we want to support it?
If yes, I have to think in another way
@hwbrzzl

hwbrzzl:
@shayan-yousefi Good catch, I think we don't need to support it for now.

shayan-yousefi:
I find gorm support composit indexes. Are we want to support it?
If yes, I have to think in another way
@hwbrzzl 

This PR is done. But I can't run tests to start writing tests. I fix mocks. The issue is that docker containers in test cases always have connection refused error, and I can't run tests to writing them. Am I wrong in running tests, or should I do anything before?

hwbrzzl:
connection refused is normal, after about 60 seconds, the error should disappear. I have a PR to optimize the test process, you can merge it later. https://github.com/goravel/framework/pull/612

hwbrzzl:
I can check the basic logic before you write test cases.

hwbrzzl:
@shayan-yousefi What system are you using?

shayan-yousefi:
> @shayan-yousefi What system are you using?

what do you mean, system?
if you mean os, I use Ubuntu 24.04

hwbrzzl:
Yes, os, got it.

shayan-yousefi:
> I can check the basic logic before you write test cases.

@hwbrzzl thanks, Are you ok with that?
about tests, I think the docker containers do not run correctly, the test fails because it cant never connect to the containers

hwbrzzl:
Yes, Docker should be run. And sorry, I'm on vacation this week, I'll check this PR when I get a chance.

database/gorm/query.go
```diff
@@ -780,6 +866,82 @@ func (r *QueryImpl) Where(query any, args ...any) ormcontract.Query {
 	return r.setConditions(conditions)
 }
 
+func (r *QueryImpl) WhereHas(relation string, callback func(query ormcontract.Query) (ormcontract.Query, error), args ...any) ormcontract.Query {
```
hwbrzzl, line 869: This method needs to be implemented like other methods, saving the conditions to r.conditions first, and adding a new `buildWhereHas` method in the `buildConditions` method.

database/gorm/query.go
```diff
@@ -780,6 +866,82 @@ func (r *QueryImpl) Where(query any, args ...any) ormcontract.Query {
...
+		return r.setError(fmt.Errorf("faild to get model schema, the model should be set before using this method. %w", err))
+	}
+
+	modifiedQuery, err := callback(r.new(r.instance))
+	if err != nil {
+		return r.setError(err)
+	}
+	rel, ok := mSchema.Relationships.Relations[relation]
+	if !ok {
+		r.setError(errors.New("the relation not found"))
```
hwbrzzl, line 909:
```suggestion
		r.setError(fmt.Errof("the relation not found: %s", relation))
```

database/gorm/query.go
```diff
@@ -780,6 +866,82 @@ func (r *QueryImpl) Where(query any, args ...any) ormcontract.Query {
...
+	needCountQuery := !((count == 0 && slices.Contains([]string{lt, lte, gt, gte}, op)) || op == "")
+	if !needCountQuery {
+		modifiedQueryImpl := modifiedQuery.Model(relModel).(*QueryImpl)
+		return r.Where("EXISTS (?)", modifiedQueryImpl.buildConditions().instance)
+	}
+
+	modifiedQueryImpl := modifiedQuery.Model(relModel).Select("count(*)").(*QueryImpl)
+	return r.Where("(?) "+op+" ?", modifiedQueryImpl.buildConditions().instance, count)
+
+}
```
hwbrzzl, line 943: Do you test the logic locally? Is there a screenshot, please? You can add it in the PR description.

hwbrzzl (commented):
Great PR, a few questions 👍

hwbrzzl:
I optimized the docker test experience, please rebase your branch.

shayan-yousefi:
@hwbrzzl I will read your reviews. And fix them. Thanks

hwbrzzl:
Hey @shayan-yousefi Are you still working on this, please?

hwbrzzl:
Closed this PR first, feel free to open it when needed.

PR #612, opened 2024-08-23
feat: optimize docker test

hwbrzzl:
## 📑 Description

1. Implement basic facades.Schema logic, I found the docker test needs to be optimized when implementing facades.Schema, and it only contains a few codes, so I add them into this PR directly.
2. Optimize the docker test experience, reduce the Linux test from ~7m to ~3m. Currently, we can use `supportdocker.Mysql()` to get a Mysql Docker instance directly, and the instance will be cached, every test case that uses the docker instance can get the singleton instance.

## ✅ Checks


devhaozi (approved):
LGTM

PR #619, opened 2024-09-02
fix: bootArtisan should be called last, otherwise the settings after that will not take effect in artisan console

zxdstyle:
…will not take effect in artisan console

## 📑 Description

@coderabbitai summary

## ✅ Checks


hwbrzzl:
Good catch 👍  About the failed Lint CI, you can rebase the v1.14.x branch, I solved it.

PR #622, opened 2024-09-04
feat: some optimize

devhaozi:
## 📑 Description

Closes goravel/goravel/issues/480
Closes goravel/goravel/issues/481
Closes goravel/goravel/issues/325

## Summary by CodeRabbit

- **New Features**
	- Introduced a benchmarking workflow for Go applications to monitor performance during CI/CD processes.
	- Added a vulnerability check workflow for Go modules to enhance security.
	- Improved session management with a new session pool for efficient reuse of session objects.

- **Chores**
	- Updated Go module version from 1.21 to 1.22 for improved compatibility and features.

## ✅ Checks


session/manager.go
```diff
@@ -44,30 +62,58 @@ func (m *Manager) Driver(name ...string) (sessioncontract.Driver, error) {
...
+		return nil, fmt.Errorf("driver [%s] not supported", driverName)
 	}
 
 	return m.drivers[driverName], nil
 }
 
-func (m *Manager) Extend(driver string, handler func() sessioncontract.Driver) sessioncontract.Manager {
-	m.customDrivers[driver] = handler()
-	return m
+func (m *Manager) Extend(driver string, handler sessioncontract.Driver) (sessioncontract.Manager, error) {
```
kkumar-gcc, line 71: We shouldn't change it. It was intentionally of the `func()` type so that the user can perform some tasks when registering it.

session/manager.go
```diff
@@ -44,30 +62,58 @@ func (m *Manager) Driver(name ...string) (sessioncontract.Driver, error) {
...
+	if m.drivers[driver] != nil {
```
kkumar-gcc, line 72: User can overwrite existing driver.

devhaozi: > User can overwrite existing driver.

This seems not suitable with Go's habits, and overwriting the driver may cause GC exceptions.

session/manager.go
```diff
@@ -44,30 +62,58 @@ func (m *Manager) Driver(name ...string) (sessioncontract.Driver, error) {
...
+		return nil, fmt.Errorf("driver [%s] already exists", driver)
+	}
+	m.drivers[driver] = handler
+	m.startGcTimer(m.drivers[driver])
```
kkumar-gcc, line 76: How will this `GcTimer` start for the built-in driver in Goravel?

session/manager.go
kkumar-gcc: Hmm, makes sense. It's okay to merge custom and built-in drivers.

session/manager.go
devhaozi: > How will this `GcTimer` start for the built-in driver in Goravel?

```
func (m *Manager) extendDefaultDriver() {
	if _, err := m.Extend("file", m.createFileDriver); err != nil {
		panic(fmt.Sprintf("failed to extend session manager: %v", err))
	}
}

func (m *Manager) createFileDriver() sessioncontract.Driver {
	lifetime := m.config.GetInt("session.lifetime")
	return driver.NewFile(m.config.GetString("session.files"), lifetime)
}
```

built-in driver also use `Extend` to register, so GC will start when register the driver.

kkumar-gcc: Oh, okay, sorry, I didn't see the updated code. But how will we handle multiple built-in drivers?

session/manager.go
```diff
@@ -2,33 +2,51 @@ package session
...
+		},
 	}
-	manager.registerDrivers()
+	manager.extendDefaultDrivers()
 	return manager
 }
 
 func (m *Manager) BuildSession(handler sessioncontract.Driver, sessionID ...string) sessioncontract.Session {
-	return NewSession(m.config.GetString("session.cookie"), handler, m.json, sessionID...)
+	session := m.AcquireSession()
```
kkumar-gcc, line 39: We shouldn't directly access properties from the session. Could you please use methods instead of accessing session.name directly?

session/manager.go
devhaozi: > Oh, okay, sorry, I didn't see the updated code. But how will we handle multiple built-in drivers?

maybe:

```
func (m *Manager) extendDefaultDrivers() {
	if _, err := m.Extend("file", m.createFileDriver); err != nil {
		panic(fmt.Sprintf("failed to extend session file driver: %v", err))
	}
	if _, err := m.Extend("file", m.createDatabaseDriver); err != nil {
		panic(fmt.Sprintf("failed to extend session database driver: %v", err))
	}
}

func (m *Manager) createFileDriver() sessioncontract.Driver {
	lifetime := m.config.GetInt("session.lifetime")
	return driver.NewFile(m.config.GetString("session.files"), lifetime)
}

func (m *Manager) createDatabaseDriver() sessioncontract.Driver {
	// TODO
}

But I think only need to implement file driver for default, users may not use database, if they need database driver they can add it manualy.
```

kkumar-gcc: This is controlled by `config/session.go`. So if the user has the `driver` set to the `database`, we should extend the database driver.

contracts/session/manager.go
```diff
@@ -6,5 +6,5 @@ type Manager interface {
 	// Driver retrieves the session driver by name.
 	Driver(name ...string) (Driver, error)
 	// Extend extends the session manager with a custom driver.
-	Extend(driver string, handler func() Driver) Manager
+	Extend(driver string, handler func() Driver) (Manager, error)
```
kkumar-gcc, line 9: If we are returning an error, we don't need to return the manager because chaining won't make sense to the user anyway.

devhaozi: > If we are returning an error, we don't need to return the manager because chaining won't make sense to the user anyway.

makes sense.

devhaozi: And I think `BuildSession` can directly pass driver name instead of handler?

kkumar-gcc: Hmm, we can pass the driver name directly but need to return an error as the second argument. Also, I believe we don't need the Driver method anymore.

cc: @hwbrzzl

devhaozi:
@hwbrzzl nilaway returned too many errors, are you sure you want to add it?

hwbrzzl:
@devhaozi Yes, we can add it, and solve the errors in another PR.

session/middleware/start_session.go
```diff
@@ -50,8 +50,12 @@ func StartSession() http.Middleware {
...
 
 		// Save session
-		if err := s.Save(); err != nil {
+		if err = s.Save(); err != nil {
 			color.Red().Printf("Error saving session: %s\n", err)
 		}
+
+		// Release session
+		// TODO - any better way to release the session?
+		session.SessionFacade.(*session.Manager).ReleaseSession(s.(*session.Session))
```
hwbrzzl, line 59: Add the `ReleaseSession` method to interface.

```suggestion
		session.SessionFacade.ReleaseSession(s)
```

hwbrzzl (commented):
Many optimizations, Great PR 👍  But it's a bit hard to review, please split it into some small PRs next time.

log/logger/daily.go
```diff
@@ -43,6 +44,7 @@ func (daily *Daily) Handle(channel string) (logrus.Hook, error) {
 		logPath+"-%Y-%m-%d"+ext,
 		rotatelogs.WithRotationTime(time.Duration(24)*time.Hour),
 		rotatelogs.WithRotationCount(uint(daily.config.GetInt(channel+".days"))),
+		rotatelogs.WithClock(rotatelogs.NewClock(carbon.Now().StdTime())),
```
devhaozi, line 47: This is a bug fix.

.github/workflows/benchmark.yml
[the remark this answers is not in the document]
devhaozi: > Good idea, and so complex. Where can we check the result, please? And can we move it to goravel/.github?

See https://github.com/benchmark-action/github-action-benchmark
We can create a GitHub Page to display the benchmark results.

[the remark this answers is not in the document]
hwbrzzl: Do you have a screenshot for this? We can check if it's meaning to make a Github gage.

hwbrzzl:
And we can remove this line:

image

We also need to change the driver name for goravel/goravel.

.github/workflows/benchmark.yml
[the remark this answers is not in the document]
devhaozi: > Do you have a screenshot for this? We can check if it's meaning to make a Github gage.

No, this workflow file is copied from gofiber, I don't think we need to show the benchmark results.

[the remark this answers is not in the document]
devhaozi: > Good idea, and so complex. Where can we check the result, please? And can we move it to goravel/.github?

Yes, will move to .github and optimize the process.

devhaozi:
> FYI, please pin me when this PR is ready.

Already prepared.

.github/workflows/benchmark.yml
hwbrzzl, line 1: Em, if we can‘t check the result, what's the meaning of it?

session/manager.go
```diff
@@ -44,30 +65,63 @@ func (m *Manager) Driver(name ...string) (sessioncontract.Driver, error) {
...
+func (m *Manager) Extend(driver string, handler func() sessioncontract.Driver) error {
+	if m.drivers[driver] != nil {
+		return fmt.Errorf("driver [%s] already exists", driver)
+	}
+	m.drivers[driver] = handler()
+	m.startGcTimer(m.drivers[driver])
+	return nil
+}
+
+func (m *Manager) AcquireSession() *Session {
```
hwbrzzl, line 83: Sort it according to the first name.

hwbrzzl (commented):
Great, almost complete, only two questions.

.github/workflows/benchmark.yml
devhaozi: > Em, if we can‘t check the result, what's the meaning of it?

Move the test cases previously used for concurrency testing to the benchmark.
And although the results cannot be viewed directly, CI will automatically compare them with previous results.

hwbrzzl: > CI will automatically compare them with previous results.

Do you mean this?

image

devhaozi: > > CI will automatically compare them with previous results.
> 
> Do you mean this?
> 
> image

Yes, currently no data.

hwbrzzl (approved):
Perfect 👍 Please remember to create a PR for goravel/goravel.

PR #650, opened 2024-09-20
feat: GetDeep function to support nested map lookups

kadirgun:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/498

@coderabbitai summary

## ✅ Checks

### DIFF
--- a/support/maps/maps.go
+++ b/support/maps/maps.go
@@ -88,6 +88,36 @@ func Get[K comparable, V any](mp map[K]V, key K, defaults ...V) V {
 	return val
 }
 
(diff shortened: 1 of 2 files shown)


hwbrzzl (changes requested):
Thanks for your PR, and yes, it can indeed be implemented by Reflect. 

But it contains many `any`, we have to assert them. In actual use, the structure (map[string]map[string]map[string]any) rarely, and doesn't recommend. We are using  generics to implement other methods, so I think the `GetDeep` method is not very good for us. 

Anyway, I appreciate your PR, Welcome any thoughts.

PR #653, opened 2024-09-24
feat: extension the Create method of Orm

hwbrzzl:
## 📑 Description

The `Create` method of Orm supports map:

```
// Not trigger model events
facades.Orm().Query().Table("users").Create(map[string]any{
  "name": "Goravel",
})

// Trigger model events
facades.Orm().Query().Model(&models.User{}).Create(map[string]any{
  "name": "Goravel",
})
```

## Summary by CodeRabbit

- **New Features**
	- Enhanced testing capabilities for attribute setting and retrieval using different data structures (maps and structs).
	- Expanded event handling in the `User` struct to accommodate various event names and set corresponding attributes.

- **Bug Fixes**
	- Improved test coverage for user creation scenarios, ensuring correct handling of IDs and timestamps.

- **Refactor**
	- Simplified method signatures and improved clarity in the `QueryImpl` struct methods.

- **Tests**
	- Added new test cases and modified existing ones to cover more scenarios and ensure expected behavior.

## ✅ Checks

### DIFF
--- a/database/gorm/test_models.go
+++ b/database/gorm/test_models.go
@@ -34,8 +34,20 @@ func (u *User) DispatchesEvents() map[ormcontract.EventType]func(ormcontract.Eve
 				if name.(string) == "event_creating_name" {
 					event.SetAttribute("avatar", "event_creating_avatar")
 				}
+				if name.(string) == "event_creating_by_map_name" {
+					event.SetAttribute("avatar", "event_creating_by_map_avatar1")
+				}
 				if name.(string) == "event_creating_FirstOrCreate_name" {
 					event.SetAttribute("avatar", "event_creating_FirstOrCreate_avatar")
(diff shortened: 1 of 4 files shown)


database/gorm/test_models.go
```diff
@@ -34,9 +34,21 @@ func (u *User) DispatchesEvents() map[ormcontract.EventType]func(ormcontract.Eve
 				if name.(string) == "event_creating_name" {
```
kkumar-gcc, line 34: Instead of this, would it be better to cast it? Because if this is not asserted, it will panic.

hwbrzzl: Sure, I'll optimize it in another PR.

kkumar-gcc (approved):

PR #667, opened 2024-10-03
fix: nilaway

hwbrzzl:
## 📑 Description

@coderabbitai summary

## ✅ Checks

### DIFF
--- a/session/session.go
+++ b/session/session.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/goravel/framework/contracts/foundation"
 	sessioncontract "github.com/goravel/framework/contracts/session"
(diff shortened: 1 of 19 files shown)


session/session.go
```diff
@@ -221,24 +220,13 @@ func (s *Session) loadSession() {
 	}
 }
 
-func (s *Session) validateDriver() error {
-	if s.driver == nil {
```
hwbrzzl, line 225: No this situation.

kkumar-gcc: Please check session.Manager

kkumar-gcc (approved):
LGTM

PR #709, opened 2024-11-02
feat: [#280] The migartion feature of v1.15 doesn't support change field

hwbrzzl:
## 📑 Description

The `Change` function is a bit difficult, different DBs and different versions of the same DB have different actions, and event Sqlite doesn't support change field, needs to add a new field to a temporary table, remove the original table, then rename the temporary table to the original.

We will spend several months on this feature, so I think we can implement it in v1.16, so that v1.15 can be released at the end of the year.

In v1.15, User can use the migration feature to create a table, add a new field to a table, run a script in the migration file. If user want to alter a field, he can use the `facades.Schema().Sql()` method, to execute an original SQL.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new method `GetCommands` to manage commands associated with blueprints.
	- Added constants for command names and default string lengths to standardize usage across the application.

- **Bug Fixes**
	- Simplified logic in various methods to enhance clarity and functionality, particularly in command handling and column modifications.

- **Documentation**
	- Updated test cases to reflect changes in method signatures and removed outdated tests.

- **Refactor**
	- Streamlined interfaces and methods by removing unnecessary parameters and methods, focusing on essential functionalities.

## ✅ Checks


database/schema/blueprint.go
```diff
@@ -56,36 +48,41 @@ func (r *Blueprint) Build(query ormcontract.Query, grammar schema.Grammar) error
...
 }
 
 func (r *Blueprint) GetAddedColumns() []schema.ColumnDefinition {
 	var columns []schema.ColumnDefinition
 	for _, column := range r.columns {
-		if column.change == nil || !*column.change {
-			columns = append(columns, column)
-		}
+		columns = append(columns, column)
+		//if column.change == nil || !*column.change {
```
hwbrzzl, line 65: Didn't remove it, because it can be used in v1.16.

database/schema/blueprint.go
```diff
@@ -56,36 +48,41 @@ func (r *Blueprint) Build(query ormcontract.Query, grammar schema.Grammar) error
...
 }
 
-func (r *Blueprint) GetChangedColumns() []schema.ColumnDefinition {
-	var columns []schema.ColumnDefinition
-	for _, column := range r.columns {
-		if column.change != nil && *column.change {
-			columns = append(columns, column)
-		}
-	}
+//func (r *Blueprint) GetChangedColumns() []schema.ColumnDefinition {
```
hwbrzzl, line 73: Ditto

database/schema/constants/constants.go
```diff
@@ -0,0 +1,10 @@
+package constants
```
kkumar-gcc, line 1: Wouldn't it be better to add a `constants` directory at the root and create separate files for different types of constants?

hwbrzzl: I think it's ok, the current constants are only used by `schema`.

kkumar-gcc (approved):
LGTM

PR #749, opened 2024-12-07
feat: use goravel/machinery

devhaozi:
## 📑 Description

@coderabbitai summary

## ✅ Checks


devhaozi:
The logger needs to be refactor. I suggest using the std `slog`. What do you think?

PR #751, opened 2024-12-11
fix: [#533] there is a bug when validating nested data

hwbrzzl:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/533

@coderabbitai summary

## ✅ Checks

### DIFF
--- a/support/docker/sqlserver.go
+++ b/support/docker/sqlserver.go
@@ -28,22 +28,19 @@ func NewSqlserver(database, username, password string) *Sqlserver {
 		username: username,
 		password: password,
 		image: &testing.Image{
-			Repository: "mcmoe/mssqldocker",
+			Repository: "mcr.microsoft.com/mssql/server",
 			Tag:        "latest",
 			Env: []string{
 				"ACCEPT_EULA=Y",
-				"MSSQL_DB=" + database,
-				"MSSQL_USER=" + username,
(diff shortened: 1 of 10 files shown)


validation/validator.go
```diff
@@ -129,19 +129,6 @@ func (v *Validator) castValue() mapstructure.DecodeHookFunc {
 			default:
 				castedValue, err = cast.ToSliceE(from.Interface())
 			}
-		case reflect.Map:
```
hwbrzzl, line 132: The `map` structure is not fixed, so we can cast the value to `StringMapString`, `StringMapBool`, etc. And there is a BUG as well, `to.Type().Key().Kind()` checks the map key, but in the `case` logic, we are using `ToStringMapBoolE`, it should be `ToBoolMap***E`. 

I think we can remove the logic directly.

support/docker/sqlserver.go
```diff
@@ -28,22 +28,19 @@ func NewSqlserver(database, username, password string) *Sqlserver {
 		username: username,
 		password: password,
 		image: &testing.Image{
-			Repository: "mcmoe/mssqldocker",
+			Repository: "mcr.microsoft.com/mssql/server",
```
hwbrzzl, line 31: The mssql image is unavailable, the CI failed, so modify it in this PR.

PR #753, opened 2024-12-12
fix: [#531] An extra log will be printed when using a custom log driver

hwbrzzl:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/531

Tested locally, it's hard to add a  test case:

image

@coderabbitai summary

## ✅ Checks

### DIFF
--- a/log/logrus_writer.go
+++ b/log/logrus_writer.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
 
 	"github.com/rotisserie/eris"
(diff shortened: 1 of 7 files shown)


log/logrus_writer.go
```diff
@@ -343,8 +357,6 @@ func registerHook(config config.Config, json foundation.Json, instance *logrus.L
 		return errors.New("Error logging channel: " + channel)
 	}
 
-	instance.SetFormatter(formatter.NewGeneral(config, json))
```
hwbrzzl, line 346: The main error is that this line is added to all channels.

PR #282, opened 2023-08-31
fix: will panic when model contains a private field

hwbrzzl:
Closes # 

## ✅ Checks

- [x] My pull request adheres to the code style of this project
- [ ] My code requires changes to the documentation
- [ ] I have updated the documentation as required
- [x] All the tests have passed

PR #759, opened 2024-12-14
feat: [#282] Print the values in ctx when calling facades.Log().WithContext(ctx).Info()

hwbrzzl:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/282

image

## Summary by CodeRabbit

- **New Features**
  - Enhanced logging capabilities with new fields: `code`, `domain`, and `hint`.
  - Introduced a new `Hook` type for improved integration with the logging framework.
  - Added utility function `getContextValues` for extracting key-value pairs from context structures.
  - Modified the `WithContext` method to handle HTTP-specific contexts.
  - Improved data formatting with additional keys processed in logs.

- **Bug Fixes**
  - Improved error handling in log entry processing.

- **Tests**
  - Added comprehensive unit tests for the new `Hook` functionality and context value extraction.
  - Updated existing tests to reflect changes in log entry formatting and context handling.

- **Chores**
  - Streamlined logging assertions and context management for clarity and efficiency.

## ✅ Checks

### DIFF
--- a/log/formatter/general.go
+++ b/log/formatter/general.go
@@ -82,14 +82,9 @@ func (general *General) formatData(data logrus.Fields) (string, error) {
 			return "", err
 		}
 
-		for _, key := range []string{"code", "context", "domain", "hint", "owner", "request", "response", "tags", "user"} {
+		for _, key := range []string{"code", "context", "with", "domain", "hint", "owner", "request", "response", "tags", "user"} {
 			if value, exists := root[key]; exists && value != nil {
-				v, err := general.json.Marshal(value)
-				if err != nil {
-					return "", err
-				}
-
-				builder.WriteString(fmt.Sprintf("%s: %v\n", key, string(v)))
+				builder.WriteString(fmt.Sprintf("%s: %+v\n", key, value))
 			}
 		}
 
--- a/log/formatter/general.go
+++ b/log/formatter/general.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/goravel/framework/contracts/config"
 	"github.com/goravel/framework/contracts/foundation"
+	"github.com/goravel/framework/support/str"
 )
 
 type General struct {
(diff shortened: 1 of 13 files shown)


log/formatter/general.go
```diff
@@ -82,14 +82,9 @@ func (general *General) formatData(data logrus.Fields) (string, error) {
 			return "", err
 		}
 
-		for _, key := range []string{"code", "context", "domain", "hint", "owner", "request", "response", "tags", "user"} {
+		for _, key := range []string{"code", "context", "with", "domain", "hint", "owner", "request", "response", "tags", "user"} {
 			if value, exists := root[key]; exists && value != nil {
-				v, err := general.json.Marshal(value)
```
hwbrzzl, line 87: Not all values can be marshaled to json.

log/logrus_writer.go
```diff
@@ -333,7 +307,6 @@ func registerHook(config config.Config, json foundation.Json, instance *logrus.L
 
 		if config.GetBool(channelPath + ".print") {
 			instance.SetOutput(os.Stdout)
-			instance.SetFormatter(formatter.NewGeneral(config, json))
```
hwbrzzl, line 336: Had set this in hook.

log/logrus_writer_test.go
```diff
@@ -298,11 +297,11 @@ func TestLogrus(t *testing.T) {
 			assert: func() {
 				expectedParts := []string{
 					`test.info: Goravel`,
-					`response: {`,
-					`"status":200`,
-					`"header":{"Content-Type":["text/plain; charset=utf-8"]}`,
-					`"body":{}`,
```
hwbrzzl, line 304: @kkumar-gcc The `body` is missing when usnig `json.Marshal`, and `map[any]any` can't be marshaled, golang throws an error.

kkumar-gcc: @hwbrzzl I tested it, and the body is not missing, but `map[any]any` cannot be parsed.

image

kkumar-gcc: @hwbrzzl I believe the reason the body is missing is that we should use `res.Origin().Body().String()` because `res.Origin().Body()` is a buffer.
However, `map[any]any` cannot be parsed.

hwbrzzl: Yes, so we don't need to use `json.Marshal`, printing them should be fine.

kkumar-gcc: Well, it will include `map` and other type information, which I’m not sure is a good idea to include in the logs.

hwbrzzl: Yes, it's fine, I prefer to know the specific type.

PR #762, opened 2024-12-16
feat: add listen method

devhaozi:
## 📑 Description

For issue https://github.com/goravel/goravel/issues/436

@coderabbitai summary

## ✅ Checks

### DIFF
--- a/contracts/route/route.go
+++ b/contracts/route/route.go
@@ -26,6 +26,8 @@ type Route interface {
 	ServeHTTP(writer http.ResponseWriter, request *http.Request)
 	// Listen starts the HTTP server and listens for incoming connections on the specified listener.
 	Listen(l net.Listener) error
+	// ListenTLS starts the HTTPS server with the provided TLS configuration and listens on the specified listener.
+	ListenTLS(l net.Listener, certFile, keyFile string) error
 	// Stop gracefully stop the serve.
 	Stop(ctx ...context.Context) error
 	// Test method to simulate HTTP requests (Fiber driver only)
--- a/contracts/route/route.go
+++ b/contracts/route/route.go
@@ -16,11 +16,13 @@ type Route interface {
 	Fallback(handler contractshttp.HandlerFunc)
 	// GlobalMiddleware registers global middleware to be applied to all routes of the router.
 	GlobalMiddleware(middlewares ...contractshttp.Middleware)
-	// Listen starts the HTTP server and listens for incoming connections on the specified listener.
+	// Listen starts the HTTP server and listens on the specified listener.
 	Listen(l net.Listener) error
-	// ListenTLS starts the HTTPS server with the provided TLS configuration and listens on the specified listener.
-	ListenTLS(l net.Listener, certFile, keyFile string) error
-	// Run starts the HTTP server and listens for incoming connections on the specified host.
+	// ListenTLS starts the HTTPS server and listens on the specified listener.
+	ListenTLS(l net.Listener) error
+	// ListenTLSWithCert starts the HTTPS server with the provided TLS configuration and listens on the specified listener.
+	ListenTLSWithCert(l net.Listener, certFile, keyFile string) error
+	// Run starts the HTTP server and listens on the specified host.
 	Run(host ...string) error
 	// RunTLS starts the HTTPS server with the provided TLS configuration and listens on the specified host.
 	RunTLS(host ...string) error
--- a/contracts/route/route.go
+++ b/contracts/route/route.go
@@ -2,6 +2,7 @@ package route
 
 import (
 	"context"
+	"net"
 	"net/http"
 
 	contractshttp "github.com/goravel/framework/contracts/http"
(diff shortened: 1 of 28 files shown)


contracts/route/route.go
```diff
@@ -23,6 +24,8 @@ type Route interface {
 	RunTLSWithCert(host, certFile, keyFile string) error
 	// ServeHTTP serves HTTP requests.
 	ServeHTTP(writer http.ResponseWriter, request *http.Request)
+	// Listen starts the HTTP server and listens for incoming connections on the specified listener.
+	Listen(l net.Listener) error
 	// Stop gracefully stop the serve.
 	Stop(ctx ...context.Context) error
```
devhaozi, line 32: Q: I think it would be better to rename this method to Shutdown, which is what both the std library and go-fiber do.

hwbrzzl (commented):
Could you introduce how to use this method and what the change of `main.go` is?

devhaozi:
> Could you introduce how to use this method and what the change of `main.go` is?

for gin

image

hwbrzzl:
Okay, what's the change of main.go?

devhaozi:
> Okay, what's the change of main.go?

Similar to this

```go
	upg, err := tableflip.New(tableflip.Options{})
	if err != nil {
		panic(err)
	}
	defer upg.Stop()

	// Listen for the process signal to trigger the tableflip upgrade.
	go func() {
		sig := make(chan os.Signal, 1)
		signal.Notify(sig, syscall.SIGHUP)
		for range sig {
			if err = upg.Upgrade(); err != nil {
				facades.Log().Errorf("[Graceful] upgrade failed: %v", err)
			}
		}
	}()

	ln, err := upg.Listen("tcp", ":3000")
	if err != nil {
		panic(err)
	}
	defer ln.Close()

	go func() {
		if err = facades.Route().Run(); err != nil {
			facades.Log().Errorf("Route run error: %v", err)
		}
	}()

	// tableflip ready
	if err = upg.Ready(); err != nil {
		panic(err)
	}

	facades.Log().Infof("[Graceful] ready for upgrade")
	<-upg.Exit()

	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
	defer cancel()

	// Wait for connections to drain.
	log.Println("[Graceful] shutting down")
	return facades.Route().Shutdown(ctx)
```

hwbrzzl:
Oh, so complicated, can we wrap it?

devhaozi:
> Oh, so complicated, can we wrap it?

Seems unnecessary, users may not use graceful restart.

hwbrzzl:
I think this PR is solving the interruption when upgrading the application, it's unnecessary if User is using K8S, because K8S can build new pods before destroying the old. But if user lanch the application directly instead of K8S, this feature will useful. 

Currently, User have to launch the latest version with a different port and use Nginx to listen to it, otherwise, there is a short interruption during the upgrade.

contracts/route/route.go
```diff
@@ -23,6 +24,10 @@ type Route interface {
 	RunTLSWithCert(host, certFile, keyFile string) error
 	// ServeHTTP serves HTTP requests.
 	ServeHTTP(writer http.ResponseWriter, request *http.Request)
+	// Listen starts the HTTP server and listens for incoming connections on the specified listener.
+	Listen(l net.Listener) error
+	// ListenTLS starts the HTTPS server with the provided TLS configuration and listens on the specified listener.
+	ListenTLS(l net.Listener, certFile, keyFile string) error
```
devhaozi, line 30: This method is required because the std library uses an unexported method.

image

devhaozi:
> I think this PR is solving the interruption when upgrading the application, it's unnecessary if User is using K8S, because K8S can build new pods before destroying the old ones. But if User launches the application directly instead of K8S, this feature will be useful.
> 
> Currently, User have to launch the latest version with a different port and use Nginx to listen to it, otherwise, there is a short interruption during the upgrade.

After using github.com/cloudflare/tableflip , only need to send HUP single to application, then it will restart automatically.

contracts/route/route.go
hwbrzzl: It was `Shutdown` previously, however, we are going to add `Stop` method to `Schedule`, `Queue`, and `Testing.Database`, so I changed it to `Stop`. But now, I think `Shutdown` is better, it means functions are closed gracefully. 👍 I'll create a PR for it.

hwbrzzl:
Yes, [cloudflare/tableflip](https://github.com/cloudflare/tableflip) is a good feature, it should be better if we can wrap the complicated logic and use a simple way to implement it.

devhaozi:
> Yes, [cloudflare/tableflip](https://github.com/cloudflare/tableflip) is a good feature, it should be better if we can wrap the complicated logic and use a simple way to implement it.

You mean direct use [tableflip](https://github.com/cloudflare/tableflip) in there?

image

hwbrzzl:
Or just a few changes in main.go

devhaozi:
> > Yes, [cloudflare/tableflip](https://github.com/cloudflare/tableflip) is a good feature, it should be better if we can wrap the complicated logic and use a simple way to implement it.
> 
> You mean direct use [tableflip](https://github.com/cloudflare/tableflip) in there?
> 
> image


But the `select{}` here will prevent tableflip from running.
image

hwbrzzl:
`select{}` can be removed if needed. We can make a minimal implement first, to check the logic is expected.

hwbrzzl (approved):
Great

PR #767, opened 2024-12-17
chore: bump mockery to latest

devhaozi:
## 📑 Description

@coderabbitai summary

## ✅ Checks


hwbrzzl (commented):
Good catch. However, we'd better specify the mockery version. This is because there will be a huge update when mockery upgrades. Imagine if mockery is upgraded in a normal PR. There are a lot of mock files that update, which is unexpected.

PR #768, opened 2024-12-17
fix: should not print error when session doesn't exist

hwbrzzl:
## 📑 Description

image

## Summary by CodeRabbit

- **New Features**
	- Enhanced robustness of session handling by preventing unnecessary processing of empty values.

- **Bug Fixes**
	- Adjusted expected outcomes in tests to reflect the new behavior of the `Read` method, which now returns no error for non-existent sessions.
	- Removed specific session not found error handling, simplifying error management.

- **Tests**
	- Updated assertions in the test suite to expect no errors when reading from destroyed or garbage-collected files.

## ✅ Checks

### DIFF
--- a/session/driver/file.go
+++ b/session/driver/file.go
@@ -5,7 +5,6 @@ import (
 	"path/filepath"
 	"sync"
 
-	"github.com/goravel/framework/errors"
 	"github.com/goravel/framework/support/carbon"
 	"github.com/goravel/framework/support/file"
 )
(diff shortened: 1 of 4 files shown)


session/driver/file.go
```diff
@@ -80,7 +79,7 @@ func (f *File) Read(id string) (string, error) {
 		}
 	}
 
-	return "", errors.SessionNotFound.Args(id)
+	return "", nil
```
devhaozi, line 82: Q: Does `errors.SessionNotFound` need to be deleted?

hwbrzzl: Good question, we can remove it.

devhaozi (approved):
LGTM

PR #776, opened 2024-12-18
feat: [#507] Add `Abort` method for Response

kkumar-gcc:
## 📑 Description

RelatedTo https://github.com/goravel/goravel/issues/507

@coderabbitai summary

## ✅ Checks


contracts/http/response.go
```diff
@@ -49,6 +49,11 @@ type Response interface {
 	Render() error
 }
 
+type AbortResponse interface {
+	Response
+	Abort() error
```
hwbrzzl, line 54: Any concerns about not adding `Abort` to `Response` directly?

kkumar-gcc: I decided not to implement the abort method for streaming, downloads, and other formats as it was unnecessary. Currently, it is implemented only for `StringResponse` and `JsonResponse`, which makes more sense to me.

hwbrzzl: I think `Data` and `Redirect` are needed, too.

contracts/http/response.go
```diff
@@ -49,6 +49,11 @@ type Response interface {
 	Render() error
 }
 
+type AbortResponse interface {
```
hwbrzzl, line 52:
```suggestion
type ResponseWithAbort interface {
```

hwbrzzl (dismissed):
LGTM

hwbrzzl (approved):

PR #795, opened 2024-12-28
feat: [#507] Add `Abort` method for Response

kkumar-gcc:
Reverts goravel/framework#782


almas1992:
The name `ResponseWithAbort` is not very intuitive and can be misleading, making people think that this response has been aborted. I suggest renaming it to `ResponseCanAbort` or `AbortableResponse` to indicate that it is a response that can be aborted.

hwbrzzl:
> The name `ResponseWithAbort` is not very intuitive and can be misleading, making people think that this response has been aborted. I suggest renaming it to `ResponseCanAbort` or `AbortableResponse` to indicate that it is a response that can be aborted.

Good name, at first, it's `AbortResponse`, `Abort` + `Response` is a bit weird, but `AbortableResponse` makes sense.

hwbrzzl (approved):
LGTM

hwbrzzl (approved):

PR #799, opened 2024-12-30
chore: replace color with ctx.Success/Error

hwbrzzl:
## 📑 Description

@coderabbitai summary

## ✅ Checks

### DIFF
--- a/console/application.go
+++ b/console/application.go
@@ -11,10 +11,13 @@ import (
 )
 
 type Application struct {
-	instance *cli.App
+	instance  *cli.App
+	isArtisan bool
 }
 
-func NewApplication(name, usage, usageText, version string) console.Artisan {
+// NewApplication Create a new Artisan application.
+// The artisan parameter is used by goravel/installer.
+func NewApplication(name, usage, usageText, version string, artisan ...bool) console.Artisan {
 	instance := cli.NewApp()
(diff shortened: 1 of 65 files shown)


console/application.go
```diff
@@ -11,23 +11,20 @@ import (
 )
 
 type Application struct {
-	instance  *cli.App
-	isArtisan bool
+	instance *cli.App
 }
 
-func NewApplication(name, usage, usageText, version string, artisan ...bool) console.Artisan {
+func NewApplication(name, usage, usageText, version string) console.Artisan {
```
kkumar-gcc, line 17: This was added for `installer`.. NewApplication directly being used in goravel installer.

hwbrzzl: Oh, good catch, No wonder I didn't find the usage in framework.

almas1992: Is it not possible to use env.IsArtisan() for the check? Wouldn’t it be better to avoid relying on explicit parameter passing?

hwbrzzl: Good idea, I think it's possible.

hwbrzzl: Will create another PR to handle this.

hwbrzzl:
Removed the `IsArtisan` logic, will add it in another PR, this PR only contains replacing color logic.

PR #820, opened 2025-01-16
feat: [#547] Auth supports session driver

KlassnayaAfrodita:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/547

@coderabbitai summary

## ✅ Checks

### DIFF
--- a/auth/auth_session.go
+++ b/auth/auth_session.go
@@ -1,0 +1,153 @@
+package auth
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/spf13/cast"
+	"gorm.io/gorm/clause"
+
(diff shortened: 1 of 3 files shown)


hwbrzzl (commented):
Thanks, partial code should can be reuse, but optimization is required.

 The current logic is equal to `auth` + `JWT driver`, so for this issue:

1. We need to define a Driver interface
2. The JWT driver should be separated from the current logic
3. Implement a new session driver
4. Inject correct driver according to configration when calling NewAuth

hwbrzzl:
Close this PR first, feel free to reopen it when you have a chance to optimize it, thanks. @KlassnayaAfrodita

PR #867, opened 2025-02-08
feat: added take and last methods to query

saqijaan:
## 📑 Description

**Its new Feature.**

The First method provided by orm.Query automatically adds an ORDER BY clause. This meant that if someone wanted to retrieve a single record using a custom order, there was no built-in method available. To address this, GORM provides the Take method, which fetches a single record without applying any ORDER BY condition.

Additionally, the Last method was missing in the query, so I added it to quickly fetch the last record without requiring an ORDER BY clause. This method is also provided by GORM.

Using ORDER BY with First can lead to performance issues because it automatically includes the primary key in the ORDER BY clause along with any other fields specified for ordering. In the absence of a composite index, this can result in a significantly slow query.

@coderabbitai summary

## ✅ Checks


devhaozi:
@saqijaan Please run `mockery` in your local to generate new mocks for your methods.

hwbrzzl:
Here: https://www.goravel.dev/prologue/contributions.html#create-pr

image

saqijaan:
Sure I'll add

hwbrzzl (commented):
LGTM, could you add some test cases in `tests/queue_test.go` for the logic?

saqijaan:
Sure, I will. I’m just adding this whenever I have time so I don’t miss this feature 😉. It caused me a major performance issue. I have a table with around 2 million records, and using first with orderBy was making the query take around 7-9 seconds. Later, when checking the query log, I found out that the issue was caused by the first method.

hwbrzzl:
@saqijaan Or try to add some indexes?

saqijaan:
Indexes are already there on required columns. But I don't want to order by id and this still requires me to add composite index on id and other column I'm using for ordering my records.  So adding indexes is not a proper solution because it should only include columns in OrderBy which are provided by user.

hwbrzzl:
Sorry, I checked Laravel document, the action of these two funtions are different with that. It's unexpected, close this PR first. Feel free to make a deep discussion, Thanks!

PR #898, opened 2025-02-19
fix: Problem with submodels id string

dzham:
## 📑 Description

I found an error while working with the model.
link problem with the GetIDByReflect method. This method does not properly handle submodel with id string. 
Also, this method does not process variants with double nesting.

The problem is here, not in nested fields it works correctly
```
cast.ToString(id) == "" || cast.ToInt(id) == 0
```
The problem with such models

```
type User struct {
   TestStructString
   Name string
}

ts := User{Name: "name"}
ts.ID = "testId"
```

And with nested models

```
type UserFirst struct {
	TestStructString
	Name string
}
type UserSecond struct {
	UserFirst
	Avatar string
}

ts := UserSecond{}
ts.ID = "testId"
```

And I also refactored this method a little

### DIFF
--- a/support/database/database.go
+++ b/support/database/database.go
@@ -13,10 +13,6 @@ func GetID(dest any) any {
 	t := reflect.TypeOf(dest)
 	v := reflect.ValueOf(dest)
 
-	if t.Kind() == reflect.Pointer {
-		return GetIDByReflect(t.Elem(), v.Elem())
-	}
-
 	return GetIDByReflect(t, v)
 }
 
--- a/support/database/database.go
+++ b/support/database/database.go
@@ -3,8 +3,6 @@ package database
 import (
 	"reflect"
 	"strings"
-
-	"github.com/spf13/cast"
 )
 
 func GetID(dest any) any {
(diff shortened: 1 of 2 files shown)


support/database/database.go
```diff
@@ -36,20 +36,7 @@ func GetIDByReflect(t reflect.Type, v reflect.Value) any {
...
 		}
 		if v.Field(i).Type().Kind() == reflect.Struct {
-			structField := v.Field(i).Type()
-			for j := 0; j < structField.NumField(); j++ {
-				if !structField.Field(j).IsExported() {
-					continue
-				}
-				if strings.Contains(structField.Field(j).Tag.Get("gorm"), "primaryKey") {
-					id := v.Field(i).Field(j).Interface()
-					if cast.ToString(id) == "" || cast.ToInt(id) == 0 {
```
dzham, line 46: Problem this

hwbrzzl (approved):
Thanks, good catch!

hwbrzzl (changes requested):
Sorry, CI failed, I'm checking it.

support/database/database.go
hwbrzzl, line 1: Could you please update this file to below? The current logic has a bug.

```
package database

import (
	"reflect"
	"strings"
)

func GetID(dest any) any {
	if dest == nil {
		return nil
	}

	t := reflect.TypeOf(dest)
	v := reflect.ValueOf(dest)

	return GetIDByReflect(t, v)
}

func GetIDByReflect(t reflect.Type, v reflect.Value) any {
	if t.Kind() == reflect.Pointer {
		t = t.Elem()
		v = v.Elem()
	}

	for i := 0; i < t.NumField(); i++ {
		if !t.Field(i).IsExported() {
			continue
		}
		if strings.Contains(t.Field(i).Tag.Get("gorm"), "primaryKey") {
			if v.Field(i).IsZero() {
				return nil
			}

			return v.Field(i).Interface()
		}
		if t.Field(i).Type.Kind() == reflect.Struct && t.Field(i).Anonymous {
			return GetIDByReflect(t.Field(i).Type, v.Field(i))
		}
	}

	return nil
}
```

dzham: I didn't want to break the current logic. Added changes.

support/database/database.go
```diff
@@ -3,8 +3,6 @@ package database
 import (
 	"reflect"
 	"strings"
-
-	"github.com/spf13/cast"
 )
 
 func GetID(dest any) any {
```
hwbrzzl, line 8: Please remove line 16~18

hwbrzzl:
Hey @dzham Added you to the contributor list, and if you use Discord, you can send a message to me, I'll add you the contributor role.

PR #899, opened 2025-02-19
chore: optimize config facade

kkumar-gcc:
## 📑 Description

- **Fix type casting issue**: Previously, methods like `GetInt` and `GetString` accepted an `any` type as the default value, which could lead to runtime panics if casting failed (e.g., `default[0].(int)` on a string).
```go
config.GetInt("APP_PORT", "2000")
```
- **Add `GetDuration` method**: Provides a convenient way to retrieve time-related configuration values.

## ✅ Checks


hwbrzzl (approved):
Great 👍

PR #902, opened 2025-02-20
feat: [#538] track mockery tool dependencies in go.mod

almas1992:
## 📑 Description

This update introduces the management of the Mockery tool version within the framework’s go.mod, ensuring consistency across environments. By specifying the version, we avoid discrepancies in generated code caused by version mismatches of Mockery. The command execution now uses `go run github.com/vektra/mockery/v2` instead of the previous `mockery` command. _This approach is applicable to Go versions prior to 1.24, as later versions include built-in tool version management._

In Go 1.24 and later, you can execute the Mockery tool with the following command:
`go get --tool github.com/vektra/mockery/v2@v2.52.0`, and then use `go tool mockery` to run Mockery.

**Mockery is distinguished by a `//go:build tools `build tag, so its dependencies are not included in the normal build or packaging process.**

image

Closes https://github.com/goravel/goravel/issues/538

@coderabbitai summary

## ✅ Checks


kkumar-gcc:
I believe this is only necessary for framework developers; adding more dependencies for framework users is unnecessary.

almas1992:
> I believe this is only necessary for framework developers; adding more dependencies for framework users is unnecessary.  Simply adding an instruction about mockery in the contribution guide will also work.

I mentioned, “Mockery is distinguished by a //go:build tools build tag, so its dependencies are not included in the normal build or packaging process.”

almas1992:
Contributing code to the framework has become more common, so managing the versions of development tools is essential. First, it prevents developers from needing to install additional tools on their own machines. Second, the Mockery tool updates frequently and often requires adjustments to the production code.

hwbrzzl (commented):
LGTM, could you update [the action file](https://github.com/goravel/framework/blob/master/.github/workflows/mockery.yml) to use the new command? And what's the command in Golang v1.24? Could you update it in the description? We can remove the dependency after Golang v1.23 is not maintained. And please create another PR to update the Goravel document.

almas1992:
> LGTM, could you update [the action file](https://github.com/goravel/framework/blob/master/.github/workflows/mockery.yml) to use the new command? And what's the command in Golang v1.24? Could you update it in the description? We can remove the dependency after Golang v1.23 is not maintained. And please create another PR to update the Goravel document.

Done🔥

hwbrzzl (approved):

PR #625, opened 2024-09-06
feat: optimize the ctx parameter

hwbrzzl:
## 📑 Description

The ctx is unnecessary, change it to `...context.Context`. @zxdstyle

@coderabbitai summary

PR #955, opened 2025-03-12
fix: [#625] Throttle doesn't work when using the Redis cache driver

hwbrzzl:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/625

## ✅ Checks


http/limit/store.go
```diff
@@ -70,157 +25,133 @@ func NewStore(tokens uint64, interval time.Duration) (Store, error) {
...
-	if ok {
-		return b.take()
+func (s *Store) Take(ctx context.Context, key string) (tokens uint64, remaining uint64, reset uint64, ok bool, err error) {
+	lock, err := s.lock(key)
+	if err != nil {
+		return 0, 0, 0, false, err
 	}
 
-	nb := NewBucket(s.tokens, s.interval)
-	if err := s.cache.Put(key, nb, s.interval); err != nil {
```
hwbrzzl, line 96: The bug is caused by this line, `struct` can't be stored in Redis directly, `Put` will return an error here. It means this feature can't be used entirely when using the Redis driver. Through this thing, the next major work is to improve test coverage, before more people use the framework. cc @goravel/core-developers

devhaozi (dismissed):
LGTM

PR #956, opened 2025-03-15
feat: optimize refresh method

hwbrzzl:
## 📑 Description

We need to refresh all modules when testing in goravel/example with different drivers.

## ✅ Checks

### DIFF
--- a/foundation/container.go
+++ b/foundation/container.go
@@ -323,12 +323,12 @@ func (c *Container) MakeWith(key any, parameters map[string]any) (any, error) {
 
 func (c *Container) Refresh(bindings ...any) {
 	if len(bindings) == 0 {
-		c.instances.Range(func(key, value any) bool {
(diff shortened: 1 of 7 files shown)


foundation/container.go
```diff
@@ -321,8 +321,22 @@ func (c *Container) MakeWith(key any, parameters map[string]any) (any, error) {
...
 }
 
-func (c *Container) Refresh(key any) {
-	c.instances.Delete(key)
+func (c *Container) Refresh(bindings ...any) {
+	if len(bindings) == 0 {
+		// configInstance, ok := c.instances.Load(contracts.BindingConfig)
+
+		c.instances.Range(func(key, value any) bool {
+			if key != contracts.BindingConfig {
```
hwbrzzl, line 329: Shouldn't delete the base service provider.

PR #977, opened 2025-03-28
chore: add unit tests for package:install and package:uninstall commands

almas1992:
## 📑 Description

## ✅ Checks

### DIFF
--- a/foundation/console/package_install_command_test.go
+++ b/foundation/console/package_install_command_test.go
@@ -145,14 +145,6 @@ func (s *PackageInstallCommandTestSuite) TestHandle() {
 				}), "Package package@unknown installed successfully")
 			},
 		},
-		{
-			name: "package install success(real)",
-			setup: func() {
-				// TODO: A real package needs to be created to test installation
-				// when goravel/goravel upgrades all internal default dependencies to the latest version.
-			},
-			assert: func() {},
-		},
 	}
 	for _, test := range tests {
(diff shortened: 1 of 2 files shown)


foundation/console/package_install_command_test.go
```diff
@@ -0,0 +1,164 @@
...
+				s.Contains(color.CaptureOutput(func(w io.Writer) {
+					s.NoError(NewPackageInstallCommand().Handle(mockContext))
+				}), "Package package@unknown installed successfully")
+			},
+		},
+		{
+			name: "package install success(real)",
+			setup: func() {
+				// TODO: A real package needs to be created to test installation
+				// when goravel/goravel upgrades all internal default dependencies to the latest version.
```
hwbrzzl, line 152: Is it possible to implement here? If not, we may test it in goravel/example, and remove these lines.

almas1992: We can clone the goravel/goravel repository, then use make:package to create a local package. After that, install it with package:install and perform some assertions. However, some dependencies in goravel/goravel, such as gin, are not up-to-date, which causes the artisan commands to fail.

hwbrzzl: Yes, I suggest removing this case here and testing the logic in goravel/example.

hwbrzzl (approved):
lgtm

PR #646, opened 2024-09-15
feat: [#280] Implement Create and DropIfExists methods of Schema

hwbrzzl:
## 📑 Description

1. Implement `facades.Schema.Create()` and `facades.Schema.DropIfExists` methods.
2. Implement `table.ID()`, `table.String()`, and `table.Integer()` methods.
3. These methods can be used in a migration file in the next PR.
4. We can support more fields in the future, this is a good start.

Code:

image

Result:

image

## Summary by CodeRabbit

- **New Features**
	- Introduced new interfaces for enhanced database migration management, including `Blueprint`, `ColumnDefinition`, `Grammar`, `Migrator`, and `Repository`.
	- Added methods for creating, dropping, and managing database tables and columns.
	- Enhanced SQL generation and migration operations with new utility functions.
	- Improved schema construction with additional configuration and logging capabilities.

- **Bug Fixes**
	- Improved error handling and logging capabilities within the migration framework.

- **Tests**
	- Added comprehensive test suites for `Blueprint`, `ColumnDefinition`, and other migration functionalities to ensure reliability.

- **Documentation**
	- Enhanced comments and documentation for new methods and interfaces to assist developers in understanding their usage.

## ✅ Checks

- [x] Added test cases for my code

PR #986, opened 2025-04-02
feat: [#646] The Where function of Orm supports func condition

almas1992:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/646

```go
facades.Orm().Query().Where(func(query orm.Query) orm.Query {
		return query.Where("name != ?", "admin")
	}).FindOrFail(&user, 100)
```

## ✅ Checks


database/gorm/query.go
```diff
@@ -1215,6 +1225,11 @@ func (r *Query) buildWhere(db *gormio.DB) *gormio.DB {
 	}
 
 	for _, item := range r.conditions.where {
+		if sub, ok := item.query.(func(contractsorm.Query) contractsorm.Query); ok {
```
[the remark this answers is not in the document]
almas1992: It is implemented through Gorm’s [Group Conditions](https://gorm.io/docs/advanced_query.html#Group-Conditions), so theoretically, OrWhere should be supported. However, since Gorm processes OR conditions based on their position, there might be some issues with how it’s handled.

database/gorm/query.go
```diff
@@ -1198,6 +1198,18 @@ func (r *Query) buildSharedLock(db *gormio.DB) *gormio.DB {
...
+func (r *Query) buildSubquery(sub func(contractsorm.Query) contractsorm.Query) *gormio.DB {
+	db := r.instance.Session(&gormio.Session{NewDB: true, Initialized: true})
+	queryImpl := NewQuery(r.ctx, r.config, r.dbConfig, db, r.grammar, r.log, r.modelToObserver, nil)
+	query := sub(queryImpl)
+	var ok bool
+	if queryImpl, ok = query.(*Query); ok {
+		return queryImpl.buildWhere(db)
+	}
+
+	return db
```
hwbrzzl, line 1210: Can `nil` be returned here? Then `continue` at line 1231 if `r.buildSubquery(sub) == nil`.

almas1992: Yes, to prevent a panic when returning nil.

hwbrzzl: Em, I mean we can return `nil` here,  don't need to return a empty db.

almas1992: Yes, since the closure function returns an orm.Query interface, returning nil will still compile successfully. Therefore, it’s better to handle the nil case properly.

hwbrzzl (approved):
Nice

PR #994, opened 2025-04-07
feat: [#653] Update github.com/urfave/cli/v2 to v3

almas1992:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/653

## ✅ Checks

### DIFF
--- a/console/application.go
+++ b/console/application.go
@@ -22,12 +22,6 @@ var (
 		HideDefault: true,
 		Usage:       "Force disable ANSI output",
 	}
-
-	globalFlags = []cli.Flag{
-		noANSIFlag,
-		cli.HelpFlag,
-		cli.VersionFlag,
-	}
 )
 
 type Application struct {
--- a/console/application.go
+++ b/console/application.go
@@ -99,6 +99,8 @@ func (r *Application) CallAndExit(command string) {
 func (r *Application) Run(args []string, exitIfArtisan bool) error {
 	if noANSI || env.IsNoANSI() || slices.Contains(args, "--no-ansi") {
 		color.Disable()
+	} else {
+		color.Enable()
 	}
 
 	artisanIndex := -1
--- a/console/application.go
+++ b/console/application.go
@@ -1,11 +1,12 @@
 package console
 
 import (
+	"context"
 	"os"
 	"slices"
 	"strings"
 
-	"github.com/urfave/cli/v2"
+	"github.com/urfave/cli/v3"
 
 	"github.com/goravel/framework/contracts/console"
 	"github.com/goravel/framework/contracts/console/command"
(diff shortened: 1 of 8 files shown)


console/cli_helper.go
```diff
@@ -196,11 +199,53 @@ func getFlagName(flag cli.Flag) string {
...
+	return fullName
+}
+
 func indent(spaces int, v string) string {
 	pad := strings.Repeat(" ", spaces)
 	return pad + strings.Replace(v, "\n", "\n"+pad, -1)
 }
 
+// lexicographicLess compares strings alphabetically considering case.
+func lexicographicLess(i, j string) bool {
```
hwbrzzl, line 224: Could you add some test cases for this new function?

console/application.go
```diff
@@ -181,20 +181,25 @@ func flagsToCliFlags(flags []command.Flag) []cli.Flag {
...
 				Required: flag.Required,
-				Value:    flag.Value,
+				Value:    int64(flag.Value),
 			})
 		case command.FlagTypeIntSlice:
 			flag := flag.(*command.IntSliceFlag)
+			var int64Slice []int64
+			for _, v := range flag.Value {
+				int64Slice = append(int64Slice, int64(v))
+			}
```
hwbrzzl, line 191: The cli code is weird: `IntSliceFlag` vs `[]int64`

image

console/cli_helper.go
almas1992: This function is copied from [here](https://github.com/urfave/cli/blob/20335c61280dd50c922910e323451076bc90002b/sort.go#L6), since the cli.CommandsByName function was removed in v3. I simply reused the original implementation from the source. Given that it’s unchanged, do we still need to add additional test cases for it?

hwbrzzl: Yes, we can add some test cases to show the action of this function. It should be very easy to add it via AI.

console/application.go
almas1992: Yeah, it’s definitely weird. In v3, the original `int` was replaced with `int64`, so the flag no longer returns an `int`. I added a conversion in `console/application.go `to preserve the original behavior.

devhaozi: I think `Int*` can removed, only keep `Int64*`.

hwbrzzl: Yes, `case command.FlagTypeInt` and `case command.FlagTypeInt64` can use the same flow (in our framework, still distinguish FlagTypeInt and FlagTypeInt64). The same with `slice`.

console/cli_helper.go
almas1992: Done!

almas1992:
I’ve added a new function:
```go
func handleNoANSI() {
	if noANSI || env.IsNoANSI() {
		color.Disable()
	} else {
		color.Enable()
	}
}
```

This ensures that the --no-ansi option doesn’t have side effects between commands.
For example, in the following case:

```go
type Test struct {
}

// Signature The name and signature of the console command.
func (r *Test) Signature() string {
	return "app:test"
}

// Description The console command description.
func (r *Test) Description() string {
	return "Command description"
}

// Extend The console command extend.
func (r *Test) Extend() command.Extend {
	return command.Extend{Category: "app"}
}

// Handle Execute the console command.
func (r *Test) Handle(ctx console.Context) error {
	artisan := facades.Artisan()
	_ = artisan.Call("help --no-ansi build")
	ctx.NewLine()
	_ = artisan.Call("help migrate")
	return nil
}
```

Result:

image

image

hwbrzzl (approved):
Thanks, LGTM

PR #995, opened 2025-04-07
feat: add Routes method to support getting all registered routes

devhaozi:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/632

## ✅ Checks

### DIFF
--- a/contracts/route/route.go
+++ b/contracts/route/route.go
@@ -73,11 +73,7 @@ type Router interface {
 
 // Info represents a request route's specification which contains method and path and its handler.
 type Info struct {
-	Method      string
-	Path        string
-	Handler     string
-	HandlerFunc contractshttp.HandlerFunc
+	Method  string
+	Path    string
+	Handler string
 }
-
-// Infos defines a Info slice.
-type Infos []Info
--- a/contracts/route/route.go
+++ b/contracts/route/route.go
@@ -22,8 +22,10 @@ type Route interface {
 	ListenTLS(l net.Listener) error
 	// ListenTLSWithCert starts the HTTPS server with the provided TLS configuration and listens on the specified listener.
 	ListenTLSWithCert(l net.Listener, certFile, keyFile string) error
-	// Recover allows you to set a custom recovery when a request panics
+	// Recover allows you to set a custom recovery when a request panics.
 	Recover(recoverFunc func(ctx contractshttp.Context, err any))
+	// Routes returns all registered routes.
+	Routes() []Info
 	// Run starts the HTTP server and listens on the specified host.
 	Run(host ...string) error
 	// RunTLS starts the HTTPS server with the provided TLS configuration and listens on the specified host.
--- a/contracts/route/route.go
+++ b/contracts/route/route.go
@@ -70,3 +70,14 @@ type Router interface {
 	// StaticFS registers a new route with a path prefix to serve static files from the provided file system.
 	StaticFS(relativePath string, fs http.FileSystem)
 }
+
+// Info represents a request route's specification which contains method and path and its handler.
+type Info struct {
+	Method      string
+	Path        string
+	Handler     string
+	HandlerFunc contractshttp.HandlerFunc
+}
+
+// Infos defines a Info slice.
+type Infos []Info
--- a/mocks/route/Route.go
+++ b/mocks/route/Route.go
@@ -694,6 +694,53 @@ func (_c *Route_Resource_Call) RunAndReturn(run func(string, http.ResourceContro
 	return _c
 }
 
+// Routes provides a mock function with no fields
+func (_m *Route) Routes() []route.Info {
+	ret := _m.Called()
+
+	if len(ret) == 0 {
+		panic("no return value specified for Routes")
+	}
+
+	var r0 []route.Info
+	if rf, ok := ret.Get(0).(func() []route.Info); ok {
+		r0 = rf()
+	} else {
+		if ret.Get(0) != nil {
+			r0 = ret.Get(0).([]route.Info)
+		}
+	}
+
+	return r0
+}
+
+// Route_Routes_Call is a *mock.Call that shadows Run/Return methods with type explicit version for method 'Routes'
+type Route_Routes_Call struct {
+	*mock.Call
+}
+
+// Routes is a helper method to define mock.On call
+func (_e *Route_Expecter) Routes() *Route_Routes_Call {
+	return &Route_Routes_Call{Call: _e.mock.On("Routes")}
+}
+
+func (_c *Route_Routes_Call) Run(run func()) *Route_Routes_Call {
+	_c.Call.Run(func(args mock.Arguments) {
+		run()
+	})
+	return _c
+}
+
+func (_c *Route_Routes_Call) Return(_a0 []route.Info) *Route_Routes_Call {
+	_c.Call.Return(_a0)
+	return _c
+}
+
+func (_c *Route_Routes_Call) RunAndReturn(run func() []route.Info) *Route_Routes_Call {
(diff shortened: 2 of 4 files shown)


contracts/route/route.go
```diff
@@ -70,3 +70,10 @@ type Router interface {
 	// StaticFS registers a new route with a path prefix to serve static files from the provided file system.
 	StaticFS(relativePath string, fs http.FileSystem)
 }
+
+// Info represents a request route's specification which contains method and path and its handler.
+type Info struct {
+	Method  string
```
hwbrzzl, line 76: How about defining some constants for this field? User may not know what this is: GET or get?

devhaozi: > How about defining some constants for this field? User may not know what this is: GET or get?

image

Looks like can't use these fields directly, since they are defined as string.

almas1992 (commented):
Should the `Route interface` include a `Routes() []route.Info` method to expose route information?

contracts/route/route.go
hwbrzzl: Okay, if so, `string` should be fine. Could you add a comment to this field to tell the user what this field will be like?

contracts/route/route.go
```diff
@@ -22,8 +22,10 @@ type Route interface {
 	ListenTLS(l net.Listener) error
 	// ListenTLSWithCert starts the HTTPS server with the provided TLS configuration and listens on the specified listener.
 	ListenTLSWithCert(l net.Listener, certFile, keyFile string) error
-	// Recover allows you to set a custom recovery when a request panics
+	// Recover allows you to set a custom recovery when a request panics.
 	Recover(recoverFunc func(ctx contractshttp.Context, err any))
+	// Routes returns all registered routes.
+	Routes() []Info
```
hwbrzzl, line 28: It will be `facades.Route().Routes`, a bit strange. How about `List` or something else?

devhaozi: > It will be `facades.Route().Routes`, a bit strange. How about `List` or something else?

 
image

How about GetRoutes?

almas1992:
I feel like there’s something off with the current implementation of Routes() []route.Info.

The returned handler names are mostly like `github.com/goravel/gin.(*Group).xxx.handlerToGinHandler.func1`, which aren’t meaningful or helpful.

For example, when I register a route like:

```go
userController := controllers.NewUserController()
facades.Route().Get("/users/{id}", userController.Show)
```

The current `Routes()` output gives me a handler name like `github.com/goravel/gin.handlerToGinHandler.func1`, which doesn’t reflect the actual method being used.

Ideally, it should return something more meaningful, like `goravel/app/http/controllers.(*UserController), which is what dlv (Delve debugger) displays. That format is much more informative and would be very useful for tools like `artisan route:list` to display readable and traceable route-handler mappings.

image

devhaozi:
> I feel like there’s something off with the current implementation of Routes() []route.Info.
> 
> The returned handler names are mostly like `github.com/goravel/gin.(*Group).xxx.handlerToGinHandler.func1`, which aren’t meaningful or helpful.
> 
> For example, when I register a route like:
> 
> ```go
> userController := controllers.NewUserController()
> facades.Route().Get("/users/{id}", userController.Show)
> ```
> 
> The current `Routes()` output gives me a handler name like `github.com/goravel/gin.handlerToGinHandler.func1`, which doesn’t reflect the actual method being used.
> 
> Ideally, it should return something more meaningful, like `goravel/app/http/controllers.(*UserController)`, which is what dlv (Delve debugger) displays. That format is much more informative and would be very useful for tools like `artisan route:list` to display readable and traceable route-handler mappings.
> 
> image

Seems hard to do this, any suggestions? Or maybe just remove the `Handler` field?

almas1992:
By the way, you can retrieve the method name like `goravel/app/http/controllers.(*UserController).Show-fm` within the `func (r *Group) Get(relativePath string, handler httpcontract.HandlerFunc)` function.

For example:
```go
func (r *Group) Get(relativePath string, handler httpcontract.HandlerFunc) {
    pc := reflect.ValueOf(handler).Pointer()
    fn := runtime.FuncForPC(pc)
    fmt.Println("fn.Name() ========>", fn.Name())

    r.getRoutesWithMiddlewares().GET(
        pathToGinPath(relativePath),
        []gin.HandlerFunc{handlerToGinHandler(handler)}...,
    )
    r.clearMiddlewares()
}

```

Output:
```shell
fn.Name()=========> goravel/app/http/controllers.(*UserController).Show-fm
```

In other words, the` []route.Info` should ideally be populated during route definition, since all the necessary context—including the handler’s full name—is available at that point.

devhaozi:
> > **If performance is a concern, one possible optimization is to store `reflect.ValueOf(handler).Pointer()` during route definition, and defer the more expensive call to `runtime.FuncForPC` until the handler name actually needs to be displayed.**
> 
> There is no performance issue, because the routes do not need to be registered frequently. I am considering that this will require many modifications.

And need to support Group route, which seems not easy too.

contracts/route/route.go
almas1992: I agree—List seems more appropriate.

almas1992:
The required changes are indeed a bit extensive. For each routing method—such as `Any`, `Get`, `Post`, `Put`, `Resource`, etc.—we need to extract the handler’s function name and compute the full route path at the time of registration. This information should be stored in a global `[]route.Info` slice to ensure all routes are collected in one place.

A global variable is necessary because if `[]route.Info` were a field on individual structs, then route definitions—especially from grouped routes—would be scattered across multiple Group instances, making centralized tracking difficult.

Note: In Gin, computing the full route path for grouped routes requires combining both the originPrefix and prefix fields from the `Group` struct.

almas1992:
I’m wondering whether it’s necessary to build a _framework built-in routing system_, instead of relying on third-party libraries like Gin or Fiber. Since [Go 1.22 introduced built-in routing enhancements](https://go.dev/blog/routing-enhancements), the standard library is now capable of handling routing natively.

This would also give us full flexibility to implement structured routing, grouping, middleware, and other features without limitations—allowing the design to better align with the philosophy of our framework.   @hwbrzzl 

> I know this would be a massive undertaking, but as the saying goes, “The road ahead is long and arduous, I will seek far and wide.”

hwbrzzl:
> I’m wondering whether it’s necessary to build a _framework built-in routing system_, instead of relying on third-party libraries like Gin or Fiber. Since [Go 1.22 introduced built-in routing enhancements](https://go.dev/blog/routing-enhancements), the standard library is now capable of handling routing natively.
> 
> This would also give us full flexibility to implement structured routing, grouping, middleware, and other features without limitations—allowing the design to better align with the philosophy of our framework. @hwbrzzl
> 
> > I know this would be a massive undertaking, but as the saying goes, “The road ahead is long and arduous, I will seek far and wide.”

Yes, maybe in the future. For now, it should not be our main work, can't give us a lot of benefits.

devhaozi:
> I’m wondering whether it’s necessary to build a _framework built-in routing system_, instead of relying on third-party libraries like Gin or Fiber. Since [Go 1.22 introduced built-in routing enhancements](https://go.dev/blog/routing-enhancements), the standard library is now capable of handling routing natively.
> 
> This would also give us full flexibility to implement structured routing, grouping, middleware, and other features without limitations—allowing the design to better align with the philosophy of our framework. @hwbrzzl
> 
> > I know this would be a massive undertaking, but as the saying goes, “The road ahead is long and arduous, I will seek far and wide.”

Yes, that why I create a new package mux, I want to do some test first.

devhaozi:
Due to my work schedule, I am unable to continue this PR, if anyone is willing, please feel free to take over.

PR #998, opened 2025-04-09
feat: [#635] Optimize the process of generating migration and model

kkumar-gcc:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/635

This PR introduces support for generating a model directly from an existing database table using the `--table` option.

```bash
go run . artisan make:model --table=users -f User
```

## Example
Given the following table definition:
```go
return facades.Schema().Create("users", func(table schema.Blueprint) {
	table.ID()
	table.String("name").Nullable().Comment("User's full name")
	table.String("email").Comment("User's email address")
	table.String("password").Comment("User's hashed password")
	table.Boolean("is_active").Default(true).Comment("Indicates if the user is active")
	table.Integer("age").Nullable().Comment("User's age")
	table.BigInteger("balance").Default(0).Comment("User's account balance")
	table.Float("rating").Nullable().Comment("User rating score")
	table.Decimal("credits").Default(0.00).Comment("User credits balance")
	table.Enum("role", []any{"admin", "user", "moderator"}).Default("user").Comment("User role")
	table.Json("preferences").Nullable().Comment("User settings in JSON format")
	table.Date("birthdate").Nullable().Comment("User's birthdate")
	table.DateTime("last_login").Nullable().Comment("Last login timestamp")
	table.TimestampTz("created_at_2")
	table.Timestamps(6)
	table.SoftDeletes()
})
```
	
## Generated Model

```go
package models

import (
	"github.com/goravel/framework/database/orm"
)

type User struct {
	orm.Model
	orm.SoftDeletes
	Name        *string          `json:"name" gorm:"column:name"`
	Email       string           `json:"email" gorm:"column:email"`
	Password    string           `json:"password" gorm:"column:password"`
	IsActive    bool             `json:"is_active" gorm:"column:is_active"`
	Age         *int             `json:"age" gorm:"column:age"`
	Balance     int64            `json:"balance" gorm:"column:balance"`
	Rating      *float64         `json:"rating" gorm:"column:rating"`
	Credits     float64          `json:"credits" gorm:"column:credits"`
	Role        any              `json:"role" gorm:"column:role"`
	Preferences *string          `json:"preferences" gorm:"column:preferences"`
	Birthdate   *carbon.DateTime `json:"birthdate" gorm:"column:birthdate"`
	LastLogin   *carbon.DateTime `json:"last_login" gorm:"column:last_login"`
	CreatedAt2  carbon.DateTime  `json:"created_at_2" gorm:"column:created_at_2"`
}

func (m *User) TableName() string {
	return "users"
}
```

### Overriding the mapping:
Users can override the default mapping in their app's `app/providers/database_service_provider.go`:
```go
facades.Schema().Extend(&contractsschema.Extension{
    GoTypes: []contractsschema.GoType{
        {
            Pattern: "uuid", 
            Type: "uuid.UUID", 
            NullType: "uuid.NullUUID", 
            Imports: []string{"github.com/google/uuid"},
        },
        {
            Pattern: "point", 
            Type: "geom.Point", 
            NullType: "*geom.Point", 
            Imports: []string{"github.com/twpayne/go-geom"},
        },
    },
})
```

## ✅ Checks


support/enum/enum.go
```diff
@@ -0,0 +1,54 @@
+package enum
+
+import "github.com/spf13/cast"
+
+type Enum[K comparable, V any] struct {
+	key   K
+	value V
+}
+
+func NewEnum[K comparable, V any](key K, value V) Enum[K, V] {
```
almas1992, line 10: How about renaming `NewEnum` to `New`?

Since it’s in the `enum` package, `enum.NewEnum` feels redundant — `enum.New` should be clear enough.

krishan-kumar-meesho: Hmm, make sense

kkumar-gcc: Hmm, make sense

database/console/model_make_command.go
```diff
@@ -37,35 +65,253 @@ func (r *ModelMakeCommand) Extend() command.Extend {
...
+	if schemaType != "" {
+		if typeInfo, ok := goTypeMapping[schemaType]; ok {
+			goType = typeInfo.Type
+			if column.Nullable && typeInfo.NullType != "" {
+				goType = typeInfo.NullType
+			}
+
+			imports = typeInfo.Imports
+
+			// TODO: discuss precision based typing
```
kkumar-gcc, line 261: @hwbrzzl  How should we parse the precision of data type from the database? Do we have any api here?

kkumar-gcc: Not implemented yet, but thinking about allowing Go type mapping based on DB column precision. For example, for double, you could do:

```go
PrecisionBasedType: map[string]schema.GoTypeMapping{
    ":6": {
      Type:     "float32",
      NullType: "*float32",
      Imports:  []string{},
    },
    "7:": {
      Type:     "float64",
      NullType: "*float64",
      Imports:  []string{},
    },
}
```

So if precision is 6 or less → `float32`, if 7 or more → `float64`. Could support ranges like :6, 7:, 5:10, or exact like 8.

hwbrzzl (commented):
Amazing PR 👍

database/console/model_make_command.go
```diff
@@ -37,35 +65,253 @@ func (r *ModelMakeCommand) Extend() command.Extend {
...
 
-	if err := file.PutContent(m.GetFilePath(), r.populateStub(r.getStub(), m.GetPackageName(), m.GetStructName())); err != nil {
-		return err
+	table := ctx.Option("table")
+	modelInfo := modelInfo{}
+	structName := m.GetStructName()
+
+	if table != "" {
+		if !r.schema.HasTable(table) {
+			ctx.Error(fmt.Sprintf("table %s does not exist", table))
```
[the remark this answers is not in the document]
kkumar-gcc: These errors are only used in this particular command and will not be used elsewhere, so it is unnecessary to add them to `errors/list.go`.

database/console/model_make_command.go
```diff
@@ -37,35 +65,253 @@ func (r *ModelMakeCommand) Extend() command.Extend {
...
+
+			imports = typeInfo.Imports
+
+			// TODO: discuss precision based typing
+		}
+	}
+
+	tagParts := []string{fmt.Sprintf(`json:"%s"`, column.Name)}
+
+	gormTag := fmt.Sprintf("column:%s", column.Name)
```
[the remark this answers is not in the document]
kkumar-gcc: Hmm, make sense

database/schema/blueprint.go
```diff
@@ -610,10 +611,10 @@ func (r *Blueprint) addImpliedCommands(grammar driver.Grammar) {
 	r.addAttributeCommands(grammar)
 }
 
-func (r *Blueprint) createAndAddColumn(ttype, name string) *ColumnDefinition {
+func (r *Blueprint) createAndAddColumn(ttype schema.ColumnType, name string) *ColumnDefinition {
```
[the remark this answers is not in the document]
kkumar-gcc: Looks like you missed usage of `createAndAddColumn`. Basically, instead of hardcoding ttype like `bigInteger`, I created an enum type called `ColumnType` and defined all supported ttype values in one place.

database/schema/schema.go
```diff
@@ -418,3 +446,133 @@ func (r *Schema) build(blueprint contractsschema.Blueprint) error {
 func (r *Schema) createBlueprint(table string) contractsschema.Blueprint {
 	return NewBlueprint(r, r.prefix, table)
 }
+
+func getDefaultGoTypeMapping() map[string]contractsschema.GoTypeMapping {
+	return map[string]contractsschema.GoTypeMapping{
+		contractsschema.TypeBigInteger.Value(): {
+			Type:     "int64",
+			NullType: "*int64",
```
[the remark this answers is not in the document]
kkumar-gcc: It's not always *; sometimes the user can customize different null types—for example, using `string` for normal types and `sql.NullString` for nullable ones.

contracts/database/schema/schema.go
```diff
@@ -42,6 +42,8 @@ type Schema interface {
 	GetTypes() ([]driver.Type, error)
 	// GetViews Get the views that belong to the database.
 	GetViews() ([]driver.View, error)
+	// GoTypeMap returns the mapping of schema types to Go types.
+	GoTypeMap() map[string]GoTypeMapping
```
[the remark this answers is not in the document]
kkumar-gcc: Here, string refers to the column type, like `bigInteger`, `tinyInteger,` etc,  which are types defined by us.

contracts/database/driver/grammar.go
```diff
@@ -126,6 +126,8 @@ type SchemaGrammar interface {
 	TypeSmallInteger(column ColumnDefinition) string
 	// TypeString Create the column definition for a string type.
 	TypeString(column ColumnDefinition) string
+	// TypePatternMapping returns a list of regex-based mappings from database-specific types to schema types.
+	TypePatternMapping() []TypePatternMapping
```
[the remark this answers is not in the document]
kkumar-gcc: Not sure—I thought `TypePattern` was meant to describe the `pattern` of a type, like `TypeString` representing `string`.

database/console/model_make_command.go
```diff
@@ -37,35 +65,253 @@ func (r *ModelMakeCommand) Extend() command.Extend {
...
+			hasDeletedAt = true
+		}
+	}
+
+	embedOrmModel := hasID && hasCreatedAt && hasUpdatedAt
+	embedOrmTimestamps := !embedOrmModel && hasCreatedAt && hasUpdatedAt
+	embedOrmSoftDeletes := hasDeletedAt
+
+	if embedOrmModel {
+		info.Embeds = append(info.Embeds, "orm.Model")
```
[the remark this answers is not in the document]
kkumar-gcc: This will complicate things.

database/schema/schema.go
```diff
@@ -281,6 +281,34 @@ func (r *Schema) GetViews() ([]driver.View, error) {
 	return views, nil
 }
 
+func (r *Schema) GoTypeMap() map[string]contractsschema.GoTypeMapping {
+	typeMapping := getDefaultGoTypeMapping()
+	mappingConfig, ok := r.config.Get("database.model.mapping").(map[string]contractsschema.GoTypeMapping)
```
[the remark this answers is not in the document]
kkumar-gcc: Then we need to introduce functionality like an `Extend` method to extend the core.

database/console/model_make_command.go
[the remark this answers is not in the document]
hwbrzzl: When we implement the localization for the framework internal errors, it's necessary to move all errors into one place.

database/console/model_make_command.go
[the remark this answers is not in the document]
hwbrzzl: `orm.Model` will be deprecated.

database/schema/schema.go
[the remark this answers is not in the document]
hwbrzzl: Yes, maybe.

database/console/model_make_command.go
```diff
@@ -264,12 +264,6 @@ func generateField(column driver.Column, typePatternMapping []driver.TypePattern
 
 	tagParts := []string{fmt.Sprintf(`json:"%s"`, column.Name)}
 
-	gormTag := fmt.Sprintf("column:%s", column.Name)
-	if column.Autoincrement {
-		gormTag += ";autoIncrement"
-	}
```
hwbrzzl: This check should be required.

database/console/model_make_command.go
```diff
@@ -37,35 +65,247 @@ func (r *ModelMakeCommand) Extend() command.Extend {
...
+}
+
+func generateField(column driver.Column, typePatternMapping []driver.TypePatternMapping, goTypeMapping map[string]schema.GoTypeMapping) fieldInfo {
+	schemaType := getSchemaType(column.Type, typePatternMapping)
+
+	goType := "any"
+	var imports []string
+
+	if schemaType != "" {
+		if typeInfo, ok := goTypeMapping[schemaType]; ok {
```
hwbrzzl, line 253: I think the transform process is `grammar.TypePatternMapping` -> `schemaType` -> `goType`. Is schemaType required? Can it be removed? `grammar.TypePatternMapping` -> `goType` directly.

kkumar-gcc: The pattern type will be set at the driver level, so users don’t have to worry about all the details of each DB type. But they’ll still need to write configs for every DB type across different drivers, which can get tricky.

For example, just to represent a boolean, they might have to list out boolean, bool, and tinyint(1).

database/schema/blueprint.go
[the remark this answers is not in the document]
hwbrzzl: Okay, is there any benefit, please? If it is only used here.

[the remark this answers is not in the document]
kkumar-gcc: Yes, it's a good idea to avoid hard-coding values like this in multiple places. Defining them in one central location makes it easier to manage. That way, I’ll have a clear list of supported types, and updating a type won’t break the code elsewhere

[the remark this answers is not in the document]
hwbrzzl: Good point. I saw `Enum` is only used in this case. If so, how about define some constants for supported types directly instead of `Enum`? Is `Enum` required?

database/console/model_make_command.go
hwbrzzl: Sorry, I didn't get your meaning. Could you show a specific case? In my opinion, maybe below is enough:

```
sqlite: tinyint(1) -> Bool (go type)
mysql: tinyint(1) -> Bool (go type)
postgres: boolean -> Bool (go type)
sqlserver: bit -> Bool (go type)
```

contracts/database/schema/schema.go
```diff
@@ -83,3 +85,10 @@ type Connection interface {
 	// Connection Get the connection for the migration.
 	Connection() string
 }
+
+type GoTypeMapping struct {
+	Type                string
+	NullType            string
+	Imports             []string
+	PrecisionBasedTypes map[string]GoTypeMapping
```
hwbrzzl, line 93: I found there is no specific difference between `Type` and `NullType`, only an extra `*`. So I think `NullType` should be unnecessary for now. 

`Imports []string` can be changed to `Import string`, and it is only related to `Type`.

We can add `NullType` and `NullImport` in the future if necessary.

`PrecisionBasedTypes` should be unnecessary as well, given `TypePatternMapping.Pattern` can be a regexp, we can set `^int[1-8]$` (DbType) -> int (GoType) directly.

database/console/model_make_command.go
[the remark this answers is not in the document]
kkumar-gcc: Sure—I’ve implemented it so that it can be easily deprecated in the future

contracts/database/driver/grammar.go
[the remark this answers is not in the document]
kkumar-gcc: Removed this field, as it’s no longer needed in the two-layer architecture.

database/console/model_make_command.go
```diff
@@ -37,35 +64,254 @@ func (r *ModelMakeCommand) Extend() command.Extend {
...
+		case "deleted_at":
+			hasDeletedAt = true
+			isDeletedAtNullable = col.Nullable
+			standardColumns["deleted_at"] = true
+		}
+	}
+
+	var modelEmbed, timestampsEmbed, softDeletesEmbed string
+
+	hasNullableTimestamps := (hasCreatedAt && isCreatedAtNullable) || (hasUpdatedAt && isUpdatedAtNullable)
```
hwbrzzl, line 158:
```suggestion
	hasNullableTimestamps := (hasCreatedAt && isCreatedAtNullable) && (hasUpdatedAt && isUpdatedAtNullable)
```

database/console/model_make_command.go
```diff
@@ -37,35 +64,254 @@ func (r *ModelMakeCommand) Extend() command.Extend {
...
+	}
+
+	if len(info.Embeds) > 0 {
+		info.Imports["github.com/goravel/framework/database/orm"] = struct{}{}
+	}
+
+	goTypeMapping := r.schema.GoTypes()
+
+	for _, col := range columns {
+		if (modelEmbed != "" || timestampsEmbed != "") &&
```
hwbrzzl, line 202: `modelEmbed` and `timestampsEmbed` should be split.

database/console/model_make_command.go
```diff
@@ -37,35 +64,254 @@ func (r *ModelMakeCommand) Extend() command.Extend {
...
+	typeInfo := getSchemaType(column.Type, typeMapping)
+
+	goType := typeInfo.Type
+	if column.Nullable && typeInfo.NullType != "" {
+		goType = typeInfo.NullType
+	}
+
+	tagParts := []string{fmt.Sprintf(`json:"%s"`, column.Name)}
+
+	gormTag := fmt.Sprintf("column:%s", column.Name)
```
hwbrzzl, line 291: The `column` tag is unnecessary, and please add a `db` tag.

database/console/model_make_command.go
```diff
@@ -37,35 +64,254 @@ func (r *ModelMakeCommand) Extend() command.Extend {
...
+		field := generateField(col, goTypeMapping)
+
+		for _, importPath := range field.Imports {
+			info.Imports[importPath] = struct{}{}
+		}
+
+		info.Fields = append(info.Fields, r.buildField(field.Name, field.Type, field.Tags))
+	}
+
+	info.NeedsTableName = true
```
hwbrzzl, line 220: We can compare `structName` and `tableName`, set `NeedsTableName` according to the result.

database/console/model_make_command_test.go
hwbrzzl, line 1: Perfect 👍

hwbrzzl (commented):
Great PR 👍

database/console/model_make_command.go
```diff
@@ -37,35 +64,254 @@ func (r *ModelMakeCommand) Extend() command.Extend {
...
+			ctx.Error(err.Error())
+			return nil
+		}
+
+		model, err = r.generateModelInfo(columns, structName, table)
+		if err != nil {
+			ctx.Error(err.Error())
+			return nil
+		}
+		ctx.Info(fmt.Sprintf("Generated %d fields and %d embeds from table '%s'", len(model.Fields), len(model.Embeds), table))
```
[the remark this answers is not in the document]
kkumar-gcc: No, It was used for debug propose only

contracts/database/schema/schema.go
```diff
@@ -83,3 +87,17 @@ type Connection interface {
...
+	GoTypes []GoType
+}
+
+// GoType represents a database column type to Go type mapping
+// This is kept for backward compatibility
+type GoType struct {
+	Pattern  string
+	Type     string
+	NullType string
+	Imports  []string
```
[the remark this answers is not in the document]
kkumar-gcc: I don't see any use case for differentiating between `Import` and `Null-Import`.

database/console/model_make_command.go
kkumar-gcc: Good catch!

database/schema/schema.go
```diff
@@ -418,3 +429,192 @@ func (r *Schema) build(blueprint contractsschema.Blueprint) error {
...
+					break
+				}
+			}
+		}
+	}
+
+	r.goTypes = result
+}
+
+func defaultGoTypes() []contractsschema.GoType {
```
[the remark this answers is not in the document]
kkumar-gcc: Let's add the default in one place for better manageability. This should be database-dependent, not driver-dependent.

contracts/database/schema/schema.go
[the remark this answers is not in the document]
hwbrzzl: If `Type` and `NullType` can be different, `import` may not be the same.

tests/schema_test.go
```diff
@@ -2660,3 +2661,163 @@ func TestSqlserverSchema(t *testing.T) {
...
+			s.True(found, "jsonb type should exist in original types")
+			s.Equal("string", originalJsonb.Type, "Original jsonb type should be string")
+
+			extendedJsonb, found := findGoType(jsonbPattern, extendedGoTypes)
+			s.True(found, "jsonb type should exist in extended types")
+			s.Equal("jsonb.RawMessage", extendedJsonb.Type, "Extended jsonb type should be jsonb.RawMessage")
+			s.Equal("*jsonb.RawMessage", extendedJsonb.NullType)
+			s.Contains(extendedJsonb.Imports, "github.com/jmoiron/sqlx/types")
+
+			// Create a table with both original and extended types to test if they work
```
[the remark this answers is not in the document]
kkumar-gcc: This isn't testing `schema.Create` — it's checking that `Extend` can override or add columns, and that they show up in `GoTypes`. Real type usage is tested in `console/make_model_command.go`.

contracts/database/schema/schema.go
```diff
@@ -22,6 +22,8 @@ type Schema interface {
 	DropColumns(table string, columns []string) error
 	// DropIfExists Drop a table from the schema if exists.
 	DropIfExists(table string) error
+	// Extend the schema with given extend parameter.
+	Extend(extend *Extension) Schema
```
hwbrzzl, line 26: How about `struct` instead of `pointer`? I think `pointer` is unnecessary.

```suggestion
	Extend(extend Extension) Schema
```

database/orm/model.go
hwbrzzl, line 1: So sorry for the back and forth, due to https://github.com/goravel/goravel/issues/642#issuecomment-2868615799, this change may be unnecessary for now. We need a deep investigation into this. We can merge this PR first and create another PR to remove `BaseModel`, `NullableSoftDeletes` and `NullableTimestamps` if you want.

database/schema/schema.go
```diff
@@ -430,84 +430,47 @@ func (r *Schema) createBlueprint(table string) contractsschema.Blueprint {
 	return NewBlueprint(r, r.prefix, table)
 }
 
-func (r *Schema) extendGoTypes(goTypes []contractsschema.GoType) {
-	if len(goTypes) == 0 {
+// extendGoTypes merges user-provided GoType overrides and additions into the schema's default mappings.
+// New patterns (not present in defaults) are prepended for highest priority. Existing patterns are updated with non-zero override fields.
+func (r *Schema) extendGoTypes(overrides []contractsschema.GoType) {
```
hwbrzzl, line 435: Could you add some tests for this function? `tests/schema_test.go` ensures the basic logic is expected in real DBs, unit tests ensure the internal logic of the function is robust enough.

database/orm/model.go
kkumar-gcc: will remove from this PR

kkumar-gcc: Will create a separate PR for this

database/schema/schema.go
kkumar-gcc: Done

database/console/model_make_command_test.go
```diff
@@ -2,42 +2,418 @@ package console
...
+	assert.Nil(t, modelMakeCommand.Handle(mockContext))
+	assert.True(t, file.Exists("app/models/product.go"))
+	assert.True(t, file.Contain("app/models/product.go", "package models"))
+	assert.True(t, file.Contain("app/models/product.go", "type Product struct"))
+	assert.True(t, file.Contain("app/models/product.go", "orm.NullableModel"))
+	assert.True(t, file.Contain("app/models/product.go", "orm.NullableSoftDeletes"))
+	assert.True(t, file.Contain("app/models/product.go", "Name"))
+	assert.True(t, file.Contain("app/models/product.go", "Price"))
+	assert.True(t, file.Contain("app/models/product.go", "Stock"))
+	assert.True(t, file.Contain("app/models/product.go", "TableName"))
```
hwbrzzl, line 98: Can we confirm the entire content? Eg: assert.Equel(t, `package ***`, file.GetContent("app/models/product.go"))

database/console/model_make_command_test.go
```diff
@@ -2,42 +2,418 @@ package console
...
+
+	info, err := modelMakeCommand.generateModelInfo(columns, "User", "users")
+	assert.Nil(t, err)
+	assert.Equal(t, 1, len(info.Fields))
+	assert.Equal(t, 2, len(info.Embeds))
+	assert.True(t, strings.Contains(info.Fields[0], "Name"))
+	assert.True(t, strings.Contains(info.TableNameMethod, "func (r *User) TableName() string"))
+	assert.Contains(t, info.Embeds, "orm.Model")
+	assert.Contains(t, info.Embeds, "orm.NullableSoftDeletes")
+	assert.Contains(t, info.Imports, "github.com/goravel/framework/database/orm")
```
hwbrzzl, line 150: Ditto

database/console/model_make_command_test.go
```diff
@@ -2,42 +2,418 @@ package console
...
+		Imports:         map[string]struct{}{},
+		TableNameMethod: "",
+	}
+
+	result, err := modelMakeCommand.populateStub(stub, "models", "User", modelInfo)
+	assert.Nil(t, err)
+	assert.Contains(t, result, "package models")
+	assert.Contains(t, result, "type User struct")
+	assert.Contains(t, result, "ID   uint")
+	assert.Contains(t, result, "Name string")
```
hwbrzzl, line 292: Ditto

hwbrzzl (commented):
There are some issues when testing it locally:

image

database/console/model_make_command.go
```diff
@@ -37,35 +62,264 @@ func (r *ModelMakeCommand) Extend() command.Extend {
...
+	}
+
+	imports := make([]string, 0, 2)
+	if typeInfo.Import != "" {
+		imports = append(imports, typeInfo.Import)
+	}
+
+	if typeInfo.NullImport != "" {
+		imports = append(imports, typeInfo.NullImport)
+	}
```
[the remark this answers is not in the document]
kkumar-gcc: These imports are deduplicated when building the `modelDefinition`, so only unique imports are retained.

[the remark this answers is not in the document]
hwbrzzl: But if `Import` and `NullImport` are different, only one is required.

database/console/model_make_command.go
```diff
@@ -37,35 +62,264 @@ func (r *ModelMakeCommand) Extend() command.Extend {
...
 
-	return stub
+	tagParts := []string{
+		fmt.Sprintf(`json:"%s"`, column.Name),
+		fmt.Sprintf(`db:"%s"`, column.Name),
+	}
+
+	if column.Autoincrement {
+		tagParts = append(tagParts, fmt.Sprintf(`gorm:"%s"`, "autoIncrement"))
```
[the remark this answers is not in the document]
kkumar-gcc: PrimaryKey and AutoIncrement are two separate concepts, right?

kkumar-gcc:
> There are some issues when testing it locally:
> image

That first error was my mistake—I renamed NullableModel to BaseModel. In any case, could you please share your migration?

database/console/model_make_command_test.go
kkumar-gcc: This test doesn't generate model file

database/console/model_make_command.go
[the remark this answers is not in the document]
hwbrzzl: Yes, but a AutoIncrement column should be PrimaryKey.

hwbrzzl:
CI failed

database/console/model_make_command_test.go
```diff
@@ -2,42 +2,456 @@ package console
...
+
+	mockSchema.EXPECT().GoTypes().Return(goTypes).Once()
+
+	info, err = modelMakeCommand.generateModelInfo(columns, "Post", "posts")
+	assert.Nil(t, err)
+	assert.Equal(t, 1, len(info.Fields))
+	assert.Equal(t, 2, len(info.Embeds))
+	assert.True(t, strings.Contains(info.Fields[0], "Title"))
+	assert.Contains(t, info.Embeds, "orm.NullableTimestamps")
+	assert.Contains(t, info.Embeds, "orm.SoftDeletes")
```
hwbrzzl, line 201: Ditto

database/console/model_make_command_test.go
```diff
@@ -2,42 +2,456 @@ package console
...
+
+	typeMapping := []schema.GoType{
+		{Pattern: "varchar", Type: "string", NullType: "sql.NullString", NullImport: "database/sql"},
+	}
+
+	field := generateField(column, typeMapping)
+	assert.Equal(t, "Name", field.Name)
+	assert.Equal(t, "string", field.Type)
+	assert.Contains(t, field.Tags, "json:\"name\"")
+	assert.Contains(t, field.Tags, "db:\"name\"")
```
hwbrzzl, line 349: Ditto

hwbrzzl (commented):
Great PR 👍 LGTM, only nitpicks on the test cases.

hwbrzzl (approved):
Great 👍 Thanks

PR #1005, opened 2025-04-12
feat: enhance session manager to support custom drivers

merouanekhalili:
Refactor session manager architecture to allow for custom driver integration. Update file driver implementation to be compatible with these changes, making the session system more extensible and flexible.

### Overview
This pull request introduces a significant refactor of the Session Manager to allow the use of custom session drivers, including Redis. The initial goal was to implement a Redis session driver directly; however, I discovered that the current implementation of the Session Manager did not support such integration. To overcome this limitation, I refactored both the Session Manager and its associated file driver, paving the way for adding custom session drivers.

### Updated Session Configuration:
```go
package config

import (
	"github.com/goravel/framework/contracts/session" // Required for Driver type in factory signature
	fileriver "github.com/goravel/framework/session/driver" // Import the package containing the driver AND its factory
	"github.com/goravel/framework/support/path"
	"github.com/goravel/framework/support/str"
)

func init() {
	config := facades.Config()
	config.Add("session", map[string]any{
		// Default Session Driver
		//
		// This option controls the default session "driver" that will be used on
		// requests. By default, we will use the lightweight file session driver, but you
		// may specify any of the other wonderful drivers provided here.
		//
		// Supported: "file"
		"driver": config.Env("SESSION_DRIVER", "file"),

		// Session Lifetime
		//
		// Here you may specify the number of minutes that you wish the session
		// to be allowed to remain idle before it expires. If you want them
		// to immediately expire when the browser is closed, then you may
		// indicate that via the expire_on_close configuration option.
		"lifetime": config.Env("SESSION_LIFETIME", 120),

		"expire_on_close": config.Env("SESSION_EXPIRE_ON_CLOSE", false),

		// Session File Location
		//
		// When using the file session driver, we need a location where the
		// session files may be stored. A default has been set for you, but a
		// different location may be specified. This is only needed for file sessions.
		"files": path.Storage("framework/sessions"),

		// Session Garbage Collection Running Time Interval (in minutes)
		//
		// Here you may specify how many minutes you want the session to be allowed
		// to remain idle before its garbage collection routine runs. If you want
		// to avoid hitting the garbage collection routine, you may set this value to -1.
		"gc_interval": config.Env("SESSION_GC_INTERVAL", 30),

		// Session Cookie Name
		//
		// Here you may change the name of the cookie used to identify a session
		// in the application. The name specified here will get used every time
		// a new session cookie is created by the framework for every driver.
		"cookie": config.Env("SESSION_COOKIE", str.Of(config.GetString("app.name")).Snake().Lower().String()+"_session"),

		// Session Cookie Path
		//
		// The session cookie path determines the path for which the cookie will
		// be regarded as available.Typically, this will be the root path of
		// your application, but you are free to change this when necessary.
		"path": config.Env("SESSION_PATH", "/"),

		// Session Cookie Domain
		//
		// Here you may change the domain of the cookie used to identify a session
		// in your application.This will determine which domains the cookie is
		// available to in your application.A sensible default has been set.
		"domain": config.Env("SESSION_DOMAIN", ""),

		// HTTPS Only Cookies
		//
		// By setting this option to true, session cookies will only be sent back
		// to the server if the browser has an HTTPS connection. This will keep
		// the cookie from being sent to you if it cannot be done securely.
		"secure": config.Env("SESSION_SECURE", false),

		// HTTP Access Only
		//
		// Setting this to true will prevent JavaScript from accessing the value of
		// the cookie, and the cookie will only be accessible through the HTTP protocol.
		"http_only": config.Env("SESSION_HTTP_ONLY", true),

		// Same-Site Cookies
		//
		// This option determines how your cookies behave when cross-site requests
		// take place, and can be used to mitigate CSRF attacks.By default, we
		// will set this value to "lax" since this is a secure default value.
		"same_site": config.Env("SESSION_SAME_SITE", "lax"),

		// Session Driver Configurations
		//
		// This map defines the available session drivers and how to create them.
		// The Session Manager reads this map to register driver factories.
		"drivers": map[string]any{

			// File Driver Configuration
			"file": map[string]any{
				// func() (session.Driver, error)
				"via": fileriver.FileDriverFactory,
			},

			// Example: Redis Driver Configuration (User would add this)
			/*
				"redis": map[string]any{
					"driver":     "redis",
					"connection": "default"

					// 'via' points to the factory function for the Redis driver
					"via": func() (session.Driver, error) {
						return redisfacades.RedisSession("redis"), nil
					},
				},
			*/
		},
	})
}
```

### DIFF
--- a/session/driver/file.go
+++ b/session/driver/file.go
@@ -1,94 +1,214 @@
 package driver
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"sync"
 
+	sessioncontract "github.com/goravel/framework/contracts/session"
+	"github.com/goravel/framework/facades"
 	"github.com/goravel/framework/support/carbon"
 	"github.com/goravel/framework/support/file"
 )
 
+// File implements the session.Driver interface using the filesystem.
 type File struct {
-	path    string
-	minutes int
-	mu      sync.RWMutex
+	path    string       // Directory where session files are stored
+	minutes int          // Session lifetime in minutes
+	mu      sync.RWMutex // Protects access during file operations
 }
 
-func NewFile(path string, minutes int) *File {
+func newFile(path string, minutes int) (*File, error) {
+	if path == "" {
+		return nil, fmt.Errorf("session file path cannot be empty")
+	}
+
 	return &File{
 		path:    path,
 		minutes: minutes,
+	}, nil
+}
+
+// FileDriverFactory creates an instance of the file session driver using framework config.
+// This function matches the 'func() (session.Driver, error)' signature for config 'via'.
+func FileDriverFactory() (sessioncontract.Driver, error) {
+	config := facades.Config()
+
+	lifetime := config.GetInt("session.lifetime")
+	filesPath := config.GetString("session.files")
+
+	if filesPath == "" {
+		return nil, fmt.Errorf("session.files path is not configured; required for file session driver") // Return error: driver cannot function without a path
 	}
+
+	instance, err := newFile(filesPath, lifetime)
+	if err != nil {
+		return nil, err
+	}
+	return instance, nil
 }
 
 func (f *File) Close() error {
 	return nil
 }
 
+// Destroy removes a session file by its ID.
 func (f *File) Destroy(id string) error {
-	f.mu.Lock()
+	f.mu.Lock() // Exclusive lock for delete operation
 	defer f.mu.Unlock()
 
-	return file.Remove(f.getFilePath(id))
-}
+	if f.path == "" {
 
-func (f *File) Gc(maxLifetime int) error {
-	f.mu.Lock()
-	defer f.mu.Unlock()
-
-	cutoffTime := carbon.Now(carbon.UTC).SubSeconds(maxLifetime)
-
-	if !file.Exists(f.path) {
-		return nil
+		return fmt.Errorf("session path not configured") // Return error
 	}
 
-	return filepath.Walk(f.path, func(path string, info os.FileInfo, err error) error {
-		if err != nil {
-			return err
-		}
+	filePath := f.getFilePath(id)
 
-		if !info.IsDir() && info.ModTime().Before(cutoffTime.StdTime()) {
-			return os.Remove(path)
-		}
+	err := file.Remove(filePath)
+	// Log errors unless the file simply didn't exist
+	if err != nil && !os.IsNotExist(err) {
 
-		return nil
-	})
-}
-
-func (f *File) Open(string, string) error {
+		return fmt.Errorf("failed to destroy session file '%s': %w", id, err)
+	}
 	return nil
 }
 
+// Gc performs garbage collection, removing expired session files.
+func (f *File) Gc(maxLifetime int) error {
+	f.mu.Lock() // Exclusive lock for GC potentially deleting many files
+	defer f.mu.Unlock()
+
+	if f.path == "" {
+
+		return fmt.Errorf("session path not configured") // Return error
+	}
+	if maxLifetime <= 0 {
+
+		return fmt.Errorf("invalid maxLifetime for GC: %d", maxLifetime)
+	}
+
+	cutoffTime := carbon.Now(carbon.UTC).SubSeconds(maxLifetime)
+
+	var filesRemoved int
+	var errorsEncountered int
+
+	err := filepath.Walk(f.path, func(path string, info os.FileInfo, walkErr error) error {
+		if walkErr != nil {
+			// Log errors during walk (e.g., permissions) and potentially stop
+
+			errorsEncountered++
+			// Decide whether to stop walking on error. Returning the error stops.
+			// Returning nil allows it to potentially continue with other files/subdirs.
+			return nil
+		}
+
+		// Skip the root directory itself and any subdirectories immediately under it.
+		// Session files should be directly in f.path, not in subdirs.
+		if info.IsDir() {
+			if path == f.path {
+				return nil // Allow entering the root directory
+			}
+
+			return filepath.SkipDir
+		}
+
+		//Convert ModTime to utc
+		modTime := info.ModTime().UTC()
+
+		// Check modification time against cutoff for actual files
+		if modTime.Before(cutoffTime.StdTime()) {
+
+			removeErr := os.Remove(path)
+			if removeErr != nil && !os.IsNotExist(removeErr) {
+
+				errorsEncountered++
+				// Continue GC even if one file fails to delete
+			} else if removeErr == nil {
+				filesRemoved++
+			}
+		}
+		return nil // Continue walking
+	})
+
+	// Return the error from filepath.Walk if it terminated due to an issue.
+	if err != nil {
+
+		return fmt.Errorf("session GC failed for path '%s': %w", f.path, err)
+	}
+	return nil
+}
+
+// Open initializes the session driver. No action needed for file driver.
+func (f *File) Open(savePath string, sessionName string) error {
+	return nil
+}
+
+// Read retrieves session data from a file by ID. Returns empty string if not found or expired.
 func (f *File) Read(id string) (string, error) {
-	f.mu.RLock()
+	f.mu.RLock() // Read lock is sufficient
 	defer f.mu.RUnlock()
 
-	path := f.getFilePath(id)
-	if file.Exists(path) {
-		modified, err := file.LastModified(path, carbon.UTC)
+	if f.path == "" {
+
+		// Consistent behavior: return "" and no error if path is missing.
+		return "", nil
+	}
+
+	filePath := f.getFilePath(id)
+
+	// 1. Check existence first (optimization)
+	if !file.Exists(filePath) {
+		return "", nil
+	}
+
+	// 2. Check if expired (based on modification time)
+	if f.minutes > 0 {
+		modified, err := file.LastModified(filePath, carbon.UTC)
 		if err != nil {
-			return "", err
+
+			// Treat as unreadable if mod time fails
+			return "", fmt.Errorf("failed to check session expiry for '%s': %w", id, err)
 		}
-		if modified.After(carbon.Now(carbon.UTC).SubMinutes(f.minutes).StdTime()) {
-			data, err := os.ReadFile(path)
-			if err != nil {
-				return "", err
-			}
-			return string(data), nil
+
+		expiryTime := carbon.Now(carbon.UTC).SubMinutes(f.minutes)
+		if modified.Before(expiryTime.StdTime()) {
+
+			return "", nil
 		}
 	}
 
-	return "", nil
+	// 3. Read the file content
+	data, err := os.ReadFile(filePath)
+	if err != nil {
+
+		return "", fmt.Errorf("failed to read session data for '%s': %w", id, err)
+	}
+
+	return string(data), nil
 }
 
+// Write saves session data to a file by ID.
 func (f *File) Write(id string, data string) error {
-	f.mu.Lock()
+	f.mu.Lock() // Exclusive lock for write operation
 	defer f.mu.Unlock()
 
-	return file.PutContent(f.getFilePath(id), data)
+	if f.path == "" {
+		return fmt.Errorf("cannot write session file: session path is not configured")
+	}
+
+	filePath := f.getFilePath(id)
+
+	err := file.PutContent(filePath, data)
+	if err != nil {
+		return fmt.Errorf("failed to write session data for '%s': %w", id, err)
+	}
+	return nil
 }
 
 func (f *File) getFilePath(id string) string {
 	return filepath.Join(f.path, id)
 }
+
+// Ensure File implements the Driver interface at compile time.
+var _ sessioncontract.Driver = (*File)(nil)
--- a/session/driver/file_test.go
+++ b/session/driver/file_test.go
@@ -1,12 +1,14 @@
 package driver
 
 import (
+	"os"
+	"path/filepath"
 	"testing"
 
-	"github.com/stretchr/testify/suite"
-
+	sessioncontract "github.com/goravel/framework/contracts/session"
 	"github.com/goravel/framework/support/carbon"
 	"github.com/goravel/framework/support/file"
+	"github.com/stretchr/testify/suite"
 )
 
 type FileTestSuite struct {
@@ -17,12 +19,28 @@ func TestFileTestSuite(t *testing.T) {
 	suite.Run(t, &FileTestSuite{})
 }
 
-func (f *FileTestSuite) BeforeTest() {
-	f.Nil(file.Remove(f.getPath()))
+func (f *FileTestSuite) BeforeTest(suiteName, testName string) {
+
+	err := file.Remove(f.getPath())
+
+	f.Require().True(err == nil || os.IsNotExist(err), "Failed to clean test directory '%s' before test: %v", f.getPath(), err)
+}
+
+func (f *FileTestSuite) AfterTest(suiteName, testName string) {
+	err := file.Remove(f.getPath())
+	f.Require().True(err == nil || os.IsNotExist(err), "Failed to clean test directory '%s' after test: %v", f.getPath(), err)
+}
+
+func (f *FileTestSuite) TestNewFile_EmptyPathError() {
+	driver, err := newFile("", f.getMinutes())
+	f.Error(err, "newFile should return error for empty path")
+	f.Nil(driver, "Driver should be nil on error")
+	f.Contains(err.Error(), "session file path cannot be empty", "Error message mismatch")
 }
 
 func (f *FileTestSuite) TestClose() {
 	driver := f.getDriver()
+	f.Require().NotNil(driver)
 	f.Nil(driver.Close())
 }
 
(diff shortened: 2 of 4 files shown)


kkumar-gcc:
Looks like changes to the `file` driver are not required.

session/driver/file.go
```diff
@@ -1,94 +1,214 @@
 package driver
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"sync"
 
+	sessioncontract "github.com/goravel/framework/contracts/session"
```
hwbrzzl, line 9:
```suggestion
	contractssession "github.com/goravel/framework/contracts/session"
```

session/driver/file.go
```diff
@@ -1,94 +1,214 @@
...
+	}
+	return nil
 }
 
 func (f *File) getFilePath(id string) string {
 	return filepath.Join(f.path, id)
 }
+
+// Ensure File implements the Driver interface at compile time.
+var _ sessioncontract.Driver = (*File)(nil)
```
hwbrzzl, line 214: Please move this to the start of the file.

session/driver/file.go
```diff
@@ -1,94 +1,214 @@
...
+
 	return &File{
 		path:    path,
 		minutes: minutes,
+	}, nil
+}
+
+// FileDriverFactory creates an instance of the file session driver using framework config.
+// This function matches the 'func() (session.Driver, error)' signature for config 'via'.
+func FileDriverFactory() (sessioncontract.Driver, error) {
```
hwbrzzl, line 35: This function is unnecessary, `NewFile` is good enough.

session/driver/file.go
```diff
@@ -1,94 +1,214 @@
...
+	config := facades.Config()
```
hwbrzzl, line 36: `facades.Config()` should not be used in the internal framework. Please check other implementations; it should be passed from external.

session/driver/file.go
```diff
@@ -1,94 +1,214 @@
...
+	f.mu.Lock() // Exclusive lock for GC potentially deleting many files
 	defer f.mu.Unlock()
 
-	cutoffTime := carbon.Now(carbon.UTC).SubSeconds(maxLifetime)
+	if f.path == "" {
 
-	if !file.Exists(f.path) {
-		return nil
+		return fmt.Errorf("session path not configured") // Return error
 	}
```
hwbrzzl, line 85: It was checked when initiating, so these lines can be removed.

session/driver/file.go
```diff
@@ -1,94 +1,214 @@
...
+	if maxLifetime <= 0 {
 
-	return filepath.Walk(f.path, func(path string, info os.FileInfo, err error) error {
-		if err != nil {
-			return err
+		return fmt.Errorf("invalid maxLifetime for GC: %d", maxLifetime)
```
hwbrzzl, line 88: All framework errors are in `errors/list.go`, please move this as well.

session/driver/file.go
```diff
@@ -1,94 +1,214 @@
...
+				filesRemoved++
+			}
+		}
+		return nil // Continue walking
 	})
+
+	// Return the error from filepath.Walk if it terminated due to an issue.
+	if err != nil {
+
+		return fmt.Errorf("session GC failed for path '%s': %w", f.path, err)
```
hwbrzzl, line 137: Ditto

session/driver/file.go
```diff
@@ -1,94 +1,214 @@
...
-	return filepath.Walk(f.path, func(path string, info os.FileInfo, err error) error {
-		if err != nil {
-			return err
+		return fmt.Errorf("invalid maxLifetime for GC: %d", maxLifetime)
+	}
+
+	cutoffTime := carbon.Now(carbon.UTC).SubSeconds(maxLifetime)
+
+	var filesRemoved int
+	var errorsEncountered int
```
hwbrzzl, line 94: It seems they're only added, not used. Are they necessary? Can they be removed?

session/driver/file.go
```diff
@@ -1,94 +1,214 @@
...
+	if err != nil {
+
+		return fmt.Errorf("session GC failed for path '%s': %w", f.path, err)
+	}
+	return nil
 }
 
-func (f *File) Open(string, string) error {
+// Open initializes the session driver. No action needed for file driver.
+func (f *File) Open(savePath string, sessionName string) error {
```
hwbrzzl, line 143:
```suggestion
func (f *File) Open(_ string, _ string) error {
```

session/driver/file.go
```diff
@@ -1,94 +1,214 @@
...
 	defer f.mu.RUnlock()
 
-	path := f.getFilePath(id)
-	if file.Exists(path) {
-		modified, err := file.LastModified(path, carbon.UTC)
+	if f.path == "" {
+
+		// Consistent behavior: return "" and no error if path is missing.
+		return "", nil
+	}
```
hwbrzzl, line 156: Ditto

session/driver/file.go
```diff
@@ -1,94 +1,214 @@
...
+
+	// 2. Check if expired (based on modification time)
+	if f.minutes > 0 {
+		modified, err := file.LastModified(filePath, carbon.UTC)
 		if err != nil {
-			return "", err
+
+			// Treat as unreadable if mod time fails
+			return "", fmt.Errorf("failed to check session expiry for '%s': %w", id, err)
```
hwbrzzl, line 171: Ditto

session/driver/file.go
```diff
@@ -1,94 +1,214 @@
...
+			return "", nil
 		}
 	}
 
-	return "", nil
+	// 3. Read the file content
+	data, err := os.ReadFile(filePath)
+	if err != nil {
+
+		return "", fmt.Errorf("failed to read session data for '%s': %w", id, err)
```
hwbrzzl, line 185: Ditto

session/driver/file.go
```diff
@@ -1,94 +1,214 @@
...
+// Write saves session data to a file by ID.
 func (f *File) Write(id string, data string) error {
-	f.mu.Lock()
+	f.mu.Lock() // Exclusive lock for write operation
 	defer f.mu.Unlock()
 
-	return file.PutContent(f.getFilePath(id), data)
+	if f.path == "" {
+		return fmt.Errorf("cannot write session file: session path is not configured")
+	}
```
hwbrzzl, line 198: Ditto

session/driver/file.go
```diff
@@ -1,94 +1,214 @@
...
+
+	filePath := f.getFilePath(id)
+
+	err := file.PutContent(filePath, data)
+	if err != nil {
+		return fmt.Errorf("failed to write session data for '%s': %w", id, err)
```
hwbrzzl, line 204: Ditto

session/driver/file_test.go
```diff
@@ -1,12 +1,14 @@
 package driver
 
 import (
+	"os"
+	"path/filepath"
 	"testing"
 
-	"github.com/stretchr/testify/suite"
-
+	sessioncontract "github.com/goravel/framework/contracts/session"
```
hwbrzzl, line 8:
```suggestion
	contractssession "github.com/goravel/framework/contracts/session"
```

session/driver/file_test.go
```diff
@@ -17,12 +19,28 @@ func TestFileTestSuite(t *testing.T) {
 	suite.Run(t, &FileTestSuite{})
 }
 
-func (f *FileTestSuite) BeforeTest() {
-	f.Nil(file.Remove(f.getPath()))
+func (f *FileTestSuite) BeforeTest(suiteName, testName string) {
```
hwbrzzl, line 22:
```suggestion
func (f *FileTestSuite) SetupTest() {
```

session/driver/file_test.go
```diff
@@ -17,12 +19,28 @@ func TestFileTestSuite(t *testing.T) {
...
+
+	err := file.Remove(f.getPath())
+
+	f.Require().True(err == nil || os.IsNotExist(err), "Failed to clean test directory '%s' before test: %v", f.getPath(), err)
+}
+
+func (f *FileTestSuite) AfterTest(suiteName, testName string) {
```
hwbrzzl, line 29:
```suggestion
func (f *FileTestSuite) TearDownTest() {
```

session/driver/file_test.go
```diff
@@ -17,12 +19,28 @@ func TestFileTestSuite(t *testing.T) {
...
+	err := file.Remove(f.getPath())
```
hwbrzzl, line 30: It's unnecessar to remove the file in `Start` and `After`, only one is enough.

session/manager.go
```diff
@@ -1,84 +1,167 @@
 package session
 
 import (
+	"fmt"
 	"sync"
 	"time"
 
 	"github.com/goravel/framework/contracts/config"
 	"github.com/goravel/framework/contracts/foundation"
 	sessioncontract "github.com/goravel/framework/contracts/session"
```
hwbrzzl, line 10: Sorry, we are using `contractssession` due to `contracts/session`, it should be missed before. Could you optimize it here?

```suggestion
	contractssession "github.com/goravel/framework/contracts/session"
```

session/manager.go
```diff
@@ -1,84 +1,167 @@
...
+
+			continue
+		}
+
+		// Wrapper for Extend signature: func() sessioncontract.Driver
+		factoryWrapper := func() sessioncontract.Driver {
+			driverInstance, err := viaFactoryWithError()
+			if err != nil {
+
+				return nil
```
hwbrzzl, line 133: Should return the error here.

session/manager.go
```diff
@@ -1,84 +1,167 @@
...
+			}
+			if driverInstance == nil {
+
+				return nil
```
hwbrzzl, line 137: An error should be returned.

session/manager.go
```diff
@@ -1,84 +1,167 @@
...
+			}
+			return driverInstance
+		}
+
+		// Register using Extend. Errors logged internally by Extend.
+		_ = m.Extend(name, factoryWrapper)
```
hwbrzzl, line 143: Should catch the error.

session/manager.go
```diff
@@ -120,9 +193,12 @@ func (m *Manager) startGcTimer(driver sessioncontract.Driver) {
...
-			if err := driver.Gc(lifetime); err != nil {
+			if err := driverInstance.Gc(lifetime); err != nil {
 				color.Errorf("Error performing garbage collection: %s\n", err)
 			}
 		}
 	}()
 }
+
+// Ensure interface implementation
+var _ sessioncontract.Manager = (*Manager)(nil)
```
hwbrzzl, line 204: Ditto

session/manager.go
```diff
@@ -1,84 +1,167 @@
...
+			}
+			return driverInstance
+		}
+
+		// Register using Extend. Errors logged internally by Extend.
+		_ = m.Extend(name, factoryWrapper)
+	}
+}
+
+func (m *Manager) BuildSession(handler sessioncontract.Driver, sessionID ...string) (sessioncontract.Session, error) {
```
hwbrzzl, line 147: We set the location of functions based on the rule:

```
Public functions:
  A
  B

Private functions:
  a
  b
```

Hence, please don't move this function.

session/manager.go
```diff
@@ -1,84 +1,167 @@
...
+	driverInstance, instanceExists = m.drivers[driverName]
+	if instanceExists {
+		return driverInstance, nil
 	}
 
-	if m.drivers[driverName] == nil {
-		return nil, errors.SessionDriverNotSupported.Args(driverName)
+	driverInstance = factory()
+	if driverInstance == nil {
+		return nil, errors.New(fmt.Sprintf("Session driver %s factory returned nil", driverName))
```
hwbrzzl, line 80: Ditto

session/manager_test.go
```diff
@@ -2,24 +2,75 @@ package session
...
+}
+
+func MockDriverFactory(mockDriverInstance *MockSessionDriver) func() (sessioncontract.Driver, error) {
+	return func() (sessioncontract.Driver, error) {
+		if mockDriverInstance == nil {
+			return nil, fmt.Errorf("mock driver instance not provided to factory")
+		}
+		return mockDriverInstance, nil
+	}
+}
```
hwbrzzl, line 53: It's unnecessary, we can use the `mocks/session/Driver.go` file.

session/manager_test.go
```diff
@@ -28,173 +79,197 @@ func TestManagerTestSuite(t *testing.T) {
...
 	s.ErrorIs(err, errors.SessionDriverIsNotSet)
 	s.Nil(driver)
 }
 
 func (s *ManagerTestSuite) TestExtend() {
-	s.mockConfig.On("GetInt", "session.gc_interval", 30).Return(30).Once()
-	err := s.manager.Extend("test", NewCustomDriver)
+
+	s.mockConfig.On("GetString", "session.driver").Return("file").Maybe()
+	s.mockConfig.On("GetInt", "session.gc_interval").Return(30).Maybe()
```
hwbrzzl, line 172: If we can, `Maybe` should not be used.

session/manager_test.go
```diff
@@ -28,173 +79,197 @@ func TestManagerTestSuite(t *testing.T) {
...
 	s.Equal("*session.CustomDriver", fmt.Sprintf("%T", driver))
+}
+
+func (s *ManagerTestSuite) TestDriver_NotSupported() {
 
-	// not supported a driver
-	s.mockConfig.On("GetString", "session.driver").Return("not_supported").Once()
-	driver, err = s.manager.Driver()
+	s.mockConfig.ExpectedCalls = nil
+	s.mockConfig.On("GetString", "session.driver").Return("not_supported")
```
hwbrzzl, line 149:
```suggestion
	s.mockConfig.On("GetString", "session.driver").Return("not_supported").Once()
```

hwbrzzl (commented):
Thanks, a great PR. For the file driver, we don't need to add `"via": fileriver.FileDriverFactory,` in the configuration file, it's an internal driver. User can add `drivers` when using a custom driver.

hwbrzzl:
> It looks like no changes are needed for the `file` driver. We already support custom drivers: https://www.goravel.dev/the-basics/session.html#add-custom-session-drivers.

It seems only structure is optimized, logic is not changed.

session/driver/file.go
```diff
@@ -1,94 +1,214 @@
...
-		modified, err := file.LastModified(path, carbon.UTC)
+	if f.path == "" {
+
+		// Consistent behavior: return "" and no error if path is missing.
+		return "", nil
+	}
+
+	filePath := f.getFilePath(id)
+
+	// 1. Check existence first (optimization)
```
kkumar-gcc, line 160: Let's consider if all these comments are really necessary

Ref: https://refactoring.guru/smells/comments

merouanekhalili:
Hi Team,

I reverted the file driver back to the original content as you suggested, I only updated the GC to convert info.ModTime to UTC before checking
I also made sure all suggestions are implemented

Thanks

merouanemc:
Hi Team,

Just for information I just pushed the redis driver for Session module :

https://github.com/merouanekhalili/session-redis-driver

Thanks

merouanekhalili:
Hi Team,

Just for information I just pushed the redis driver for Session module :

https://github.com/merouanekhalili/session-redis-driver

Thanks

session/driver/file.go
```diff
@@ -48,7 +48,7 @@ func (f *File) Gc(maxLifetime int) error {
 			return err
 		}
 
-		if !info.IsDir() && info.ModTime().Before(cutoffTime.StdTime()) {
+		if !info.IsDir() && info.ModTime().UTC().Before(cutoffTime.StdTime()) {
```
hwbrzzl, line 51: Good catch

session/manager.go
```diff
@@ -51,65 +59,139 @@ func (m *Manager) BuildSession(handler sessioncontract.Driver, sessionID ...stri
...
+
+	driverInstance, instanceExists = m.drivers[driverName]
+	if instanceExists {
+		return driverInstance, nil
+	}
+
+	if factoryExists {
+		driverInstance = factory() // Call the factory registered via Extend/config
+		if driverInstance == nil {
+			return nil, errors.New(fmt.Sprintf("Factory for session driver '%s' returned nil", driverName))
```
hwbrzzl, line 95: Please move the error to `errors/list.go`

session/manager.go
```diff
@@ -51,65 +59,139 @@ func (m *Manager) BuildSession(handler sessioncontract.Driver, sessionID ...stri
...
+		viaFactoryWithError, ok := viaFactoryAny.(func() (contractssession.Driver, error))
+		if !ok {
+			continue
+		}
 
-func (m *Manager) createFileDriver() sessioncontract.Driver {
-	lifetime := m.config.GetInt("session.lifetime")
-	return driver.NewFile(m.config.GetString("session.files"), lifetime)
+		// Wrapper for Extend signature: func() contractssession.Driver
+		factoryWrapper := func() contractssession.Driver {
```
hwbrzzl, line 172: The error should be returned instead of `color.Errorf`.

```suggestion
		factoryWrapper := func() (contractssession.Driver, error) {
```

session/manager.go
```diff
@@ -51,65 +59,139 @@ func (m *Manager) BuildSession(handler sessioncontract.Driver, sessionID ...stri
...
+				color.Errorf("Driver instance for '%s' is nil\n", name)
+				return nil
+			}
+			return driverInstance
+		}
+
+		err := m.Extend(name, factoryWrapper)
+		if err != nil {
+			color.Errorf("Failed to register driver '%s': %s\n", name, err)
+		}
```
hwbrzzl, line 188:
```suggestion
		return m.Extend(name, factoryWrapper)
```

hwbrzzl:
> Hi Team,
> 
> Just for information I just pushed the redis driver for Session module :
> 
> https://github.com/merouanekhalili/session-redis-driver
> 
> Thanks

Amazing. Could you merge it into https://github.com/goravel/redis? We can set it as an official implementation.

session/manager.go
```diff
@@ -51,65 +60,106 @@ func (m *Manager) BuildSession(handler sessioncontract.Driver, sessionID ...stri
...
 
+func (m *Manager) custom(driver string) (contractssession.Driver, error) {
+	if custom, ok := m.config.Get(fmt.Sprintf("session.drivers.%s.via", driver)).(contractssession.Driver); ok {
+		return custom, nil
+	}
+	if custom, ok := m.config.Get(fmt.Sprintf("session.drivers.%s.via", driver)).(func() (contractssession.Driver, error)); ok {
+		return custom()
+	}
+
+	return nil, errors.CacheStoreContractNotFulfilled.Args(driver)
```
hwbrzzl, line 123: The error is incorrect.

session/manager.go
```diff
@@ -51,65 +60,106 @@ func (m *Manager) BuildSession(handler sessioncontract.Driver, sessionID ...stri
...
-}
+	for name := range driversMap {
 
-func (m *Manager) createFileDriver() sessioncontract.Driver {
-	lifetime := m.config.GetInt("session.lifetime")
-	return driver.NewFile(m.config.GetString("session.files"), lifetime)
+		driver := m.config.GetString(fmt.Sprintf("session.drivers.%s.driver", name))
+
+		switch driver {
+		case "custom":
```
hwbrzzl, line 146: `case "file"` can be added here. And the file driver can be added in the configuration file:

```
drivers: 
  file:
    driver: file
```

hwbrzzl (commented):
Great, almost complete. 👍

hwbrzzl (approved):
Thanks, perfect!

PR #1011, opened 2025-04-16
feat: [#649]  auto-register migration and seeder when created via CLI

almas1992:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/649

This change improves workflow efficiency by automatically registering migration and seeder files created via artisan `make:migration` and `artisan make:seeder`.

These files are now injected into database/kernel.go, eliminating the need for manual inclusion and streamlining the setup process.

## ✅ Checks

### DIFF
--- a/database/console/migration/migrate_make_command.go
+++ b/database/console/migration/migrate_make_command.go
@@ -2,11 +2,16 @@ package migration
 
 import (
 	"fmt"
+	"runtime/debug"
 
 	"github.com/goravel/framework/contracts/console"
 	"github.com/goravel/framework/contracts/console/command"
 	"github.com/goravel/framework/contracts/database/migration"
 	"github.com/goravel/framework/errors"
+	"github.com/goravel/framework/packages/match"
+	"github.com/goravel/framework/packages/modify"
+	"github.com/goravel/framework/support/path"
+	"github.com/goravel/framework/support/str"
 )
 
 type MigrateMakeCommand struct {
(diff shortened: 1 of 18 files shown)


database/console/migration/migrate_make_command.go
```diff
@@ -57,12 +62,25 @@ func (r *MigrateMakeCommand) Handle(ctx console.Context) error {
...
 		ctx.Error(errors.MigrationCreateFailed.Args(err).Error())
 		return nil
 	}
 
 	ctx.Success(fmt.Sprintf("Created Migration: %s", name))
 
+	info, _ := debug.ReadBuildInfo()
+	structName := str.Of(fileName).Prepend("m_").Studly().String()
+	if err = modify.GoFile(path.Database("kernel.go")).
+		Find(match.Imports()...).Modify(modify.AddImport(fmt.Sprintf("%s/database/migrations", info.Main.Path))).
```
hwbrzzl, line 76: Is it possible to remove the `...`? `Find(match.Imports())`. All `Imports`, `Migrations`, etc. need to use `...` when are called.

database/console/migration/migrate_make_command_test.go
hwbrzzl, line 1: Could you add a `Migration registered successfully` test case?

support/console/console_test.go
```diff
@@ -48,6 +58,16 @@ func (s *MakeTestSuite) TestGetStructName() {
 	s.Equal("Lowercase", s.make.GetStructName())
 }
 
+func (s *MakeTestSuite) TestGetImportPath() {
```
hwbrzzl, line 61:
```suggestion
func (s *MakeTestSuite) TestGetPackageImportPath() {
```

hwbrzzl (commented):
Nice

database/console/migration/migrate_make_command.go
almas1992: Yes, it’s possible, but we’d need to change the parameter type of the `Find` function from `...match.GoNode` to `[]match.GoNode`. After that change, if we want to pass a single matcher to `Find`, we’ll have to wrap it like this: `[]match.GoNode{xxxxx}`. Should we go ahead and make that change?

database/console/migration/migrate_make_command_test.go
almas1992: Okay

hwbrzzl (approved):
Great 👍

PR #664, opened 2024-10-03
feat: [#280] Manage migrations table

hwbrzzl:
## 📑 Description

https://github.com/goravel/goravel/issues/280

1. Implement `GetTables` and `HasTable`;
2. Implement managing repository(the migrations table), will be used when running a migration file;

## Summary by CodeRabbit

- **New Features**
	- Introduced new methods for managing database tables, including `CompileTables`, `GetTables`, and `HasTable`.
	- Added error handling to existing methods like `Create` and `DropIfExists`.
	- New struct `File` added to enhance migration functionality.
	- Simplified the `Blueprint` and `Schema` structures for improved clarity.

- **Bug Fixes**
	- Updated method signatures to improve error handling and return types across various interfaces.

- **Tests**
	- Added a comprehensive test suite for repository functionality, ensuring robust testing of migration operations.
	- Restructured test suite for better handling of database queries.

- **Chores**
	- Streamlined the initialization process for schema and migration, improving overall code clarity and maintainability.

## ✅ Checks

- [x] Added test cases for my code

PR #1020, opened 2025-04-29
fix: [#664] resolve date mismatch between log file name and log entries

almas1992:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/644

## ✅ Checks

### DIFF
--- a/log/application_test.go
+++ b/log/application_test.go
@@ -18,19 +18,19 @@ func TestNewApplication(t *testing.T) {
 	assert.Nil(t, err)
 	assert.NotNil(t, app)
 
-	mockConfig := &mocksconfig.Config{}
-	mockConfig.On("GetString", "logging.default").Return("test")
-	mockConfig.On("GetString", "logging.channels.test.driver").Return("single")
-	mockConfig.On("GetString", "logging.channels.test.path").Return("test")
-	mockConfig.On("GetString", "logging.channels.test.level").Return("debug")
-	mockConfig.On("GetBool", "logging.channels.test.print").Return(true)
+	mockConfig := mocksconfig.NewConfig(t)
+	mockConfig.EXPECT().GetString("logging.default").Return("test").Once()
+	mockConfig.EXPECT().GetString("logging.channels.test.driver").Return("single").Once()
+	mockConfig.EXPECT().GetString("logging.channels.test.path").Return("test").Once()
+	mockConfig.EXPECT().GetString("logging.channels.test.level").Return("debug").Once()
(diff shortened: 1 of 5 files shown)


log/logger/daily.go
```diff
@@ -44,7 +44,7 @@ func (daily *Daily) Handle(channel string) (logrus.Hook, error) {
 		logPath+"-%Y-%m-%d"+ext,
 		rotatelogs.WithRotationTime(time.Duration(24)*time.Hour),
 		rotatelogs.WithRotationCount(uint(daily.config.GetInt(channel+".days"))),
-		rotatelogs.WithClock(rotatelogs.NewClock(carbon.Now().StdTime())),
+		rotatelogs.WithClock(rotatelogs.NewClock(carbon.Now(daily.config.GetString("app.timezone")).StdTime())),
```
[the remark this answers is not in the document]
almas1992: No need, it’s already handled in carbon.New.

log/application_test.go
```diff
@@ -18,72 +18,74 @@ func TestNewApplication(t *testing.T) {
...
 
-	mockConfig := &mocksconfig.Config{}
-	mockConfig.On("GetString", "logging.default").Return("test")
-	mockConfig.On("GetString", "logging.channels.test.driver").Return("single")
-	mockConfig.On("GetString", "logging.channels.test.path").Return("test")
-	mockConfig.On("GetString", "logging.channels.test.level").Return("debug")
-	mockConfig.On("GetBool", "logging.channels.test.print").Return(true)
+	mockConfig := mocksconfig.NewConfig(t)
+	mockConfig.EXPECT().GetString("logging.default").Return("test").Once()
+	mockConfig.EXPECT().GetString("logging.channels.test.driver").Return("single").Once()
```
hwbrzzl, line 23: Nice 👍

log/logger/daily.go
hwbrzzl, line 47: Should we use [SetTimezone](https://github.com/goravel/framework/blob/master/support/carbon/carbon.go#L35) to set the global timezone?

almas1992: SetTimezone is indeed called during the framework initialization (see [here](https://github.com/goravel/framework/blob/ee0bc553a6b3ec050155f1b38c58e87bc6d1a9b2/foundation/application.go#L68C2-L68C17)). However, its position after registerConfiguredServiceProviders is what causes the issue.

hwbrzzl: Good catch, I think `r.setTimezone()` should be moved behind `r.registerConfiguredServiceProviders()`.

hwbrzzl (approved):
Great PR 👍

PR #1049, opened 2025-05-26
feat: [#635] Optimize the process of generating migration and model (#2)

kkumar-gcc:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/635

## ✅ Checks

### DIFF
--- a/contracts/database/schema/schema.go
+++ b/contracts/database/schema/schema.go
@@ -46,6 +46,8 @@ type Schema interface {
 	GetViews() ([]driver.View, error)
 	// GoTypes returns the mapping of schema types to Go types.
 	GoTypes() []GoType
+	// GetModel Get the model from the registered models by name.
+	GetModel(name string) Model
 	// HasColumn Determine if the given table has a given column.
 	HasColumn(table, column string) bool
 	// HasColumns Determine if the given table has given columns.
@@ -91,6 +93,7 @@ type Connection interface {
 // Extension represents an extension for the schema
 type Extension struct {
 	GoTypes []GoType
+	Models  []Model
 }
 
 // GoType represents a database column type to Go type mapping
@@ -101,4 +104,17 @@ type GoType struct {
 	Import     string
 	NullType   string
 	NullImport string
+}
+
+type Model struct {
+	Name string
+	Type any
+}
+
+func (r Model) IsZero() bool {
+	return r.Name == "" && r.Type == nil
+}
+
+func (r Model) IsValid() bool {
+	return r.Name != "" && r.Type != nil
 }
--- a/database/schema/schema.go
+++ b/database/schema/schema.go
@@ -25,6 +25,7 @@ type Schema struct {
 	processor  driver.Processor
 	schema     string
 	goTypes    []contractsschema.GoType
+	models     []contractsschema.Model
 }
 
 func NewSchema(config config.Config, log log.Log, orm contractsorm.Orm, driver driver.Driver, migrations []contractsschema.Migration) (*Schema, error) {
@@ -49,6 +50,7 @@ func NewSchema(config config.Config, log log.Log, orm contractsorm.Orm, driver d
 		processor:  processor,
 		schema:     schema,
 		goTypes:    defaultGoTypes(),
+		models:     make([]contractsschema.Model, 0),
 	}, nil
 }
 
@@ -170,6 +172,7 @@ func (r *Schema) DropIfExists(table string) error {
 
 func (r *Schema) Extend(extend contractsschema.Extension) contractsschema.Schema {
 	r.extendGoTypes(extend.GoTypes)
+	r.extendModels(extend.Models)
 	return r
 }
 
@@ -290,6 +293,16 @@ func (r *Schema) GetViews() ([]driver.View, error) {
 
 func (r *Schema) GoTypes() []contractsschema.GoType {
 	return r.goTypes
+}
+
+func (r *Schema) GetModel(name string) contractsschema.Model {
+	for _, model := range r.models {
+		if model.Name == name {
+			return model
+		}
+	}
+
+	return contractsschema.Model{}
 }
 
 func (r *Schema) HasColumn(table, column string) bool {
@@ -474,6 +487,20 @@ func (r *Schema) extendGoTypes(overrides []contractsschema.GoType) {
 	}
 
 	r.goTypes = result
+}
+
+func (r *Schema) extendModels(models []contractsschema.Model) {
+	if len(models) == 0 {
+		return
+	}
+
+	for _, model := range models {
+		if !slices.ContainsFunc(r.models, func(m contractsschema.Model) bool {
+			return m.Name == model.Name
+		}) {
+			r.models = append(r.models, model)
+		}
+	}
 }
 
 func defaultGoTypes() []contractsschema.GoType {
--- a/support/structmeta/structmeta.go
+++ b/support/structmeta/structmeta.go
@@ -1,0 +1,177 @@
+// Package structmeta is only for internal use in the Goravel framework. It provides
+// functionality to introspect Go structs and their methods, extracting metadata
+// such as field names, types, tags, and method signatures. This is useful for
+// performing reflection-based operations on structs, such as generating
+// migrations from model definitions or binding methods to runtime instances.
+package structmeta
+
+import (
+	"reflect"
+)
+
+const maxEmbeddedParseDepth = 2
+
+// WalkStruct inspects "v" (which may be a struct value, a *struct, or even a
+// nil *struct) and builds a StructMetadata.  If v is not a struct (or pointer to one),
+// it returns an empty StructMetadata.
+//
+// For fields: it flattens up to two levels of embedded structs. For methods:
+//   - It finds every method declared on T (value-receiver) and *T (pointer-receiver).
+//   - It then tries to "bind" that method onto the runtime instance (v). If v was a
+//     non-addressable struct literal or an interface holding a struct, it will make
+//     a temporary *T copy so pointer-receivers still work.
+//   - If v is a nil pointer, pointer-receiver bindings will remain invalid.
+func WalkStruct(v interface{}) StructMetadata {
+	if v == nil {
+		return StructMetadata{}
+	}
+
+	valV := reflect.ValueOf(v)
+	typV := valV.Type()
+
+	var typT reflect.Type
+	switch typV.Kind() {
+	case reflect.Ptr:
+		if typV.Elem().Kind() != reflect.Struct {
+			return StructMetadata{}
+		}
+		typT = typV.Elem()
+
+	case reflect.Struct:
+		typT = typV
+
+	default:
+		return StructMetadata{}
+	}
+
+	structName := typT.String()
+	fields := parseFields(typT, 0, nil)
+
+	var methods []MethodMetadata
+
+	bindMethod := func(methodName string, wantPtrReceiver bool) reflect.Value {
+		if wantPtrReceiver && valV.Kind() == reflect.Ptr {
+			if m := valV.MethodByName(methodName); m.IsValid() {
+				return m
+			}
+		}
+		if !wantPtrReceiver {
+			if m := valV.MethodByName(methodName); m.IsValid() {
+				return m
+			}
+		}
+		if wantPtrReceiver && valV.Kind() == reflect.Struct && valV.CanAddr() {
+			addrV := valV.Addr()
+			if m := addrV.MethodByName(methodName); m.IsValid() {
+				return m
+			}
+		}
+		if wantPtrReceiver && valV.Kind() == reflect.Struct && !valV.CanAddr() {
+			tmpPtr := reflect.New(valV.Type())
+			tmpPtr.Elem().Set(valV)
+			if m := tmpPtr.MethodByName(methodName); m.IsValid() {
+				return m
+			}
+		}
+		return reflect.Value{}
+	}
+
+	for i := 0; i < typT.NumMethod(); i++ {
+		mDef := typT.Method(i)
+		mType := mDef.Type
+		var params []string
+		for j := 1; j < mType.NumIn(); j++ {
+			params = append(params, mType.In(j).String())
+		}
+		var returns []string
+		for j := 0; j < mType.NumOut(); j++ {
+			returns = append(returns, mType.Out(j).String())
+		}
+
+		bound := bindMethod(mDef.Name, false /*value receiver*/)
+		methods = append(methods, MethodMetadata{
+			Name:         mDef.Name,
+			Receiver:     mType.In(0).String(),
+			Parameters:   params,
+			Returns:      returns,
+			ReflectValue: bound,
+		})
+	}
+
+	ptrType := reflect.PointerTo(typT)
+	for i := 0; i < ptrType.NumMethod(); i++ {
+		mDef := ptrType.Method(i)
+		mType := mDef.Type
+
+		name := mDef.Name
+		already := false
+		for _, prev := range methods {
+			if prev.Name == name {
+				already = true
+				break
+			}
+		}
+		if already {
+			continue
+		}
+
+		var params []string
+		for j := 1; j < mType.NumIn(); j++ {
+			params = append(params, mType.In(j).String())
+		}
+		var returns []string
+		for j := 0; j < mType.NumOut(); j++ {
+			returns = append(returns, mType.Out(j).String())
+		}
+
+		bound := bindMethod(mDef.Name, true /*pointer receiver*/)
+		methods = append(methods, MethodMetadata{
+			Name:         mDef.Name,
+			Receiver:     mType.In(0).String(),
+			Parameters:   params,
+			Returns:      returns,
+			ReflectValue: bound,
+		})
+	}
+
+	return StructMetadata{
+		Name:         structName,
+		Fields:       fields,
+		Methods:      methods,
+		ReflectValue: valV,
+	}
+}
+
+func parseFields(t reflect.Type, depth int, basePath []int) []FieldMetadata {
+	var fields []FieldMetadata
+	if t.Kind() != reflect.Struct {
+		return fields
+	}
+
+	for i := 0; i < t.NumField(); i++ {
+		f := t.Field(i)
+		indexPath := append(append([]int(nil), basePath...), i)
+
+		meta := FieldMetadata{
+			Name:      f.Name,
+			Type:      f.Type.String(),
+			Kind:      f.Type.Kind(),
+			Anonymous: f.Anonymous,
+			Tag:       NewTagMetadata(f.Tag),
+			IndexPath: indexPath,
+		}
+		fields = append(fields, meta)
+
+		if f.Anonymous && depth < maxEmbeddedParseDepth {
+			ft := f.Type
+			if ft.Kind() == reflect.Ptr {
+				ft = ft.Elem()
+			}
+			if ft.Kind() == reflect.Struct {
+				embedded := parseFields(ft, depth+1, indexPath)
+				fields = append(fields, embedded...)
+			}
+		}
+	}
+	return fields
+}
--- a/contracts/database/migration/migrator.go
+++ b/contracts/database/migration/migrator.go
@@ -13,7 +13,7 @@ type Status struct {
 
 type Migrator interface {
 	// Create a new migration file.
-	Create(name string) (string, error)
+	Create(name string, modelName string) (string, error)
 	// Fresh the migrations.
 	Fresh() error
 	// Reset the migrations.
--- a/database/schema/schema_test.go
+++ b/database/schema/schema_test.go
@@ -8,6 +8,25 @@ import (
 	contractsschema "github.com/goravel/framework/contracts/database/schema"
 )
 
+type User struct {
+	ID int64 `gorm:"primaryKey"`
+}
+
+type Address struct {
+	ID int64 `gorm:"primaryKey"`
+}
+
+type Relation struct {
+	UserID    int64 `gorm:"primaryKey"`
+	AddressID int64 `gorm:"primaryKey"`
+}
+
+func defaultModels() []contractsschema.Model {
+	return []contractsschema.Model{
+		{Name: "User", Type: &User{}},
+	}
+}
+
 type SchemaTestSuite struct {
 	suite.Suite
 }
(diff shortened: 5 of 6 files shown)


hwbrzzl:
Hi @kkumar-gcc Thanks for your great PR, and FYI, v1.16 is going to be released in about one/two weeks. It would be great if it can be merged before that. Or v1.17 is fine as well.

database/schema/schema.go
```diff
@@ -498,6 +512,39 @@ func (r *Schema) extendGoTypes(overrides []contractsschema.GoType) {
 	r.goTypes = result
 }
 
+func (r *Schema) extendModels(models []contractsschema.Model) {
```
krishankumar01: The model registration (discovery) can be optimized further. Currently, it registers model instances, but this can be improved. Looking for more suggestions on how to make this process cleaner and more developer friendly.

contracts/database/schema/schema.go
```diff
@@ -46,6 +46,8 @@ type Schema interface {
 	GetViews() ([]driver.View, error)
 	// GoTypes returns the mapping of schema types to Go types.
 	GoTypes() []GoType
+	// GetModel Get the model from the registered models by name.
+	GetModel(name string) Model
```
hwbrzzl: Please sort it.

database/migration/stubs.go
hwbrzzl, line 1: Nice 💯

database/migration/stubs.go
```diff
@@ -4,85 +4,95 @@ type Stubs struct {
...
+func (r *{{.StructName}}) Up() error {
+	if !facades.Schema().HasTable("{{.Table}}") {
+		return facades.Schema().Create("{{.Table}}", func(table schema.Blueprint) {
+			{{- if .SchemaFields}}
+			{{range .SchemaFields}}
+			{{.}}
+			{{- end}}
+			{{- else}}
 			table.ID()
 			table.TimestampsTz()
```
hwbrzzl, line 53: 1. The columns are missing. 
2. The file name and the migration name are unexpected.

Image

database/migration/migrator_test.go
hwbrzzl, line 1: Please add some test cases for the new logic.

database/migration/model_mapper.go
```diff
@@ -0,0 +1,772 @@
...
+	"strconv"
+	"strings"
+	"unicode"
+
+	"github.com/goravel/framework/support/str"
+	"github.com/goravel/framework/support/structmeta"
+)
+
+var (
+	errInvalidModel = errors.New("model must be a struct or pointer to struct")
```
hwbrzzl: Move it to `errors/list.go`

database/schema/schema.go
hwbrzzl: It's unfriendly to define a schema.Model that is distinguished by `db.Model`. Can `db.Model` be used directly? Then use `%T` to get the Model name.

database/migration/migrator.go
```diff
@@ -29,16 +34,51 @@ func NewMigrator(artisan console.Artisan, schema contractsschema.Schema, table s
...
 }
 
-func (r *Migrator) Create(name string) (string, error) {
+func (r *Migrator) Create(name string, modelName string) (string, error) {
 	table, create := TableGuesser{}.Guess(name)
 
+	var schemaFields []string
+	if modelName != "" {
+		if model := r.schema.GetModel(modelName); model.IsValid() {
+			table, schemaFields, _ = Generate(model.Type)
```
hwbrzzl: The error should not be ignored.

database/schema/schema.go
hwbrzzl: And users have to register the model via `facades.Schema().Extend()` when they want to use this feature, it's a bit complicated, not sure if this step can be omitted. If cannot, a warning message may be printed to prompt users to register it first.

support/structmeta/structmeta_test.go
```diff
@@ -0,0 +1,132 @@
...
+		assert.Equal(t, expect.Receiver, m.Receiver)
+		assert.Equal(t, expect.Params, m.Parameters)
+		assert.Equal(t, expect.Returns, m.Returns)
+		assert.Equal(t, expect.IsCallable, m.ReflectValue.IsValid(), "%s callable", m.Name)
+		delete(methods, m.Name)
+	}
+
+	if len(methods) > 0 {
+		t.Errorf("missing expected methods: %+v", methods)
+	}
```
hwbrzzl: Mistake?

support/structmeta/structmeta_test.go
```diff
@@ -0,0 +1,132 @@
...
+func TestWalkStruct_EmbeddedFields(t *testing.T) {
+	meta := WalkStruct(embedLevel3{})
+
+	names := make([]string, len(meta.Fields))
+	for i, f := range meta.Fields {
+		names[i] = f.Name
+	}
+
+	expected := []string{"embedLevel2", "embedLevel1", "Level1", "Level2", "Level3"}
+	assert.Equal(t, expected, names)
```
hwbrzzl: Could you assert all fields instead of only the name?

support/structmeta/structmeta.go
```diff
@@ -0,0 +1,178 @@
...
+// nil *struct) and builds a StructMetadata.  If v is not a struct (or pointer to one),
+// it returns an empty StructMetadata.
+//
+// For fields: it flattens up to two levels of embedded structs. For methods:
+//   - It finds every method declared on T (value-receiver) and *T (pointer-receiver).
+//   - It then tries to "bind" that method onto the runtime instance (v). If v was a
+//     non-addressable struct literal or an interface holding a struct, it will make
+//     a temporary *T copy so pointer-receivers still work.
+//   - If v is a nil pointer, pointer-receiver bindings will remain invalid.
+func WalkStruct(v interface{}) StructMetadata {
```
hwbrzzl:
```suggestion
func WalkStruct(v any) StructMetadata {
```

support/structmeta/structmeta_test.go
```diff
@@ -0,0 +1,132 @@
...
+	assert.Len(t, meta.Fields, 2)
+}
+
+func TestWalkStruct_NonStructInput(t *testing.T) {
+	assert.Empty(t, WalkStruct(nil).Fields)
+	assert.Empty(t, WalkStruct(42).Fields)
+	assert.Empty(t, WalkStruct("not a struct").Fields)
+}
+
+func TestNewTagMetadata(t *testing.T) {
```
hwbrzzl: Move this case to `type_test.go`

support/structmeta/structmeta.go
```diff
@@ -0,0 +1,178 @@
...
+			tmpPtr := reflect.New(valV.Type())
+			tmpPtr.Elem().Set(valV)
+			if m := tmpPtr.MethodByName(methodName); m.IsValid() {
+				return m
+			}
+		}
+		return reflect.Value{}
+	}
+
+	for i := 0; i < typT.NumMethod(); i++ {
```
hwbrzzl: Can these code be moved to a `parseMethods` function?

database/migration/model_mapper.go
```diff
@@ -0,0 +1,772 @@
...
+func scanEmbedded(meta structmeta.StructMetadata) map[string]bool {
+	embedded := make(map[string]bool, embeddedFieldCapacity)
+
+	for _, field := range meta.Fields {
+		if !field.Anonymous {
+			continue
+		}
+
+		t := field.Type
+		if t == "gorm.Model" || t == "*gorm.Model" || t == "orm.Model" || t == "*orm.Model" {
```
hwbrzzl: Can we get the string dynamically instead of the fixed string? To avoid the package being modified.

database/migration/model_mapper.go
```diff
@@ -0,0 +1,772 @@
...
+			if elem != nil {
+				kind = elem.Kind()
+			}
+		}
+	}
+
+	return kind == reflect.Struct
+}
+
+func scanEmbedded(meta structmeta.StructMetadata) map[string]bool {
```
hwbrzzl: The returned value can be a struct instead of `map[string]bool`.

database/migration/model_mapper.go
```diff
@@ -0,0 +1,772 @@
...
+
+	return !isRelation(field)
+}
+
+func isRelation(field *structmeta.FieldMetadata) bool {
+	// Time types are not relations despite being structs
+	t := field.Type
+	if strings.Contains(t, "time.Time") || strings.Contains(t, "gorm.DeletedAt") {
+		return false
+	}
```
hwbrzzl: Can we get the string dynamically as well (use %T or reflect)? And move them to a list at the top of this file. The `carbon` type is missing.

database/migration/model_mapper.go
```diff
@@ -0,0 +1,772 @@
...
+	s.hasID = embedded["ID"]
+	s.timestamps = embedded["CreatedAt"] && embedded["UpdatedAt"]
+	s.softDelete = embedded["DeletedAt"]
+
+	for _, field := range meta.Fields {
+		if !shouldInclude(&field, embedded) {
+			continue
+		}
+
+		if field.Name == "ID" {
```
hwbrzzl: We need to use `gorm:primaryKey` to judge the `ID` instead of the name.

database/migration/model_mapper.go
```diff
@@ -0,0 +1,772 @@
...
+		}
+
+		if idx := strings.IndexByte(part, ':'); idx != -1 {
+			key := strings.ToLower(strings.TrimSpace(part[:idx]))
+			val := strings.TrimSpace(part[idx+1:])
+			parseKeyValue(t, key, val)
+		} else {
+			parseKeyValue(t, strings.ToLower(part), "")
+		}
+	}
```
hwbrzzl: The tag info should be parsed in `FieldMetadata`, correct?

database/migration/model_mapper.go
```diff
@@ -0,0 +1,772 @@
...
+
+	if goType == "string" {
+		if t.size > 0 {
+			return "String", []any{t.size}
+		}
+		return "Text", nil
+	}
+
+	switch goType {
```
hwbrzzl: Could you move these cases to a map at the top of this file?

database/migration/model_mapper.go
```diff
@@ -0,0 +1,772 @@
...
+		return
+	}
+
+	tagStr := field.Tag.Get("gorm")
+	if tagStr == "" {
+		return
+	}
+
+	parts := strings.Split(tagStr, ";")
+	for _, part := range parts {
```
hwbrzzl: The tag info should be parsed in `FieldMetadata`, correct?

database/migration/model_mapper.go
```diff
@@ -0,0 +1,772 @@
...
+			continue
+		}
+
+		col := buildColumn(&field)
+		if col == nil {
+			continue
+		}
+
+		s.columns = append(s.columns, *col)
+		collectIndexes(&field, col.name, s.indexes)
```
hwbrzzl:
```suggestion
		s.indexes = collectIndexes(&field, col.name)
```

database/migration/model_mapper.go
hwbrzzl, line 1: This file is amazing, we 'd better to add more test cases for the internal functions.

database/migration/model_mapper.go
```diff
@@ -0,0 +1,772 @@
...
+		if !field.Anonymous {
+			continue
+		}
+
+		t := field.Type
+		if t == "gorm.Model" || t == "*gorm.Model" || t == "orm.Model" || t == "*orm.Model" {
+			embedded["ID"] = true
+			embedded["CreatedAt"] = true
+			embedded["UpdatedAt"] = true
+			embedded["DeletedAt"] = true
```
hwbrzzl: `DeletedAt` should not be here.

database/migration/model_mapper_test.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+type BasicGormModel struct {
+	orm.Model
+	Name       string
+	Email      *string `gorm:"size:255"`
+	unexported string
+}
+
+type AllFeatures struct {
```
hwbrzzl: Add a column for Nullable.

database/migration/stubs.go
```diff
@@ -4,85 +4,95 @@ type Stubs struct {
...
 }
 
 // Up Run the migrations.
-func (r *DummyMigration) Up() error {
-	if !facades.Schema().HasTable("DummyTable") {
-		return facades.Schema().Create("DummyTable", func(table schema.Blueprint) {
+func (r *{{.StructName}}) Up() error {
+	if !facades.Schema().HasTable("{{.Table}}") {
+		return facades.Schema().Create("{{.Table}}", func(table schema.Blueprint) {
+			{{- if .SchemaFields}}
```
hwbrzzl, line 47: Image

hwbrzzl (commented):
Thanks, amazing PR. I successfully generated the migration. Left a few questions.

database/migration/model_mapper.go
krishankumar01: No, I also considered moving it to `FieldMetadata`, but the issue lies in how `GORM` tags are defined. In `JSON`, tag values are usually separated by `commas (,)`, whereas `GORM` uses key-value style tags separated by `semicolons (;)`. So, the logic differs, and since `;` isn’t a standard separator, it’s required here.

```go
	FullName      string    `gorm:"column:user_name;not null" json:"name,omitempty"`
```

database/migration/stubs.go
krishankumar01: Ohh, this was intentionally added by me, just a habit from `Java` to add a space before the actual logic, but I’ll remove it.

database/migration/stubs.go
krishankumar01: Hmm, good catch, I missed that. Forgot to update the file name; will fix it.

database/migration/migrator.go
krishankumar01: Should we fail here in case of an error when the user has defined a model, or should we generate an empty migration file instead? Ideally, we should fail, since if the user has added the `-m` flag, the expectation is that they intend to include a model.

support/structmeta/structmeta_test.go
krishankumar01: Yeah, it’s code I wrote back in May, it’s been a while, haha. Will correct it

database/migration/model_mapper.go
krishankumar01: Hmm, I also believe this logic is flawed, or maybe it can be optimized a bit. I’ll check if there’s a more dynamic way to handle it.

database/migration/migrator.go
hwbrzzl: Yes, should fail and print a warning message.

database/migration/model_mapper.go
hwbrzzl: We may use `FieldMetadata` to get the tags first, then use the gorm role to handle the tags deeply.

krishankumar01: Yes, that's how it's currently handled as well, we retrieve the `GORM` tag from `FieldMetaData`. See below.

Screenshot 2025-10-27 at 6 33 43 PM

database/migration/stubs.go
krishankumar01: @bowen Looks like the generated migration name is expected, this matches the behavior of older framework versions since we don’t manipulate the migration name. Regarding the column name, the logic for selecting the stub is quite strict: it expects one of the following regex patterns, but you entered `CreateAgentTable`, which doesn’t match either. This behavior is consistent with older versions.
```go
var CreatePatterns = []string{
	`^create_(\w+)_table$`,
	`^create_(\w+)$`,
}
```
Example (Go 1.16.x):
```go
go run . artisan make:migration HelloWorld
```
Screenshot 2025-10-27 at 10 08 32 PM

database/migration/model_mapper.go
krishankumar01: removed this completely, instead of these alias method, now model_mapper will generate explicit methods

database/schema/schema.go
hwbrzzl: I mean using the actual model (app/models/*.go) directly. User doesn't need to transform app/models/*.go to contractsschema.Model.

krishankumar01: Changed it to accept db.Model directly (updated PR description)

database/migration/model_mapper.go
krishankumar01: removed this since now I am using schema.Parse, which handles all of these

database/migration/stubs.go
krishankumar01: Fixed

krishankumar01:
@hwbrzzl It supports almost all features that GORM migration provides using tags. However, some index-related methods like sort, comment, and similar ones aren’t supported by the framework yet. We can generate them once IndexDefinition adds support for them.

database/console/migration/migrate_make_command_test.go
```diff
@@ -192,7 +197,8 @@ func Boot() {
 
 		mockContext.EXPECT().Argument(0).Return("create_comments_table").Once()
 		mockContext.EXPECT().OptionBool("force").Return(false).Once()
-		mockMigrator.EXPECT().Create("create_comments_table").Return("20240915060149_create_comments_table", nil).Once()
+		mockContext.EXPECT().Option("model").Return("").Once()
```
hwbrzzl, line 224: Could you add a case for non-empty model?

database/migration/stubs.go
hwbrzzl: Oh yes, sorry for the mistake.

database/migration/migrator.go
```diff
@@ -29,16 +34,58 @@ func NewMigrator(artisan console.Artisan, schema contractsschema.Schema, table s
...
+
+	tmpl, err := template.New("migration").Parse(stub)
+	if err != nil {
+		return "", errors.TemplateFailedToParse.Args(err.Error())
+	}
+
+	var buf bytes.Buffer
+	if err := tmpl.Execute(&buf, templateData); err != nil {
+		return "", err
+	}
```
hwbrzzl: The logic is duplicated with https://github.com/goravel/framework/pull/1049/files#diff-0d571b3b6b4d193b771529c2095794620681e7a09020d9e8c14d64c77b39c042R35-R60

database/migration/creator.go
```diff
@@ -31,13 +32,34 @@ func (r *Creator) GetStub(table string, create bool) string {
 	return Stubs{}.Update()
 }
 
+type StubData struct {
+	Package      string
+	StructName   string
+	Signature    string
+	Table        string
+	SchemaFields []string
```
hwbrzzl, line 41: The field is not used.

database/migration/creator.go
```diff
@@ -31,13 +32,34 @@ func (r *Creator) GetStub(table string, create bool) string {
...
+}
+
 // PopulateStub Populate the place-holders in the migration stub.
 func (r *Creator) PopulateStub(stub, signature, table string) string {
```
hwbrzzl: Could you optimize the test cases of this function as well?

database/migration/model_mapper.go
```diff
@@ -0,0 +1,508 @@
...
+	methodName      = "Name"
+	methodAlgorithm = "Algorithm"
+)
+
+// Index class constants matching GORM's index classification.
+const (
+	classUnique   = "UNIQUE"
+	classPrimary  = "PRIMARY"
+	classFullText = "FULLTEXT"
+)
```
hwbrzzl: How about adding them to the database/schema/blueprint.go? To avoid forgetting when adding a new one in the future. And please sort them according to the first letter in different group.

hwbrzzl: How about moving them to database/schema/blueprint.go? To avoid forgetting when adding new functions in the future? And please sort them in the group based on the first letter.

database/migration/model_mapper.go
```diff
@@ -0,0 +1,508 @@
...
+
+func shouldSkipField(field *schema.Field) bool {
+	// Skip ignored or embedded fields
+	if field.IgnoreMigration || field.EmbeddedSchema != nil || field.DBName == "" {
+		return true
+	}
+
+	// Skip relation fields
+	rels := &field.Schema.Relationships
+	rels.Mux.RLock()
```
hwbrzzl: Is it necessary to lock it?

database/migration/model_mapper.go
```diff
@@ -0,0 +1,508 @@
...
+	_, isRel := rels.Relations[field.Name]
+	rels.Mux.RUnlock()
+	if isRel {
+		return true
+	}
+
+	// Skip foreign key fields (belong to a relation)
+	for _, rel := range field.Schema.Relationships.Relations {
```
hwbrzzl: Why is the foreign key canceled? For example?

database/migration/model_mapper.go
```diff
@@ -0,0 +1,508 @@
...
+		switch idx.Class {
+		case classUnique:
+			method = methodUnique
+		case classFullText:
+			method = methodFullText
+		case classPrimary:
+			method = methodPrimary
+		}
+
+		add(idx.Class+":"+strings.Join(cols, ","), formatIndex(method, cols, idx))
```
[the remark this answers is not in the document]
krishankumar01: Why would I have column name a,b?

database/migration/model_mapper_test.go
```diff
@@ -0,0 +1,216 @@
...
+		wantErr     error
+		wantAnyErr  bool
+	}{
+		{
+			name:      "ORM Model (embedded)",
+			model:     &User{},
+			wantTable: "users",
+			wantColumns: []string{
+				`table.TimestampTz("created_at").Nullable()`,
+				`table.TimestampTz("updated_at").Nullable()`,
```
hwbrzzl, line 80: Can they be merged to `TimestampsTz`?

database/migration/model_mapper_test.go
```diff
@@ -0,0 +1,216 @@
...
+				`table.BigIncrements("id")`,
```
hwbrzzl, line 81: Can it be `ID`?

database/migration/model_mapper_test.go
```diff
@@ -0,0 +1,216 @@
...
+				`table.String("name").Nullable()`,
```
hwbrzzl: The `Name` is not a pointer, so I think `Nullable` makes no sense.

database/migration/model_mapper.go
hwbrzzl, line 1: Could you add more test cases for the sub-functions in this file?

hwbrzzl (commented):
Great PR 👍

database/migration/model_mapper_test.go
krishankumar01: I think it’s fine since it’s generated by code. Keeping it generic is better; otherwise, I’d need to add unnecessary checks for wrapper methods like ID() and SoftDeletes() etc without any benefit. That’s why I prefer a simpler internal implementation, in the end, it does the same thing anyway, and it makes handling easier for me. The same reasoning applies to all similar comments above.

database/migration/model_mapper.go
krishankumar01: Yes, since Relationship has an internal lock, I thought it’s better to respect it.

database/migration/model_mapper.go
krishankumar01: Good catch, It shouldn't remove foreign key field (we may not create relation in our migration directly).

process/pool_windows_test.go
```diff
@@ -257,6 +257,7 @@ func TestPool_Concurrency_Windows(t *testing.T) {
 }
 
 func TestPool_OnOutput_Windows(t *testing.T) {
+	// TODO: fix sync issue when writing concurrently
```
hwbrzzl, line 260: Will you create another PR for this?

database/migration/creator.go
```diff
@@ -31,13 +32,32 @@ func (r *Creator) GetStub(table string, create bool) string {
...
+		return stub
+	}
+
+	formatted, err := format.Source(buf.Bytes())
+	if err != nil {
+		return stub
+	}
 
-	return stub
+	return string(formatted)
```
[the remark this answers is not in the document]
hwbrzzl: +1

process/pool_windows_test.go
krishankumar01: Yeah, I will revisit it in a separate PR. This issue exists in the test cases, not in the actual implementation, but I’ll check if we need any improvements in the implementation as well.

hwbrzzl:
Found an issue when running locally:

image

krishankumar01:
> Found an issue when running locally:
> image

Fixed! For gorm.DeletedAt, the field is now correctly marked as .Nullable().
Regarding the serializer:json tag: it won't automatically make the column a JSON type. GORM uses this tag only for Go-side serialization/deserialization, but stores the data as TEXT or VARCHAR in the database. If you want it to map to Json than you can use gorm:"migration:Json" tag.

database/migration/stubs.go
```diff
@@ -4,85 +4,96 @@ type Stubs struct {
...
-	})
+func (r *{{.StructName}}) Up() error {
+	return facades.Schema().Table("{{.Table}}", func(table schema.Blueprint) {
+		{{- if .SchemaFields}}
+{{- range .SchemaFields}}
+		{{.}}
+		{{- end}}
+{{else}}
+
+{{end}}	})
```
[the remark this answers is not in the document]
krishankumar01: it will be automatically formatted by PopulateStub

PR #1050, opened 2025-05-27
chore: improve Cursor error delivery and defer logic

almas-x:
### What’s Changed

- Refactored `Query.Cursor()` to simplify the return signature by removing the `(chan, error)` tuple.
- Now, errors during row iteration are sent through the channel by wrapping them in a `Row` struct with the `err` field set.
- Deferred cleanup of resources (channel close and `Rows.Close()`) is handled reliably even in case of early errors.
- Prevents potential goroutine leaks and ensures consumers always receive a closed channel.

### Why

Previously, errors in goroutine setup (e.g., `query.instance.Rows()`) were only returned via the outer error return, which could be misleading since the channel would never yield results. This refactor ensures a more Go-idiomatic pattern: the channel always closes, and all consumers can handle errors uniformly through the same channel.

### Potential Impact

Downstream consumers of `Cursor()` should now check `Row.err` to determine if an error occurred, instead of relying on the function's return error.

### Example Usage (Updated)

```go
for row := range query.Cursor() {
	if row.Err() != nil {
		log.Println("cursor error:", row.Err())
		break
	}
	// Process row
}
```

## ✅ Checks


contracts/database/db/db.go
```diff
@@ -213,5 +213,6 @@ type ToSql interface {
 }
 
 type Row interface {
+	Err() error
```
hwbrzzl, line 216: `Error` or `Err`?

database/gorm/query.go
```diff
@@ -149,33 +149,42 @@ func (r *Query) Create(value any) error {
...
+			if err != nil {
+				cursorChan <- &Row{query: r, err: err}
+			}
+			close(cursorChan)
+		}()
+
+		if rows, err = query.instance.Rows(); err != nil {
 			return
 		}
 		defer rows.Close()
```
hwbrzzl, line 173: How about this line? Should it be called before returning error?

database/gorm/row.go
```diff
@@ -10,11 +10,20 @@ import (
...
 	row   map[string]any
 }
 
+func (r *Row) Err() error {
+	return r.err
+}
+
 func (r *Row) Scan(value any) error {
+	if r.err != nil {
+		return r.err
```
hwbrzzl, line 24: The error is returned here, is the `Err` method necessary?

hwbrzzl (commented):
LGTM, can we add some test cases for the error flow?

database/gorm/row.go
almas-x: The duplication is intentional. `Row.Err()` is there for callers who like to check the error first, while` Row.Scan()` returns the same error for those who prefer to scan directly and handle any failure in one place.

In other words, it’s a convenience “double-check” that matches both usage styles, similar to how the standard GORM API works.

almas-x:
> LGTM, can we add some test cases for the error flow?

I’ll add some basic error flow tests. Other scenarios are hard to simulate without mocking or a more complex setup.

almas-x:
> Great PR 👍 Could you fix the gorm query for the v1.15.x branch as well? They are a bit different, only need to fix the memory leak, don't make a breaking change for the `Cursor` function.

Okay

PR #1061, opened 2025-06-05
fix: [#699] database timezone

hwbrzzl:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/699

Currently, for `Timestamps` column of Postgres and Sqlserver, `2025-06-05 14:46:29 +08:00` will be saved with `2025-06-05 14:46:29`, but when reading, ORM considers that the timezone is UTC, the timezone metadata is missed.

## ✅ Checks


database/migration/stubs.go
```diff
@@ -87,131 +87,3 @@ func (r *DummyMigration) Down() error {
...
-}
-
-type PostgresStubs struct {
-}
-
-// CreateUp Create up migration content.
-func (receiver PostgresStubs) CreateUp() string {
-	return `CREATE TABLE DummyTable (
-  id SERIAL PRIMARY KEY NOT NULL,
-  created_at timestamp NOT NULL,
```
hwbrzzl, line 132: Yes, but it's removed here.

PR #1091, opened 2025-06-24
feat: [#717] support displaying handler names in artisan route:list command

almas-x:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/717

image

## ✅ Checks

### DIFF
--- a/log/utils_test.go
+++ b/log/utils_test.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"testing"
 
-	"github.com/gookit/goutil/testutil/assert"
+	"github.com/stretchr/testify/assert"
 )
 
 type testContextKey any
--- a/route/console/route_list_command.go
+++ b/route/console/route_list_command.go
@@ -90,7 +90,7 @@ func (r *RouteListCommand) Handle(ctx console.Context) error {
 	}
 
 	for _, item := range routes {
-		ctx.TwoColumnDetail(fmt.Sprintf("%s %s", r.formatMethod(item.Method), r.formatPath(item.Path)), item.Name)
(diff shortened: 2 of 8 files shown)


log/utils_test.go
```diff
@@ -4,7 +4,7 @@ import (
 	"context"
 	"testing"
 
-	"github.com/gookit/goutil/testutil/assert"
+	"github.com/stretchr/testify/assert"
```
hwbrzzl, line 7: Good catch

route/console/route_list_command.go
```diff
@@ -147,6 +147,18 @@ func (r *RouteListCommand) filterRoutes(ctx console.Context, routes []route.Info
...
+func (r *RouteListCommand) formateNameHandler(name, handler string) string {
+	if len(name) == 0 && len(handler) == 0 {
+		return ""
+	}
+
+	if len(name) > 0 && len(handler) > 0 {
+		name = name + " › "
+	}
+
+	return fmt.Sprintf("<fg=7472A3>%s%s</>", name, strings.TrimSuffix(handler, "-fm"))
```
hwbrzzl, line 159: What's the `-fm`?

almas-x: When using method values like `UserController.Show` (i.e., a bound method), Go generates a wrapper function internally to capture the receiver. The function name ends with -fm to indicate it’s a function method wrapper.

Example:

```go
type UserController struct{}
func (t *UserController) Show() {}

u := &UserController{}
fmt.Println(runtime.FuncForPC(reflect.ValueOf(u.Show).Pointer()).Name())

```

Output:

```bash
main.(*UserController).Show-fm
```

PR #1094, opened 2025-06-26
feat: [#682] Automatically reconnect to database when connection is lost, including on startup or runtime interruptions

hwbrzzl:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/682

Gorm can reconnect automatically, it's hard to add a test case for this, tested it locally:

image

## ✅ Checks

### DIFF
--- a/database/db/logger.go
+++ b/database/db/logger.go
@@ -9,6 +9,7 @@ import (
 	"github.com/goravel/framework/contracts/config"
 	"github.com/goravel/framework/contracts/database/logger"
 	"github.com/goravel/framework/contracts/log"
+	"github.com/goravel/framework/errors"
 	"github.com/goravel/framework/support/carbon"
 	"github.com/goravel/framework/support/str"
 )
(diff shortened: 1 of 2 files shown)


database/db/logger.go
```diff
@@ -94,7 +95,7 @@ func (r *Logger) Trace(ctx context.Context, begin *carbon.Carbon, sql string, ro
 	addQueryLogToContext(ctx, sql, elapsed)
 
 	switch {
-	case err != nil && r.level >= logger.Error:
+	case err != nil && r.level >= logger.Error && !errors.Is(err, gormlogger.ErrRecordNotFound):
```
hwbrzzl, line 98: Skip such logs: 

image

database/driver/gorm.go
```diff
@@ -38,6 +46,16 @@ func BuildGorm(config config.Config, logger logger.Interface, pool database.Pool
...
 		return nil, err
 	}
+	if pinger, ok := instance.ConnPool.(interface{ Ping() error }); ok {
+		if err = pinger.Ping(); err != nil {
+			printedPingWarningLock.Lock()
+			defer printedPingWarningLock.Unlock()
+			if !printedPingWarning {
+				color.Warningln(err.Error())
+				printedPingWarning = true
+			}
```
almas-x: how about replacing this with sync.Once? it fits the 'print only once' use case more cleanly

```go
var  pingWarning sync.Once

if pinger, ok := instance.ConnPool.(interface{ Ping() error }); ok {
    if err := pinger.Ping(); err != nil {
        pingWarning.Do(func() {
            color.Warningln(err.Error())
        })
    }
}

```

hwbrzzl: Good point

almas-x (approved):
LGTM

PR #1096, opened 2025-06-27
feat: [#702] apply fieldalignment & modernize fixes

almas-x:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/702

This PR includes automated refactors applied via `fieldalignment` and `modernize` to improve code quality and consistency:

- **fieldalignment**: optimized struct field order for better memory layout  
- **modernize**:
  - `efaceany`: replaced `interface{}` with Go 1.18+ `any`
  - `minmax`: simplified `if-else` assignments to built-in `min/max`
  - `rangeint`: converted classic `for i := 0; i < n; i++` loops to `for i := range n`
  - `fmtappendf`, `mapsloop`, `slicescontains`: applied equivalent modernizations

## ✅ Checks

### DIFF
--- a/auth/access/response.go
+++ b/auth/access/response.go
@@ -11,8 +11,8 @@ func NewDenyResponse(message string) access.Response {
 }
 
 type ResponseImpl struct {
-	allowed bool
 	message string
+	allowed bool
 }
 
 func (r *ResponseImpl) Allowed() bool {
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -20,8 +20,8 @@ type Auth struct {
 	contractsauth.GuardDriver
 	config           config.Config
 	ctx              http.Context
-	defaultGuardName string
 	log              log.Log
+	defaultGuardName string
 }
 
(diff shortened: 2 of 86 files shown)


hwbrzzl (commented):
Great PR 👍 Is there a script or github action to format files automatically?

almas-x:
> Great PR 👍 Is there a script or github action to format files automatically?

It’s not easy to create a fully automated script for this, unfortunately.

The `fieldalignment` tool doesn’t exclude _test.go files by default, so some manual cleanup is required. And for `modernize`, the issues it detects and fixes can vary depending on the Go version, which means we also need to review and filter changes manually.

Because of that, it’s more practical to run these optimizations periodically—such as before major releases—rather than on every commit.

auth/access/response.go
```diff
@@ -11,8 +11,8 @@ func NewDenyResponse(message string) access.Response {
 }
 
 type ResponseImpl struct {
-	allowed bool
 	message string
+	allowed bool
```
hwbrzzl, line 15: Could you add the sort rule to the PR description?

PR #1109, opened 2025-07-02
chore: optimize testing cookie

hwbrzzl:
## 📑 Description

Previously, `request.WithCookies` needs `map[string]string`, but `response.Cookies` returns `[]*http.Cookie`. They should be the same.

## ✅ Checks


testing/http/test_request.go
```diff
@@ -105,13 +105,13 @@ func (r *TestRequest) WithoutHeader(key string) contractshttp.Request {
 	return r
 }
```
almas-x, line 107: How about adding sugar functions `func Cookie(key, value string) *http.Cookie` and `func Cookies(cookies map[string]string) []*http.Cookie` for easier cookie creation?

hwbrzzl: The way doesn't support adding `MaxAge`, `Expires`, etc. variables into the cookie, it's a bit limited.

hwbrzzl: And `response.Cookites` returns `[]*http.Cookie`, it's a bit complicated to convert `[]*http.Cookie` to `map[string]string`.

almas-x: I just meant to add two package-level helper functions to simplify cookie creation.
The existing methods still receive `*http.Cookie` and `[]*http.Cookie` as usual.

This also helps avoid having both
```go
import (
     nethttp "net/http"

    "github.com/goravel/framework/testing/http"
)
```

hwbrzzl: Good idea, added.

PR #1123, opened 2025-07-09
feat: queue support attempts

devhaozi:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/696

This pull request introduces a retry mechanism for job processing in the queue system, along with enhancements to the `Worker` and its associated tests. The most significant changes include the addition of a `JobWithShouldRetry` interface, updates to the `Worker` to support configurable retry attempts, and extensive test coverage for the retry functionality.

### Retry Mechanism Enhancements:
* Added a new interface, `JobWithShouldRetry`, to allow jobs to define custom retry logic via the `ShouldRetry` method. (`contracts/queue/job.go`, [contracts/queue/job.goR47-R51](diffhunk://#diff-be2264bb9d3b233f3eed23b901921537ae5236354adac2d6578de8d5616c6b4fR47-R51))
* Updated the `Worker` to include a `tries` field and logic to handle retries based on the `JobWithShouldRetry` interface or a default retry policy. (`queue/worker.go`, [[1]](diffhunk://#diff-956d03d9916ff2d69ac24682eadffb63d96075824322b921385a4d83223da2d6R37-R45) [[2]](diffhunk://#diff-956d03d9916ff2d69ac24682eadffb63d96075824322b921385a4d83223da2d6L126-R151) [[3]](diffhunk://#diff-956d03d9916ff2d69ac24682eadffb63d96075824322b921385a4d83223da2d6L145-L148)

### Worker Updates:
* Modified the `NewWorker` function to accept a `tries` parameter, enabling configuration of the maximum retry attempts for jobs. (`queue/worker.go`, [queue/worker.goR63](diffhunk://#diff-956d03d9916ff2d69ac24682eadffb63d96075824322b921385a4d83223da2d6R63))
* Updated the `Application` to pass the `tries` parameter when creating a new `Worker`. (`queue/application.go`, [[1]](diffhunk://#diff-34aafc02eb2c7a5ca85c1a4a79bb1f8e58711850cc764d4fc58e96401b397bb3L68-R68) [[2]](diffhunk://#diff-34aafc02eb2c7a5ca85c1a4a79bb1f8e58711850cc764d4fc58e96401b397bb3L84-R84)

### Testing Improvements:
* Added mock implementations for the `JobWithShouldRetry` interface to facilitate unit testing. (`mocks/queue/JobWithShouldRetry.go`, [mocks/queue/JobWithShouldRetry.goR1-R93](diffhunk://#diff-f295d7a3d174d1f8d78ca5da1fad775a6b52c3ce2f942905749a518c5337c117R1-R93))
* Enhanced the `WorkerTestSuite` with new test cases to validate retry behavior, including scenarios for successful retries, failed retries, and job insertion into the failed jobs table. (`queue/worker_test.go`, [[1]](diffhunk://#diff-5210799b58226ee077161e9efd34924773948d2bb114f5dd9b95d0013ebe97abR489-R614) [[2]](diffhunk://#diff-5210799b58226ee077161e9efd34924773948d2bb114f5dd9b95d0013ebe97abR631-R652)

### Dependency Updates:
* Updated the `google.golang.org/genproto/googleapis/rpc` dependency to the latest version. (`tests/go.mod`, [tests/go.modL69-R69](diffhunk://#diff-8218fcbcda93ea8aeb0e1cd1bd9603a84d41f224ba6f88fa6ba44996d123b266L69-R69))

## ✅ Checks


almas-x:
Amazing PR 👍

Currently, the retry interval is calculated as `time.Sleep(time.Second * time.Duration(tries))`, but ideally, the worker could support a configurable backoff strategy. It would also be helpful to support some common backoff algorithms out of the box, such as:

- Fixed delay
- Linear backoff
- Exponential backoff
- Exponential backoff with jitter

Even better, jobs could provide a custom retry policy—for example, implementing an interface like `interface{ ShouldRetry(err error, attempt int) (retry bool, wait time.Duration) }` that receives the current error and retry count, and then determines whether to retry and how long to wait before the next attempt.

If a job implements the retry policy interface, the decision to retry and the retry interval would be determined by the policy itself, making the worker’s max tries ineffective for that job. If the job does not implement a custom retry policy, the worker would fall back to its own attempts/tries and backoff configuration for retries.

This approach would give users more flexibility to control retry behavior for different jobs and use cases.

queue/worker.go
```diff
@@ -140,6 +145,12 @@ func (r *Worker) call(task queue.Task) error {
 
 		r.printFailedLog(task, duration)
 
+		if tries < r.tries || r.tries == 0 { // If tries is 0, it means unlimited retries
```
almas-x: Shouldn't `r.failedJobChan <- models.FailedJob` only be executed after all retry attempts are exhausted and the job still fails?

queue/worker.go
```diff
@@ -113,6 +115,9 @@ func (r *Worker) Shutdown() error {
 }
 
 func (r *Worker) call(task queue.Task) error {
+	tries := 1
+reCall:
```
hwbrzzl: Recommend `for` install of `goto`.

queue/worker.go
hwbrzzl: The default of `tries` is 0, so I think the meaning of `0` should be: not retry.

hwbrzzl:
Please add the `Review Ready` if it's ready.

queue/worker.go
devhaozi: > The default of `tries` is 0, so I think the meaning of `0` should be: not retry.

But in Laravel 0 mean retry indefinitely.

https://laravel.com/docs/12.x/queues#max-attempts

image

hwbrzzl: Good catch 👍 For the old project, the tries default value will be 0, and jobs will be retried indefinitely, it's unexpected. I suggest we don't support this feature for now. Users can set a maximum number if needed. cc @almas-x Any thoughts, please?

almas-x: > Good catch 👍 For the old project, the tries default value will be 0, and jobs will be retried indefinitely, it's unexpected. I suggest we don't support this feature for now. Users can set a maximum number if needed. cc @almas-x Any thoughts, please?

I agree, implicit unlimited retries are not ideal.

contracts/queue/job.go
```diff
@@ -44,3 +44,8 @@ type ChainJob struct {
 	Job   Job       `json:"job"`
 	Args  []Arg     `json:"args"`
 }
+
+type JobWithShouldRetry interface {
```
hwbrzzl, line 48: Good implementation 👍

hwbrzzl (approved):
Nice 👍

PR #1128, opened 2025-07-16
feat: add pluralizer package

kkumar-gcc:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/731

## What’s in this PR
- Brand‑new `pluralizer` package for Go
- Out‑of‑the‑box English inflector (singular ↔ plural)
- Easy to swap languages or add your own
- Register custom rules without touching core code

## Quick Start
```go
import (
     "github.com/goravel/framework/support/pluralizer"
     "github.com/goravel/framework/support/str"
)

// default (English)
fmt.Println(pluralizer.Plural("goose"))    // "geese"
fmt.Println(pluralizer.Singular("heroes")) // "hero"

// fluent strings also provide chaining for these methods 
fmt.Println(str.Of("goose").Plural())    // "geese"
fmt.Println(str.Of("heroes").Singular()) // "hero"
```

## Swap the Global Inflector
Do this at app startup (not inside goroutines)
```go
if err := pluralizer.UseLanguage("english"); err != nil {
    log.Fatal(err)
}

// see what’s active
fmt.Println(pluralizer.GetLanguage().Name()) // "english"
```
## Custom Rules
```go
// irregular: mouse → mice
_ = pluralizer.RegisterIrregular("english",
    pluralizer.Substitution{Singular: "mouse", Plural: "mice"},
)

// mark words uninflected
_ = pluralizer.RegisterUninflected("english", "sheep", "species")

// only plural uninflected
_ = pluralizer.RegisterPluralUninflected("english", "media")

// only singular uninflected
_ = pluralizer.RegisterSingularUninflected("english", "data")
```

## Add Your Own Inflector
Implement this interface:
```go
type Language interface {
    Name() string
    SingularRuleset() Ruleset
    PluralRuleset() Ruleset
}
```
Register & use it in `init()`:
```go
pluralizer.RegisterLanguage("fantasy", &FantasyLang{})
_ = pluralizer.UseLanguage("fantasy")
```
> *Note*: Changing or registering inflectors isn’t goroutine‑safe; do it at startup.
Found a missing rule? Please open an issue!

## ✅ Checks


hwbrzzl:
Thanks, will check this after v1.16 is released.

hwbrzzl:
And please add the `Review Ready` tag when it is ready.

krishankumar01:
@hwbrzzl , this package is needed to resolve the table name from the struct name. Currently, it only supports English, but it can be easily extended to support other languages.

Ref: https://laravel.com/docs/12.x/localization#pluralization-language

support/pluralizer/english.go
hwbrzzl, line 1: Where does the rule come from, please?

krishankumar01: I referred to multiple sources:
- [Doctrine Inflector (used internally by Laravel)](https://github.com/doctrine/inflector/tree/2.0.x/src/Rules/English)
- [Symfony EnglishInflector](https://github.com/symfony/symfony/blob/7.3/src/Symfony/Component/String/Inflector/EnglishInflector.php)

support/pluralizer/pluralizer.go
```diff
@@ -0,0 +1,48 @@
+package pluralizer
+
+import "sync"
+
+type Pluralizer interface {
+	Plural(word string) string
+	Singular(word string) string
+}
```
hwbrzzl: Please move this to `contracts/support`

support/pluralizer/pluralizer.go
```diff
@@ -0,0 +1,48 @@
...
+
+type Language string
+
+const (
+	English Language = "en"
```
hwbrzzl: How about using `string` directly? It seems to be unnecessary to define a new type for this case.

support/pluralizer/pluralizer.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+)
+
+func init() {
+	once.Do(func() {
+		instance = newEnglishInflector()
+	})
+}
+
+func New() Pluralizer {
```
hwbrzzl: It seems not to be used.

support/pluralizer/pluralizer.go
```diff
@@ -0,0 +1,48 @@
...
+
+func NewForLanguage(lang Language) Pluralizer {
+	if constructor, ok := factory[lang]; ok {
+		return constructor()
+	}
+	return newEnglishInflector()
+}
+
+func Plural(word string) string {
+	return instance.Plural(word)
```
hwbrzzl: There is no way to add and switch language, add a `UseLanguage` as well?

support/pluralizer/rules.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+type RuleSet struct {
+	uncountable       map[string]bool
+	uncountableRegex  []*regexp.Regexp
+	irregularPlural   map[string]string
+	irregularSingular map[string]string
+	plural            []Rule
+	singular          []Rule
+}
```
hwbrzzl: Can it be used by user when implementing a new language? If so, the parameters should be public, right?

support/pluralizer/inflector.go
```diff
@@ -0,0 +1,83 @@
+package pluralizer
+
+import (
+	"strings"
+	"unicode"
+)
+
+type inflector struct {
```
hwbrzzl: Can it be used by user when implementing a new language?

hwbrzzl (commented):
Thanks, great PR 👍

support/pluralizer/pluralizer.go
krishankumar01: Yeah, changed it to string

support/pluralizer/pluralizer.go
krishankumar01: I've removed it and cleaned up the `init` file. Please have a look.

support/pluralizer/pluralizer.go
krishankumar01: Yeah, I have introduced a new method for this

support/pluralizer/inflector.go
krishankumar01: Yeah, I’ve exposed two new methods for this:
- `UseLanguage`: Switches the current global language if it already exists
- `RegisterLanguage`: Allows you to register a new language 

** recommended to use during application boot time

hwbrzzl (commented):
Em, the logic is a bit complicated now. If so, I don't suggest releasing it to v1.16. A deep review is required. Hence, we can delay [this PR](https://github.com/UrbanCompass/urbancompass/pull/127896/) as well, given that the another related PR (table -> model) spent many days to be merged. 

We don't need to merge them hurriedly, it's fine to release them in v1.17. As we know, errors always appear in this situation.

krishankumar01:
Yeah, pluralization is a bit tricky, so the logic ended up being a bit more complex. The earlier version wasn’t very flexible, so I changed it to make it more future-proof. No worries, take your time, we can include it in V1.17. The second PR depends on this, so we can delay that too.

krishankumar01:
@hwbrzzl Converting this PR to a draft for now, need to do more research on English word dictionaries. I'm exploring different sources to improve pattern accuracy.

krishankumar01:
This package may not always get pluralization and singularization 100% right, since English (like many languages) has lots of exceptions. But the goal is to keep improving it over time as we come across edge cases. Think of it as a work in progress that gets better the more we use it.

hwbrzzl:
> This package may not always get pluralization and singularization 100% right, since English (like many languages) has lots of exceptions. But the goal is to keep improving it over time as we come across edge cases. Think of it as a work in progress that gets better the more we use it.

Yes, we can provide a way for users to extend the special cases.

krishankumar01:
> > This package may not always get pluralization and singularization 100% right, since English (like many languages) has lots of exceptions. But the goal is to keep improving it over time as we come across edge cases. Think of it as a work in progress that gets better the more we use it.
> 
> Yes, we can provide a way for users to extend the special cases.

I've added a few methods to allow users to extend irregular and uninflected cases.

support/pluralizer/init.go
```diff
@@ -0,0 +1,127 @@
...
+		"english": inflector.New(english.New()),
+	}
+)
+
+func init() {
+	currentLanguage = "english"
+	instance = inflectorFactory[currentLanguage]
+}
+
+func UseLanguage(lang string) bool {
```
hwbrzzl:
```suggestion
func UseLanguage(lang string) error {
```

support/pluralizer/init.go
```diff
@@ -0,0 +1,127 @@
...
+		return true
+	}
+	return false
+}
+
+func GetLanguage() string {
+	return currentLanguage
+}
+
+func RegisterLanguage(language pluralizer.Language) bool {
```
hwbrzzl:
```suggestion
func RegisterLanguage(language pluralizer.Language) error {
```

support/pluralizer/init.go
```diff
@@ -0,0 +1,127 @@
...
+func UseLanguage(lang string) bool {
+	if factory, exists := inflectorFactory[lang]; exists {
+		currentLanguage = lang
+		instance = factory
+		return true
+	}
+	return false
+}
+
+func GetLanguage() string {
```
hwbrzzl:
```suggestion
func GetLanguage() pluralizer.Language {
```

support/pluralizer/init.go
```diff
@@ -0,0 +1,127 @@
...
+
+	language, factory, exists := getLanguageInstance(lang)
+	if !exists {
+		return false
+	}
+
+	language.SingularRuleset().AddUninflected(words...)
+	factory.SetLanguage(language)
+	return true
+}
```
hwbrzzl: How about add these functions to `type Language interface`? When users want to extend:

```go
language := pluralizer.GetLanguage()
language.SingularRuleset.Add...
language.PluralRuleset.Add...
language.Add...
```

krishankumar01:
> And I wonder if there is an existing package that can implement such features as well?

Yeah, there are some existing packages, but most of them are limited to English. If we want to support additional languages in the future, they won’t be sufficient. So it's better to implement it ourselves.

Examples:
- https://github.com/jinzhu/inflection
- https://github.com/gobuffalo/flect

support/pluralizer/init.go
krishankumar01: No, if we add these methods to `Language` interface, the user would have to implement them for each new custom language, which is unnecessary since the logic will be the same for all languages. `Language` interface should be minimal.

errors/list.go
```diff
@@ -136,6 +136,11 @@ var (
 	PackageRegistrationDuplicate = New("'%s' had been registered")
 	PackageRegistrationNotFound  = New("'%s' not found, cannot insert before it")
 
+	PluralizerLanguageNotFound     = New("language %s not found").SetModule(ModulePluralizer)
+	PluralizerEmptyLanguageName    = New("language name cannot be empty").SetModule(ModulePluralizer)
+	PluralizerNoSubstitutionsGiven = New("no substitutions provided").SetModule(ModulePluralizer)
+	PluralizerNoWordsGiven         = New("no words provided").SetModule(ModulePluralizer)
```
krishankumar01, line 142: Not sure if we should throw an error in these two cases.

support/pluralizer/init.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/goravel/framework/contracts/support/pluralizer"
+	"github.com/goravel/framework/errors"
+	"github.com/goravel/framework/support/pluralizer/english"
+	"github.com/goravel/framework/support/pluralizer/inflector"
+	"github.com/goravel/framework/support/pluralizer/rules"
+)
+
+var (
+	instance         pluralizer.Inflector
+	defaultLanguage  = "english"
```
hwbrzzl: We can add a string public constant for `english`.

errors/list.go
krishankumar01: @hwbrzzl ^^

hwbrzzl (approved):
Amazing PR, thanks!

hwbrzzl (approved):
LGTM

PR #1132, opened 2025-07-18
fix: Refactor str package methods for better Replace and case handling

kkumar-gcc:
## 📑 Description
- Fixed `ChopStart` and `ChopEnd` to match and remove exact strings instead of treating them as character sets (e.g., "hehe".chopStart("he") now returns "he").
- Escaped special characters in `Replace` when using case-insensitive mode to avoid incorrect regex behavior.
- Improved `Snake` case conversion to handle acronyms properly, e.g., `UserID` becomes `user_id` instead of `user_i_d`.

## ✅ Checks

### DIFF
--- a/support/str/str.go
+++ b/support/str/str.go
@@ -122,8 +122,8 @@ func (s *String) ChopEnd(needle string, more ...string) *String {
 	more = append([]string{needle}, more...)
 
 	for _, v := range more {
-		if s.EndsWith(v) {
-			s.value = strings.TrimSuffix(s.value, v)
+		if after, found := strings.CutSuffix(s.value, v); found {
+			s.value = after
 			break
 		}
 	}
@@ -135,8 +135,8 @@ func (s *String) ChopStart(needle string, more ...string) *String {
 	more = append([]string{needle}, more...)
 
 	for _, v := range more {
-		if s.StartsWith(v) {
-			s.value = strings.TrimPrefix(s.value, v)
+		if after, found := strings.CutPrefix(s.value, v); found {
+			s.value = after
 			break
 		}
 	}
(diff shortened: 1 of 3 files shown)


support/str/str.go
```diff
@@ -123,7 +123,7 @@ func (s *String) ChopEnd(needle string, more ...string) *String {
 
 	for _, v := range more {
 		if s.EndsWith(v) {
-			s.value = strings.TrimRight(s.value, v)
+			s.value = strings.TrimSuffix(s.value, v)
```
almas-x: How about using the built-in `strings.CutSuffix` function? For example:

before:
```go
if s.EndsWith(v) {
	s.value = strings.TrimSuffix(s.value, v)
	break
}

```

after:

```go
if value, found := strings.CutSuffix(s.value); found {
	s.value = value
	break
}
```

support/str/str.go
```diff
@@ -136,7 +136,7 @@ func (s *String) ChopStart(needle string, more ...string) *String {
 
 	for _, v := range more {
 		if s.StartsWith(v) {
-			s.value = strings.TrimLeft(s.value, v)
+			s.value = strings.TrimPrefix(s.value, v)
```
almas-x: Ditto

support/str/str.go
krishankumar01: Good suggestion 👍

support/str/str.go
```diff
@@ -122,8 +122,8 @@ func (s *String) ChopEnd(needle string, more ...string) *String {
 	more = append([]string{needle}, more...)
 
 	for _, v := range more {
-		if s.EndsWith(v) {
-			s.value = strings.TrimRight(s.value, v)
+		if after, found := strings.CutSuffix(s.value, v); found {
```
almas-x, line 125: 🪄 Oops, it's `before` not `after`! 🤪

krishankumar01: `after`: it’s the value after cutting the prefix/suffix 😄

almas-x: image

Both make sense, but wow, it’s super confusing.🤔

krishankumar01: Yeah, variable names depend on the context of their definition, so let’s go with ours i.e., the value after applying the operation.

hwbrzzl (approved):
Amazing, thanks!

PR #1133, opened 2025-07-19
refactor: events refactor (Listen, Dispatch)

ahmed3mar:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/730

This PR introduces a modern, flexible event handling system with `Listen()` and `Dispatch()` methods, replacing the old `Register()`/`Job()` pattern. The new system supports wildcard patterns, dynamic listener registration, and seamless queue integration.

#### 1. **Flexible Event Registration with `Listen()`**

The new `Listen()` method supports multiple event formats and listener types:

```go
// String events
app.Listen("user.created", func(event string, user User) error {
    return nil
})

// Multiple events at once
app.Listen([]string{"user.created", "user.updated"}, listener)

// Event interface types
app.Listen(&UserCreated{}, &UserCreatedListener{})

// Wildcard patterns
app.Listen("user.*", func(event string) error {
    // Handles user.created, user.updated, user.deleted, etc.
    return nil
})
```

**Comparison with old approach:**

```go
// ❌ Old way - Rigid, ceremony-heavy
app.Register(map[event.Event][]event.Listener{
    &UserCreated{}: {&UserCreatedListener{}},
    &UserUpdated{}: {&UserCreatedListener{}}, // Same listener, must repeat
})

// ✅ New way - Flexible, concise
app.Listen([]event.Event{&UserCreated{}, &UserUpdated{}}, &UserCreatedListener{})
// Or even better with wildcards:
app.Listen("user.*", &UserCreatedListener{})
```

#### 2. **Direct Event Dispatching with `Dispatch()`**

Fire events immediately and get listener responses:

```go
// Simple dispatch
responses := app.Dispatch("user.created")

// With payload
responses := app.Dispatch("order.placed", []event.Arg{
    {Value: order, Type: "Order"},
    {Value: customer, Type: "Customer"},
})

// Responses collected from all listeners
for _, response := range responses {
    log.Info("Listener returned:", response)
}
```

**Comparison with old approach:**

```go
// ❌ Old way - Indirect, requires Task creation
task := app.Job(event, args)
err := task.Dispatch() // No responses available

// ✅ New way - Direct with responses
responses := app.Dispatch(event, args) // Get all listener responses
```

#### 3. **Wildcard Pattern Support**

Match multiple events with a single listener registration:

```go
// Listen to all user events
app.Listen("user.*", func(eventName string, data any) error {
    log.Info("User event fired:", eventName)
    return nil
})

// Now all these events trigger the listener:
app.Dispatch("user.created")
app.Dispatch("user.updated")
app.Dispatch("user.deleted")
app.Dispatch("user.logged_in")
```

**Real-world example:**

```go
// Audit logging for all admin actions
app.Listen("admin.*", &AuditLogger{})

// Security monitoring for auth events
app.Listen("auth.*", &SecurityMonitor{})

// Metrics collection for all events
app.Listen("*", &MetricsCollector{})
```

#### 4. **Automatic Queue Integration**

Queue-aware listeners are automatically dispatched to the queue system:

```go
type UserCreatedListener struct{}

func (l *UserCreatedListener) Handle(event any, args ...any) error {
    // Process event
    return nil
}

func (l *UserCreatedListener) ShouldQueue() bool {
    return true // Automatically queued
}

func (l *UserCreatedListener) Signature() string {
    return "UserCreatedListener"
}

func (l *UserCreatedListener) Queue(args ...any) event.Queue {
    return event.Queue{
        Connection: "redis",
        Queue:      "events",
        Enable:     true,
    }
}

// Just listen - queueing happens automatically!
app.Listen("user.created", &UserCreatedListener{})
```

#### 5. **Dynamic Listener Arguments**

Listeners automatically receive matched arguments:

```go
// No arguments
app.Listen("app.started", func() error {
    log.Info("App started")
    return nil
})

// Event name only
app.Listen("user.created", func(eventName string) error {
    log.Info("Event:", eventName)
    return nil
})

// Event + typed payload
app.Listen("user.created", func(eventName string, user User) error {
    log.Info("User created:", user.Email)
    return nil
})

// Variadic for all arguments
app.Listen("order.placed", func(eventName string, args ...any) error {
    // Handle variable number of arguments
    return nil
})
```

### Migration Guide

The old `Register()` and `Job()` methods are now deprecated but still supported:

```go
// ❌ Old Pattern (Deprecated)
app.Register(map[event.Event][]event.Listener{
    &UserCreated{}: {&EmailListener{}, &NotificationListener{}},
    &OrderPlaced{}: {&InvoiceListener{}},
})

task := app.Job(&UserCreated{}, args)
task.Dispatch()

// ✅ New Pattern (Recommended)
app.Listen(&UserCreated{}, &EmailListener{}, &NotificationListener{})
app.Listen(&OrderPlaced{}, &InvoiceListener{})

responses := app.Dispatch(&UserCreated{}, args)
```

### Architecture Improvements

- **Thread-safe**: All operations use proper locking (RWMutex for registration, RLock for dispatch)
- **Cached wildcard matching**: Performance optimization for frequently dispatched events
- **Flexible listener types**: Functions, closures, structs - all supported
- **Type-safe**: Automatic argument type matching with reflection
- **Memory efficient**: Listener deduplication prevents memory leaks
- **High performance**: Interface assertions instead of reflection in hot paths

### Backward Compatibility

- ✅ Old `Register()` method still works
- ✅ Old `Job()` method still works
- ✅ Existing `Event` and `Listener` interfaces unchanged
- ✅ All existing tests pass
- ✅ No breaking changes

## ✅ Checks

### DIFF
--- a/event/application_dispatch.go
+++ b/event/application_dispatch.go
@@ -53,10 +53,9 @@ func (app *Application) Dispatch(evt any, payload ...[]event.Arg) []any {
 // Returns: The first return value from the function, or nil if function panics or returns nothing
 func (app *Application) call(fn reflect.Value, args []reflect.Value) any {
 	defer func() {
-		if r := recover(); r != nil {
-			// Panic occurred but we recover silently to prevent one bad listener
-			// from breaking the entire event dispatch chain
-		}
+		// Recover from panics to prevent one bad listener from breaking
+		// the entire event dispatch chain
+		_ = recover()
 	}()
 	if results := fn.Call(args); len(results) > 0 {
 		return results[0].Interface()


contracts/event/events.go
```diff
@@ -1,40 +1,75 @@
...
 
 type Event interface {
 	// Handle the event.
 	Handle(args []Arg) ([]Arg, error)
 }
 
-type Listener interface {
-	// Signature returns the unique identifier for the listener.
-	Signature() string
+type EventListener interface {
```
hwbrzzl, line 13: Can it be merged into `EventQueueListener`? Don't find it used elsewhere.

contracts/event/events.go
```diff
@@ -1,40 +1,75 @@
...
+	// Handle the event.
+	Handle(event any, args ...any) error
+}
+
+type EventQueueListener interface {
```
hwbrzzl, line 18: The `Event` prefix is duplicated with the package name.

```suggestion
type QueueListener interface {
```

contracts/event/events.go
```diff
@@ -1,40 +1,75 @@
...
 }
 
 type Queue struct {
 	Connection string
-	Queue      string
 	Enable     bool
+	Queue      string
+}
+
+type QueueListener interface {
```
hwbrzzl, line 58: Why add this function instead of using `Listener` directly?

contracts/event/events.go
hwbrzzl: `make:listener` should create a listener according to the new interface. It can be implemented in this PR or another.

contracts/event/events.go
```diff
@@ -1,40 +1,75 @@
...
+	Dispatch(event any, payload ...[]Arg) []any
+	// GetEvents gets all registered events.
+	GetEvents() map[Event][]Listener
+	// Job create a new event task.
+	// Deprecated: Use Dispatch instead. Job will be removed in a future version.
+	Job(event Event, args []Arg) Task
+	// Listen registers an event listener with the dispatcher.
+	// events can be: string, []string, Event, []Event, or any other type
+	// listener can be: function, class, or any callable
+	Listen(events any, listener ...any) error
```
hwbrzzl, line 38: The new `Listen` function can only support the new listener `EventQueueListener` for now. It doesn't need to support the old listener interface. We can notice this in the annotation, let users use the Register function for old listeners.

event/application_listen.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	return nil
+}
+
+func (app *Application) handleClosure(fn reflect.Value) error {
+	fnType := fn.Type()
+
+	if fnType.NumIn() != 1 || !app.isEventType(fnType.In(0)) {
+		return errors.New("closure must accept exactly one event parameter")
```
hwbrzzl, line 78: Please add all errors to `errors/list.go`, we will implement i18n for the error in the future.

contracts/event/events.go
```diff
@@ -1,40 +1,75 @@
...
+	// Register event listeners to the application.
+	Register(map[Event][]Listener)
```
hwbrzzl, line 40: Add the DEPRECATED tag for this function.

event/application_listen.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func (app *Application) setupEvents(e any, listener any) {
+	eventName := app.getEventName(e)
+
+	// Register queued listeners with the queue system
+	if l, ok := listener.(event.EventQueueListener); ok {
+		// Capture event in local scope to avoid closure variable capture bug
+		eventCopy := e
+		app.queue.Register([]queue.Job{
```
hwbrzzl, line 99: The `Register` function will be deprecated in the future, so we can implement new logic for the Listen flow, and the new flow should be the main, the deprecated `Job` and `Register` functions should use the new flow.

event/utils.go
```diff
@@ -0,0 +1,78 @@
...
+func (j *dynamicQueueJob) Signature() string {
+	return j.signature
+}
+
+// getEventName returns the name of the event.
+// examples:
+//
+//	getEventName("user.created") // "user.created"
+//	getEventName(&UserCreated{}) // "UserCreated"
+func (app *Application) getEventName(evt any) string {
```
hwbrzzl, line 45: It can be moved to `application_listen.go`. And how about returning an error when the name is empty?

event/utils.go
```diff
@@ -0,0 +1,78 @@
...
+			return signature.Signature()
+		}
+
+		// Use reflection to get struct type name
+		eventType := reflect.TypeOf(e)
+		if eventType.Kind() == reflect.Ptr {
+			eventType = eventType.Elem()
+		}
+
+		return eventType.Name()
```
hwbrzzl, line 61: The package name can be taken as well.

event/utils.go
```diff
@@ -0,0 +1,78 @@
...
+		if eventType != nil {
+			if eventType.Kind() == reflect.Ptr {
+				eventType = eventType.Elem()
+			}
+			// For unnamed structs, use the full type string
+			name := eventType.Name()
+			if name == "" {
+				name = eventType.String()
+			}
+			return name
```
hwbrzzl, line 74: Ditto

event/utils_test.go
```diff
@@ -0,0 +1,233 @@
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result := app.getEventName(tt.event)
+			assert.Equal(t, tt.expected, result)
+		})
+	}
+}
+
+func TestApplication_canUse(t *testing.T) {
```
hwbrzzl, line 109: Please move the case to `application_dispatch_test.go`

event/utils_test.go
```diff
@@ -0,0 +1,233 @@
...
+	
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := app.canUse(tt.val, tt.target)
+			assert.Equal(t, tt.want, got)
+		})
+	}
+}
+
+func TestApplication_convert(t *testing.T) {
```
hwbrzzl, line 158: Ditto

event/utils_test.go
```diff
@@ -0,0 +1,233 @@
...
+			}
+			
+			if tt.target.Kind() == reflect.String && tt.expectedStr != "" {
+				assert.Equal(t, tt.expectedStr, got.String())
+			}
+		})
+	}
+}
+
+func Test_eventArgsToQueueArgs(t *testing.T) {
```
hwbrzzl, line 216: Move it to `task_test.go`

event/application_listen.go
```diff
@@ -0,0 +1,131 @@
...
+//	app.Listen([]string{"user.created", "user.updated"}, listener)
+//
+//	// Listen to wildcard events
+//	app.Listen("user.*", func(args ...any) error {
+//		return nil
+//	})
+//
+//	// Listen with event structs
+//	app.Listen(UserCreatedEvent{}, &UserCreatedListener{})
+func (app *Application) Listen(events any, listener ...any) error {
```
hwbrzzl, line 37: Can multiple listeners be registered here?

contracts/event/events.go
```diff
@@ -1,40 +1,75 @@
...
+	Signature
+	ShouldQueue
 }
 
-type Task interface {
-	// Dispatch an event and call the listeners.
-	Dispatch() error
+type Instance interface {
+	// Dispatch fires an event and calls the listeners.
+	Dispatch(event any, payload ...[]Arg) []any
```
hwbrzzl, line 29: `[]any` should be `[]error`, right? How about adding a `Result` interface?

```
type Result interface {
  Error() error // use errors.Join to zip the errors
  Errors() []error
  Failed() bool
}
```

event/application_dispatch.go
```diff
@@ -0,0 +1,328 @@
...
+	if v.Kind() != reflect.Func {
+		if method := v.MethodByName("Handle"); method.IsValid() {
+			v = method
+		} else {
+			return nil
+		}
+	}
+
+	// Call with dynamic args
+	return app.callListenerHandle(v, evt, args)
```
hwbrzzl, line 101: Another assertion can be used here instead of `v.MethodByName("Handle")`, it's complex to implement `callListenerHandle` and is hard to maintain in the future.

event/application_dispatch.go
```diff
@@ -0,0 +1,328 @@
...
+// for patterns that match the event name.
+//
+// The results are cached in wildcardsCache for performance. The cache is automatically
+// cleared when new wildcard listeners are registered.
+//
+// Parameters:
+//   - eventName: The event name to match against wildcard patterns
+//
+// Returns: A slice of listeners from all matching wildcard patterns
+func (app *Application) getWildcardListeners(eventName any) []any {
```
hwbrzzl, line 195: This function is only called by `prepareListeners`, `eventName` is already `string`.

event/application_dispatch.go
```diff
@@ -0,0 +1,328 @@
...
+	var wildcardListeners []any
+	eventNameStr := app.getEventName(eventName)
+
+	for event, wildcard := range app.wildcards {
+		if str.Of(eventNameStr).Is(app.getEventName(event)) {
```
hwbrzzl, line 200: `event` is already string.

event/application_dispatch.go
```diff
@@ -0,0 +1,328 @@
...
+//
+// Returns: nil (queuing is asynchronous)
+func (app *Application) queueListener(listener event.QueueListener, payload []event.Arg) any {
+	task := app.queue.Job(listener, eventArgsToQueueArgs(payload))
+
+	// Configure queue task using interface assertions instead of reflection
+	// This is significantly faster than MethodByName approach
+
+	// Configure connection if listener implements ViaConnection
+	if configured, ok := listener.(interface{ ViaConnection() string }); ok {
```
hwbrzzl, line 309: Define `ViaConnection`, `ViaQueue`, and `WithDelay` in the contracts/event/events.go file.

event/application_dispatch.go
```diff
@@ -0,0 +1,328 @@
...
+	}
+
+	// Configure delay if listener implements WithDelay
+	if configured, ok := listener.(interface{ WithDelay() int64 }); ok {
+		task.Delay(time.Now().Add(time.Duration(configured.WithDelay()) * time.Second))
+	}
+
+	// Dispatch task - errors are intentionally not returned as per framework pattern
+	// Queued jobs handle their own error recovery and reporting
+	_ = task.Dispatch()
```
hwbrzzl, line 325: The error should be returned.

event/application_dispatch.go
```diff
@@ -0,0 +1,328 @@
...
+
+// queueListener dispatches a listener to the queue system.
+// Configures the queue task based on listener configuration methods.
+//
+// Parameters:
+//   - listener: The queue listener to dispatch
+//   - payload: Optional payload arguments
+//
+// Returns: nil (queuing is asynchronous)
+func (app *Application) queueListener(listener event.QueueListener, payload []event.Arg) any {
```
hwbrzzl, line 302:
```suggestion
func (app *Application) queueListener(listener event.QueueListener, payload []event.Arg) error {
```

event/application_dispatch.go
```diff
@@ -0,0 +1,328 @@
...
+// queueEventListener creates a dynamic queue job for an event-aware listener.
+// Wraps the listener in a closure that preserves the event context.
+//
+// Parameters:
+//   - listener: The event queue listener to wrap
+//   - evt: The event being dispatched
+//   - args: Optional payload arguments
+//
+// Returns: nil (queuing is asynchronous)
+func (app *Application) queueEventListener(listener event.EventQueueListener, evt any, args []event.Arg) any {
```
hwbrzzl, line 274:
```suggestion
func (app *Application) queueEventListener(listener event.EventQueueListener, evt any, args []event.Arg) error {
```

event/application_dispatch.go
```diff
@@ -0,0 +1,328 @@
...
+// The method uses reflection to dynamically match listener parameters with
+// the event and payload arguments.
+//
+// Parameters:
+//   - listener: The listener to invoke (function, struct with Handle method, or queue listener)
+//   - evt: The event being dispatched
+//   - args: Optional payload arguments
+//
+// Returns: The listener's return value, or nil if no value returned
+func (app *Application) callListener(listener any, evt any, args []event.Arg) any {
```
hwbrzzl, line 81:
```suggestion
func (app *Application) callListener(listener any, evt any, args []event.Arg) error {
```

event/application_listen.go
```diff
@@ -0,0 +1,131 @@
...
+	app.mu.Lock()
+	defer app.mu.Unlock()
+
+	if len(listener) == 0 {
+		// it may be a closure
+		if fn := reflect.ValueOf(events); fn.Kind() == reflect.Func {
+			return app.handleClosure(fn)
+		}
+
+		return errors.New("listener is required")
```
hwbrzzl, line 47: Users have to judge each error when listening to several events. Not sure if there is a better way to optimize the action.

hwbrzzl (commented):
Thanks, amazing PR, left some questions. PTAL.

hwbrzzl:
Closed this PR first, feel free to open it if needed.

PR #1134, opened 2025-07-19
feat: laravel collection

ahmed3mar:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/566

### Creating Collections

```go
// Create from variadic arguments
numbers := collect.New(1, 2, 3, 4, 5)

// Create from slice
items := []string{"apple", "banana", "cherry"}
fruits := collect.Of(items)
```

### Basic Operations

```go
numbers := collect.New(1, 2, 3, 4, 5)

// Get count
fmt.Println(numbers.Count()) // 5

// Get first/last
fmt.Println(*numbers.First()) // 1
fmt.Println(*numbers.Last())  // 5

// Check if empty
fmt.Println(numbers.IsEmpty()) // false

// Get all items
fmt.Println(numbers.All()) // [1 2 3 4 5]
```

### Filtering and Mapping

```go
numbers := collect.New(1, 2, 3, 4, 5, 6)

// Filter even numbers
evens := numbers.Filter(func(n int, _ int) bool {
    return n%2 == 0
})
fmt.Println(evens.All()) // [2 4 6]

// Map to double values (Laravel-style method)
doubled := numbers.Map(func(n int, _ int) interface{} {
    return n * 2
})
fmt.Println(doubled.All()) // [2 4 6 8 10 12]

// Map to string representation
strings := numbers.Map(func(n int, i int) interface{} {
    return fmt.Sprintf("item_%d_%d", i, n)
})
fmt.Println(strings.All()) // [item_0_1 item_1_2 item_2_3 item_3_4 item_4_5 item_5_6]

### Working with Structs

```go
type User struct {
    ID   int
    Name string
    Age  int
}

users := collect.Of([]User{
    {ID: 1, Name: "Alice", Age: 25},
    {ID: 2, Name: "Bob", Age: 30},
    {ID: 3, Name: "Charlie", Age: 25},
})

// Filter by struct field using different Where patterns
youngUsers := users.Where("Age", "=", 25)        // 3 parameters
youngUsers = users.Where("Age", 25)              // 2 parameters (implies '=')
fmt.Println(youngUsers.Count()) // 2

// Using callback function
adultUsers := users.Where(func(u User) bool {
    return u.Age >= 18
})

// Using different operators
olderUsers := users.Where("Age", ">", 25)
notCharlie := users.Where("Name", "!=", "Charlie")

// Handling null values
activeUsers := users.Where("DeletedAt", "=", nil)    // Find non-deleted users
deletedUsers := users.Where("DeletedAt", "!=", nil)  // Find deleted users

// Group by field
grouped := users.GroupBy(func(u User) string {
    return fmt.Sprintf("%d", u.Age)
})
fmt.Println(len(grouped)) // 2 groups

// Pluck field values
names := users.Pluck("Name")
fmt.Println(names.Count()) // 3

// Sort by field
sorted := users.SortBy(func(u User) string {
    return u.Name
})
fmt.Println(sorted.First().Name) // "Alice"
```

### Aggregation Methods

```go
numbers := collect.New(1, 2, 3, 4, 5)

// Sum
sum := numbers.Sum(func(n int) float64 {
    return float64(n)
})
fmt.Println(sum) // 15.0

// Average
avg := numbers.Avg(func(n int) float64 {
    return float64(n)
})
fmt.Println(avg) // 3.0

// Min/Max
min := numbers.Min(func(n int) float64 { return float64(n) })
max := numbers.Max(func(n int) float64 { return float64(n) })
fmt.Println(min, max) // 1.0 5.0
```

### Conditional Operations

```go
numbers := collect.New(1, 2, 3, 4, 5)

// Conditional transformations
result := numbers.
    When(true, func(c *collect.Collection[int]) *collect.Collection[int] {
        return c.Filter(func(n int, _ int) bool { return n > 2 })
    }).
    Unless(false, func(c *collect.Collection[int]) *collect.Collection[int] {
        return c.Take(2)
    })

fmt.Println(result.All()) // [3 4]

// Tap for side effects
numbers.Tap(func(c *collect.Collection[int]) {
    fmt.Println("Processing", c.Count(), "items")
})
```

### Core Methods (Alphabetical)

- `After(value)` - Get item after given value
- `All()` - Get all items as slice
- `Average(keyFunc)` - Calculate average using key function
- `Before(value)` - Get item before given value
- `Chunk(size)` - Split into chunks of given size
- `Clone()` - Create a copy of the collection
- `Collapse()` - Collapse nested arrays into single array
- `Combine(keys)` - Combine with keys to create map
- `Contains(value)` - Check if collection contains value
- `Count()` - Get item count
- `Diff(other)` - Get difference with another collection
- `Each(func)` - Iterate over each item
- `Every(predicate)` - Check if all items match predicate
- `Filter(predicate)` - Filter items by predicate
- `First()` - Get first item
- `Flatten()` - Flatten nested structures
- `GroupBy(keyFunc)` - Group items by key function
- `Intersect(other)` - Get intersection with another collection
- `IsEmpty()` - Check if collection is empty
- `IsNotEmpty()` - Check if collection is not empty
- `Join(separator)` - Join items with separator
- `Last()` - Get last item
- `Map(func)` - Transform each item with a function (returns Collection[interface{}])
- `Merge(other)` - Merge with another collection
- `Partition(predicate)` - Split into two collections by predicate
- `Pluck(field)` - Extract field values
- `Push(items...)` - Add items to end
- `Reverse()` - Reverse order
- `Search(value)` - Find index of value
- `Slice(start, length)` - Get slice of items
- `Sort(lessFunc)` - Sort by comparison function
- `SortBy(keyFunc)` - Sort by key function
- `Sum(keyFunc)` - Calculate sum using key function
- `Take(n)` - Take first n items
- `Unique()` - Get unique items
- `Where(field, operator, value)` - Filter by field comparison
- `Zip(other)` - Zip with another collection

### Where Method - Laravel-style Filtering

The `Where` method supports multiple patterns for flexible filtering:

```go
type User struct {
    ID        int
    Name      string
    Age       int
    Country   string
    Balance   float64
    DeletedAt *time.Time
}

users := collect.Of([]User{...})

// 1. Two parameters (field, value) - implies '=' operator
frenchUsers := users.Where("Country", "FR")
youngUsers := users.Where("Age", 25)

// 2. Three parameters (field, operator, value)
richUsers := users.Where("Balance", ">", 100.0)
nonFrenchUsers := users.Where("Country", "!=", "FR")
seniorUsers := users.Where("Age", ">=", 65)

// 3. Single parameter (callback function)
customFilter := users.Where(func(u User) bool {
    return u.Age > 18 && u.Country == "US"
})

// 4. Null comparisons
activeUsers := users.Where("DeletedAt", "=", nil)
deletedUsers := users.Where("DeletedAt", "!=", nil)

// 5. String operations
nameContains := users.Where("Name", "like", "john")
excludePattern := users.Where("Name", "not like", "test")
```

**Supported Operators:**
- `=`, `==` - Equality
- `!=` - Inequality  
- `>`, `>=` - Greater than, Greater than or equal
- `<`, `<=` - Less than, Less than or equal
- `like` - Case-insensitive substring match
- `not like` - Case-insensitive substring exclusion

### Utility Methods

- `Debug()` - Print collection contents
- `Dump()` - Print collection contents
- `Tap(func)` - Execute function and return collection
- `ToJSON()` - Convert to JSON string
- `When(condition, func)` - Execute function if condition is true
- `Unless(condition, func)` - Execute function if condition is false

## Map Method - Laravel-style Transformation

The `Map` method provides Laravel-style transformation capabilities:

```go
type User struct {
    ID   int
    Name string
    Age  int
}

users := collect.Of([]User{
    {ID: 1, Name: "Alice", Age: 25},
    {ID: 2, Name: "Bob", Age: 30},
})

// Transform to different types
names := users.Map(func(u User, i int) interface{} {
    return u.Name
})

ages := users.Map(func(u User, i int) interface{} {
    return u.Age
})

// Complex transformations
summaries := users.Map(func(u User, i int) interface{} {
    return map[string]interface{}{
        "id":      u.ID,
        "summary": fmt.Sprintf("%s (%d years)", u.Name, u.Age),
        "index":   i,
    }
})

// Chain with other operations
result := users.
    Map(func(u User, i int) interface{} {
        return u.Name
    }).
    Filter(func(name interface{}, _ int) bool {
        return len(name.(string)) > 3
    })
```

## Generic Functions

Some operations require type transformation and are provided as generic functions for type safety:

```go
// Type-safe Map to different type
strings := collect.New("1", "2", "3")
numbers := collect.Map(strings, func(s string, _ int) int {
    n, _ := strconv.Atoi(s)
    return n
})

// Reduce to single value
sum := collect.Reduce(numbers, func(acc int, n int, _ int) int {
    return acc + n
}, 0)

// Pluck with type conversion
users := collect.Of([]User{...})
userIDs := collect.Pluck[User, int](users, "ID")
```

## Testing

```bash
go test -v
```

## Examples

See `example_test.go` for comprehensive usage examples.

## LazyCollection

LazyCollection provides lazy evaluation for efficient processing of large datasets. Operations are not executed until a terminal operation is called.

### Creating LazyCollections

```go
// From slice
lazy := collect.LazyCollect([]int{1, 2, 3, 4, 5})

// From range
lazy := collect.LazyRange(1, 1000000)

// From generator function
lazy := collect.LazyGenerate(func(i int) int {
    return i * 2
}, 100)

// From channel
ch := make(chan int, 5)
// ... populate channel
lazy := collect.LazyFromChannel(ch)
```

### Lazy Operations

```go
// Chain operations - these don't execute until consumed
result := collect.LazyRange(1, 1000000).
    Filter(func(n int, _ int) bool { return n%2 == 0 }).
    Take(5).
    All() // This triggers execution

fmt.Println(result) // [2 4 6 8 10]
```

### Performance Benefits

```go
// Efficient for large datasets when only a portion is needed
result := collect.LazyRange(1, 1000000).
    Filter(func(n int, _ int) bool { return n%100 == 0 }).
    Take(10).
    All()

// Only processes what's needed, not all 1 million items
```

### Lazy vs Eager

```go
// Lazy - processes only what's needed
lazyResult := collect.LazyRange(1, 1000000).
    Filter(func(n int, _ int) bool { return n%2 == 0 }).
    Take(5).
    All()

// Eager - processes all items first
eagerResult := collect.New(/* large slice */).
    Filter(func(n int, _ int) bool { return n%2 == 0 }).
    Take(5).
    All()
```

### Converting Between Collections

```go
// Lazy to eager
lazy := collectioncollectLazyRange(1, 11)
eager := lazy.Collect()

// Eager to lazy
eager := collect.New(1, 2, 3, 4, 5)
lazy := collect.LazyCollect(eager.All())
```

### LazyCollection Methods

- `All()` - Materialize all items
- `Count()` - Count items
- `Filter(predicate)` - Filter items
- `Map(func)` - Transform each item with a function (returns LazyCollection[interface{}])
- `Where(params...)` - Laravel-style filtering (supports all same patterns as Collection)
- `Take(n)` - Take first n items
- `Skip(n)` - Skip first n items
- `TakeWhile(predicate)` - Take while condition is true
- `DropWhile(predicate)` - Drop while condition is true
- `Unique()` - Get unique items
- `Sort(lessFunc)` - Sort items
- `Reverse()` - Reverse order
- `Sum(keyFunc)` - Calculate sum
- `Average(keyFunc)` - Calculate average
- `Min(keyFunc)` - Find minimum
- `Max(keyFunc)` - Find maximum
- `GroupBy(keyFunc)` - Group items
- `Partition(predicate)` - Split into two collections
- `FlatMap(func)` - Flat map transformation
- `Each(func)` - Execute function for each item
- `ForEach(func)` - Execute function for each item (consumes collection)
- `Iterator()` - Get iterator for manual control
- `Collect()` - Convert to eager Collection

## License

MIT License

## ✅ Checks

### DIFF
--- a/support/collect/collection_test.go
+++ b/support/collect/collection_test.go
@@ -1469,3 +1469,354 @@ func TestSortByDesc(t *testing.T) {
 		t.Error("Expected original unchanged and empty sort")
 	}
 }
+
+// ===== PHASE 2: ADDITIONAL METHODS =====
+
+func TestChunkWhile(t *testing.T) {
+	c := New(1, 2, 2, 3, 3, 3, 4)
+	chunks := c.ChunkWhile(func(item int, _ int, chunk []int) bool { return len(chunk) == 0 || item == chunk[0] })
+	if len(chunks) != 4 || len(chunks[0]) != 1 || len(chunks[1]) != 2 || len(chunks[2]) != 3 {
+		t.Error("Expected chunks grouped by same values")
+	}
+	if len(New[int]().ChunkWhile(func(int, int, []int) bool { return true })) != 0 {
+		t.Error("Expected empty result for empty collection")
+	}
+}
+
+func TestCollectionCountBy(t *testing.T) {
+	type Product struct{ Type string }
+	products := New(Product{"fruit"}, Product{"veg"}, Product{"fruit"}, Product{"meat"})
+	counts := products.CountBy(func(p Product) string { return p.Type })
+	if counts["fruit"] != 2 || counts["veg"] != 1 || counts["meat"] != 1 {
+		t.Error("Expected correct category counts")
+	}
+	if len(New[Product]().CountBy(func(p Product) string { return p.Type })) != 0 {
+		t.Error("Expected empty map for empty collection")
+	}
+}
+
+func TestCrossJoin(t *testing.T) {
+	c1 := New(1, 2)
+	c2 := New(3, 4)
+	result := c1.CrossJoin(c2)
+	if len(result) != 4 || !reflect.DeepEqual(result[0], []int{1, 3}) || !reflect.DeepEqual(result[3], []int{2, 4}) {
+		t.Error("Expected correct cross join")
+	}
+	if len(New[int]().CrossJoin(c2)) != 0 || len(c1.CrossJoin(New[int]())) != 0 {
+		t.Error("Expected empty for empty collections")
+	}
+}
+
+func TestDiffAssoc(t *testing.T) {
+	c1 := New(1, 2, 3, 4)
+	c2 := New(1, 2, 5, 6)
+	diff := c1.DiffAssoc(c2)
+	if !reflect.DeepEqual(diff.All(), []int{3, 4}) {
+		t.Errorf("Expected [3, 4], got %v", diff.All())
+	}
+	if New[int]().DiffAssoc(c2).Count() != 0 || c1.DiffAssoc(New[int]()).Count() != 4 {
+		t.Error("Expected correct diff for empty collections")
+	}
+}
+
+func TestDropUntil(t *testing.T) {
+	c := New(1, 2, 3, 4, 5)
+	result := c.DropUntil(func(n int) bool { return n >= 3 })
+	if !reflect.DeepEqual(result.All(), []int{3, 4, 5}) {
+		t.Error("Expected [3, 4, 5]")
+	}
+	if c.DropUntil(func(n int) bool { return n > 10 }).Count() != 0 || New[int]().DropUntil(func(int) bool { return true }).Count() != 0 {
+		t.Error("Expected empty for no match and empty collection")
+	}
+}
+
+func TestDropWhile(t *testing.T) {
+	c := New(1, 2, 3, 4, 5)
+	result := c.DropWhile(func(n int) bool { return n < 3 })
+	if !reflect.DeepEqual(result.All(), []int{3, 4, 5}) {
+		t.Error("Expected [3, 4, 5]")
+	}
+	if c.DropWhile(func(n int) bool { return n < 10 }).Count() != 0 || New[int]().DropWhile(func(int) bool { return false }).Count() != 0 {
+		t.Error("Expected empty for all dropped and empty collection")
+	}
+}
+
+func TestDump(t *testing.T) {
+	c := New(1, 2, 3)
+	result := c.Dump()
+	if result != c || result.Count() != 3 {
+		t.Error("Expected Dump to return same collection unchanged")
+	}
+}
+
+func TestEachSpread(t *testing.T) {
+	count := 0
+	c := New(1, 2, 3)
+	c.EachSpread(func(items ...int) {
+		count += len(items)
+	})
+	if count != 3 {
+		t.Errorf("Expected count of 3, got %d", count)
+	}
+}
+
+func TestFlatMap(t *testing.T) {
+	c := New(1, 2, 3)
+	result := c.FlatMap(func(n int) []int { return []int{n, n * 2} })
+	if !reflect.DeepEqual(result.All(), []int{1, 2, 2, 4, 3, 6}) {
+		t.Error("Expected [1, 2, 2, 4, 3, 6]")
+	}
+	if New[int]().FlatMap(func(n int) []int { return []int{n} }).Count() != 0 {
+		t.Error("Expected empty for empty collection")
+	}
+}
+
+func TestForPage(t *testing.T) {
+	c := New(1, 2, 3, 4, 5, 6, 7, 8, 9, 10)
+	if !reflect.DeepEqual(c.ForPage(1, 3).All(), []int{1, 2, 3}) || !reflect.DeepEqual(c.ForPage(2, 3).All(), []int{4, 5, 6}) {
+		t.Error("Expected correct pagination")
+	}
+	if c.ForPage(10, 3).Count() != 0 || New[int]().ForPage(1, 3).Count() != 0 {
+		t.Error("Expected empty for out of bounds and empty collection")
+	}
+}
+
+func TestIntersectByKeys(t *testing.T) {
+	c1 := New(10, 20, 30, 40, 50)
+	c2 := New(1, 2, 3)
+	result := c1.IntersectByKeys(c2)
+	if !reflect.DeepEqual(result.All(), []int{10, 20, 30}) {
+		t.Error("Expected first 3 elements")
+	}
+	if New[int]().IntersectByKeys(c2).Count() != 0 || c1.IntersectByKeys(New[int]()).Count() != 0 {
+		t.Error("Expected empty for empty collections")
+	}
+}
+
+func TestCollectionKeys(t *testing.T) {
+	c := New(10, 20, 30)
+	keys := c.Keys()
+	if !reflect.DeepEqual(keys, []int{0, 1, 2}) {
+		t.Error("Expected [0, 1, 2]")
+	}
+	if len(New[int]().Keys()) != 0 {
+		t.Error("Expected empty keys for empty collection")
+	}
+}
+
+func TestMapSpread(t *testing.T) {
+	c := New(1, 2, 3)
+	result := c.MapSpread(func(items ...int) int {
+		sum := 0
+		for _, v := range items {
+			sum += v
+		}
+		return sum
+	})
+	if result.Count() != 3 || result.All()[0] != 1 || result.All()[1] != 2 || result.All()[2] != 3 {
+		t.Error("Expected correct MapSpread results")
+	}
+}
+
+func TestMapToDictionary(t *testing.T) {
+	type Product struct {
+		Name string
+		Type string
+	}
+	products := New(Product{"Apple", "fruit"}, Product{"Carrot", "veg"}, Product{"Banana", "fruit"})
+	dict := products.MapToDictionary(func(p Product) string { return p.Type })
+	if len(dict["fruit"]) != 2 || len(dict["veg"]) != 1 {
+		t.Error("Expected correct grouping by type")
+	}
+	if len(New[Product]().MapToDictionary(func(p Product) string { return p.Type })) != 0 {
+		t.Error("Expected empty dictionary for empty collection")
+	}
+}
+
+func TestMedian(t *testing.T) {
+	c := New(1, 2, 3, 4, 5)
+	if c.Median(func(n int) float64 { return float64(n) }) != 3.0 {
+		t.Error("Expected median of 3")
+	}
+	c2 := New(1, 2, 3, 4)
+	if c2.Median(func(n int) float64 { return float64(n) }) != 2.5 {
+		t.Error("Expected median of 2.5 for even count")
+	}
+	if New[int]().Median(func(n int) float64 { return float64(n) }) != 0 {
+		t.Error("Expected 0 for empty collection")
+	}
+}
+
+func TestMergeRecursive(t *testing.T) {
+	c1 := New(1, 2, 3)
+	c2 := New(4, 5, 6)
+	result := c1.MergeRecursive(c2)
+	if !reflect.DeepEqual(result.All(), []int{1, 2, 3, 4, 5, 6}) {
+		t.Error("Expected merged collection")
+	}
+	if New[int]().MergeRecursive(c2).Count() != 3 {
+		t.Error("Expected correct merge with empty")
+	}
+}
+
+func TestMode(t *testing.T) {
+	type Item struct{ Category string }
+	items := New(Item{"A"}, Item{"B"}, Item{"A"}, Item{"C"}, Item{"A"})
+	modes := items.Mode(func(i Item) string { return i.Category })
+	if len(modes) != 1 || modes[0] != "A" {
+		t.Error("Expected mode to be [A]")
+	}
+	if len(New[Item]().Mode(func(i Item) string { return i.Category })) != 0 {
+		t.Error("Expected empty modes for empty collection")
+	}
+}
+
+func TestNth(t *testing.T) {
+	c := New(1, 2, 3, 4, 5, 6, 7, 8, 9, 10)
+	if !reflect.DeepEqual(c.Nth(2).All(), []int{1, 3, 5, 7, 9}) {
+		t.Error("Expected every 2nd element starting from 0")
+	}
+	if c.Nth(0).Count() != 0 || c.Nth(-1).Count() != 0 || New[int]().Nth(2).Count() != 0 {
+		t.Error("Expected empty for invalid n and empty collection")
+	}
+}
+
+func TestPad(t *testing.T) {
+	c := New(1, 2, 3)
+	result := c.Pad(5, 0)
+	if !reflect.DeepEqual(result.All(), []int{1, 2, 3, 0, 0}) {
+		t.Error("Expected padded with zeros")
+	}
+	if c.Pad(2, 0).Count() != 3 {
+		t.Error("Expected original size when pad size is smaller")
+	}
+}
+
+func TestReplace(t *testing.T) {
+	c := New(1, 2, 3, 4, 5)
+	result := c.Replace(map[int]int{0: 10, 2: 30, 10: 100})
+	if !reflect.DeepEqual(result.All(), []int{10, 2, 30, 4, 5}) {
+		t.Error("Expected replaced values at indices 0 and 2")
+	}
+	if !reflect.DeepEqual(c.All(), []int{1, 2, 3, 4, 5}) {
+		t.Error("Expected original collection unchanged")
+	}
+}
+
+func TestCollectionShuffle(t *testing.T) {
+	c := New(1, 2, 3, 4, 5)
+	result := c.Shuffle()
+	if result.Count() != 5 || !reflect.DeepEqual(c.All(), []int{1, 2, 3, 4, 5}) {
+		t.Error("Expected same count and original unchanged")
+	}
+	if New[int]().Shuffle().Count() != 0 {
+		t.Error("Expected empty shuffle for empty collection")
+	}
+}
+
+func TestSkipUntil(t *testing.T) {
+	c := New(1, 2, 3, 4, 5)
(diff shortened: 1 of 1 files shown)


support/collect/collection.go
```diff
@@ -0,0 +1,1334 @@
...
+				}
+			}
+		} else {
+			flattened = append(flattened, item)
+		}
+	}
+	return &Collection[T]{items: flattened}
+}
+
+func Collect[T any](items []T) *Collection[T] {
```
almas-x: How about changing `collect.Collect` to `collect.Of`? This way, it avoids redundancy and becomes more concise.

support/collect/lazy_collection.go
```diff
@@ -0,0 +1,871 @@
...
+		var zero T
+		return zero, false
+	}
+
+	return value, true
+}
+
+func (it *lazyIterator[T]) Reset() {
+	it.done = false
+}
```
hwbrzzl, line 44: I think the function can be used as expect when `it.ch` is closed, could you confirm this and add a test case for it?

support/collect/lazy_collection.go
```diff
@@ -0,0 +1,871 @@
...
+	copy(newPipeline, lc.pipeline)
+
+	newPipeline = append(newPipeline, func(input <-chan T) <-chan T {
+		output := make(chan T)
+		go func() {
+			defer close(output)
+			taken := 0
+			for item := range input {
+				if taken >= n {
+					break
```
hwbrzzl, line 669: There may still be data in the `input` chan, but `break` directly. goroutine will leak. The same as `TakeWhile `

support/collect/lazy_collection.go
```diff
@@ -0,0 +1,871 @@
...
+	newPipeline := make([]func(<-chan T) <-chan T, len(lc.pipeline))
+	copy(newPipeline, lc.pipeline)
+
+	newPipeline = append(newPipeline, func(input <-chan T) <-chan T {
+		output := make(chan T)
+		go func() {
+			defer close(output)
+			seen := make(map[string]bool)
+			for item := range input {
+				key := fmt.Sprintf("%v", item)
```
hwbrzzl, line 735: The `key` will duplicated when `item` is struct or map, is there a better way to deal with such situation?

support/collect/lazy_collection.go
```diff
@@ -0,0 +1,871 @@
...
+	})
+}
+
+func (lc *LazyCollection[T]) Zip(other *LazyCollection[T]) [][]T {
+	var result [][]T
+	ch1 := lc.execute()
+	ch2 := other.execute()
+
+	for item1 := range ch1 {
+		item2, ok2 := <-ch2
```
hwbrzzl, line 853: The same as above, `ch1` and `ch2` may be leak.


```suggestion
	ch1 := lc.execute()
	ch2 := other.execute()
	
	defer func() {
        go func() { for range ch1 {} }()
        go func() { for range ch2 {} }()
    }()

	for item1 := range ch1 {
		item2, ok2 := <-ch2
```

support/collect/collection.go
```diff
@@ -0,0 +1,1334 @@
...
+
+	for _, item := range c.items {
+		key := fmt.Sprintf("%v", item)
+		if seen[key] {
+			duplicates = append(duplicates, item)
+		} else {
+			seen[key] = true
+		}
+	}
+	return &Collection[T]{items: duplicates}
```
hwbrzzl, line 272: If the collecction is `3,3,3`, `duplicates` will be `3,3`. It's unexpected.

support/collect/collection.go
```diff
@@ -0,0 +1,1334 @@
...
+	fmt.Printf("Collection: %+v\n", c.items)
+	return c
+}
+
+func (c *Collection[T]) Duplicates() *Collection[T] {
+	seen := make(map[string]bool)
+	var duplicates []T
+
+	for _, item := range c.items {
+		key := fmt.Sprintf("%v", item)
```
hwbrzzl, line 265: The same as LazyCollection, `%v` is unused for struct and map.

support/collect/collection.go
```diff
@@ -0,0 +1,1334 @@
...
+	}
+	return result
+}
+
+func (c *Collection[T]) Max(keyFunc func(T) float64) float64 {
+	if len(c.items) == 0 {
+		return 0
+	}
+
+	max := keyFunc(c.items[0])
```
krishankumar01, line 613: How about renaming this variable to something else, since it collides with Go’s built-in max method (and similarly for min)?

support/collect/collection.go
```diff
@@ -0,0 +1,1334 @@
...
+	}
+	c.items[index] = value
+	return c
+}
+
+func (c *Collection[T]) Random() *T {
+	if len(c.items) == 0 {
+		return nil
+	}
+	rng := rand.New(rand.NewSource(time.Now().UnixNano()))
```
krishankumar01, line 800: Initializing a random generator every time is slower. How about we initialize a global generator and reuse it? And why this method is returning a pointer of type T?

support/collect/collection.go
```diff
@@ -0,0 +1,1334 @@
...
+	return &c.items[rng.Intn(len(c.items))]
+}
+
+func Reduce[T, R any](c *Collection[T], fn func(R, T, int) R, initial R) R {
+	result := initial
+	for i, item := range c.items {
```
krishankumar01, line 806: How about we keep the input order as the index and use the remaining arguments for the fn(same for all the places where we pass index in function)?

support/collect/collection.go
```diff
@@ -0,0 +1,1334 @@
...
+func (c *Collection[T]) SearchBy(predicate func(T) bool) int {
+	for i, item := range c.items {
+		if predicate(item) {
+			return i
+		}
+	}
+	return -1
+}
+
+func (c *Collection[T]) Shift() *T {
```
krishankumar01, line 858: Why return a pointer to a generic type? If the user needs a pointer, they can simply assign T as a pointer type anyway, right? Can we fix similar issue in other methods also?

support/collect/collection.go
```diff
@@ -0,0 +1,1334 @@
...
+	first := c.items[0]
+	c.items = c.items[1:]
+	return &first
+}
+
+func (c *Collection[T]) Shuffle() *Collection[T] {
+	shuffled := make([]T, len(c.items))
+	copy(shuffled, c.items)
+
+	rng := rand.New(rand.NewSource(time.Now().UnixNano()))
```
krishankumar01, line 871: ditto

support/collect/collection.go
```diff
@@ -0,0 +1,1334 @@
...
+
+func (c *Collection[T]) SkipWhile(predicate func(T) bool) *Collection[T] {
+	return c.DropWhile(predicate)
+}
+
+func (c *Collection[T]) Slice(start, length int) *Collection[T] {
+	if start < 0 {
+		start = len(c.items) + start
+	}
+	if start < 0 {
```
krishankumar01, line 896: duplicate check for start<0?

support/collect/collection.go
```diff
@@ -0,0 +1,1334 @@
...
+
+func (c *Collection[T]) SortKeysDesc() *Collection[T] {
+	return c.Clone()
+}
+
+func (c *Collection[T]) Splice(start, deleteCount int, replacement ...T) *Collection[T] {
+	if start < 0 {
+		start = len(c.items) + start
+	}
+	if start < 0 {
```
krishankumar01, line 961: ditto

support/collect/collection.go
```diff
@@ -0,0 +1,1334 @@
...
+	})
+}
+
+func (c *Collection[T]) SortKeys() *Collection[T] {
+	return c.Clone()
+}
+
+func (c *Collection[T]) SortKeysDesc() *Collection[T] {
+	return c.Clone()
+}
```
krishankumar01, line 955: Missing implementation? As they are doing nothing

support/collect/collection.go
```diff
@@ -0,0 +1,1334 @@
...
+	copy(sorted, c.items)
+
+	sort.Slice(sorted, func(i, j int) bool {
+		return less(sorted[i], sorted[j])
+	})
+
+	return &Collection[T]{items: sorted}
+}
+
+func (c *Collection[T]) SortBy(keyFunc func(T) string) *Collection[T] {
```
krishankumar01, line 931: Forcing it to a string may fail in many cases, for example:
```go
func main() {
	type Product struct{ Price int }
	c := Of([]Product{{10}, {2}, {1}})

	y := c.SortBy(func(p Product) string {
		return fmt.Sprintf("%d", p.Price)
	})
	fmt.Println(y) // RESULT: [{1}, {10}, {2}]
}
```

Same issue with other shorting methods.

support/collect/collection.go
```diff
@@ -0,0 +1,1334 @@
...
+		start = len(c.items) + start
+	}
+	if start < 0 {
+		start = 0
+	}
+	if start > len(c.items) {
+		start = len(c.items)
+	}
+
+	end := start + deleteCount
```
krishankumar01, line 968: What if the user accidentally passes a negative deleteCount? It would end up duplicating the elements, right?

support/collect/collection.go
```diff
@@ -0,0 +1,1334 @@
...
+		return "", err
+	}
+	return string(data), nil
+}
+
+func (c *Collection[T]) Transform(fn func(T, int) T) *Collection[T] {
+	for i, item := range c.items {
+		c.items[i] = fn(item, i)
+	}
+	return c
```
krishankumar01, line 1052: Why do we return the original collection in some places and create a new one in others? How about always returning a new collection so the user has fewer surprises? For example, Transform also modifies the original if that's the expected behavior, how about documenting it?

support/collect/collection.go
```diff
@@ -0,0 +1,1334 @@
...
+}
+
+func (c *Collection[T]) Union(other *Collection[T]) *Collection[T] {
+	existing := make(map[string]bool)
+	for _, item := range c.items {
+		existing[fmt.Sprintf("%v", item)] = true
```
krishankumar01, line 1058: This is very performance-heavy. For example, if T is a large User object, checking it this way will be significantly slower. In any case, we should not use the fmt package for key generation. For example:
```go
type BigUser struct {
    ID       int
    Name     string
    Bio      string // Long text
    Metadata map[string]string
}

// IF YOU USE fmt.Sprintf("%v", user):
// 1. Go has to reflectively walk through ID, Name, Bio, and the Map.
// 2. It allocates a massive string for EVERY user.
// 3. It compares massive strings.

// IF YOU USE a proper Key (User.ID):
// 1. Go compares one integer. 
// 2. Zero allocation.
// 3. Instant.
```

So maybe we can accept a keyFunc as well, right? Also, this issue exists in every method that checks for seen and uses fmt.

support/collect/collection.go
```diff
@@ -0,0 +1,1334 @@
...
+				})
+			}
+		}
+		return c
+	default:
+		return c
+	}
+}
+
+func (c *Collection[T]) WhereIn(field string, values []interface{}) *Collection[T] {
```
krishankumar01, line 1172: Using this method will be a lot harder than expected because every time we use it we will need to convert the values to []any which a lot of pain, workaround would be to accept it like `...any`.

support/collect/collection.go
```diff
@@ -0,0 +1,1334 @@
...
+		case "!=":
+			return true
+		default:
+			return false
+		}
+	}
+
+	switch operator {
+	case "=", "==":
+		return reflect.DeepEqual(*fieldValue, value)
```
krishankumar01, line 1270: There is an issue or inconsistent behaviour, for example:
```go
func main() {
	p := Product{Price: 100}
	c := Of([]Product{p})

	// "100" (int) > "50" (string) -> True
	fmt.Println(c.Where("Price", ">", "50").Count()) // Output: 1 (Found)

	// "100" (int) == "100" (string) -> False
	fmt.Println(c.Where("Price", "=", "100").Count()) // Output: 0 (Not Found!)
}
```

krishankumar01:
I recommend against merging methods that are ambiguous, handle edge cases incorrectly, or introduce significant performance bottlenecks. It is better to omit these features than to expose unsafe or unoptimized implementations.

hwbrzzl:
> The implementation and the wide range of methods are solid. Instead of implementing everything in a single PR, should we break it into smaller PRs so that reviewing becomes easier? cc: @hwbrzzl

Yes, we recommend multiple small PRs instead of a big one. It's hard to review it. We can follow this next time.

support/collect/collection.go
hwbrzzl: Yes, it's a bit complicated, but let's keep []any for now, given orm has the same function as well.

image

krishankumar01: But we can optimize here since we already know the issue, right? For example, we accept `...any` in `WhereAny`, `Where`, and all other `Where` methods except `WhereIn` and `WhereNotIn`.

Screenshot 2025-12-02 at 9 54 46 PM

hwbrzzl: `args ...any` is for `= [value]` or `> [value]`, etc. `values []any` is actually for a slice. So basically, they are a bit different. And when I have a slice, eg: `[]string`,  there is no difference between `[]any` and `...any`.

image

`[]string` need to be transformed to `[]any` as well. So I think `[]any` is good for now.

krishankumar01: hmm make sense

PR #1142, opened 2025-07-25
fix: `cc` and `bcc` fields not set when sending mail due to incorrect condition in `SendMail`

krishankumar01:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/737

## ✅ Checks


krishankumar01:
> Good catch, could you also create a PR for v1.15.x?

https://github.com/goravel/framework/pull/1143

PR #1145, opened 2025-07-26
feat: the mail module can use html template directly

krishankumar01:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/260

### Preview 

Screenshot 2025-08-10 at 9 59 50 PM

This PR adds configurable template engine support to the mail module, allowing users to render email templates using Go's `html/template` engine with automatic caching and thread-safe access.

### Features

- **Configurable Template Engine**: Switch between different template engines via configuration
- **Built-in Caching**: Templates are parsed once and cached for improved performance
- **Thread-Safe**: Safe for concurrent use across multiple goroutines
- **Global Registry**: Template engines are cached globally to avoid recreation

#### Configuration
```go
// config/mail.go
"template": map[string]any{
	"default": config.Env("MAIL_TEMPLATE_ENGINE", "html"),
	"engines": map[string]any{
		"html": map[string]any{
			"driver": "html",
			"path":   config.Env("MAIL_VIEWS_PATH", "resources/views/mail"),
		},
	},
}
```

#### Email Templates
Create template files in your views directory:
```html
<!-- resources/views/mail/welcome.html -->
<h1>Welcome {{.Name}}!</h1>
<p>Thank you for joining {{.AppName}}.</p>
```
#### Sending with Templates
```go
// Using fluent API
facades.Mail().
    To([]string{"user@example.com"}).
    Subject("Welcome").
    Content(mail.Content{
        View: "welcome.html",
        With: map[string]any{
            "Name": "Krishan",
            "AppName": "MyApp",
        },
    }).
    Send()
```
#### Custom Template Engines
Register custom engines in configuration:
```go
"template": map[string]any{
	"default": "blade",
	"engines": map[string]any{
		"blade": map[string]any{
			"driver": "custom",
			"via": func() (mail.Template, error) {
				return NewBladeTemplateEngine(), nil
			},
		},
	},
}
```
#### Breaking Changes
Yes - `mail.NewApplication` returns error

## ✅ Checks

### DIFF
--- a/mail/job_test.go
+++ b/mail/job_test.go
@@ -73,7 +73,7 @@ func (r *SendMailJobTestSuite) TestHandle_WrongArgumentTypes() {
 				[]string{"to"}, []string{"cc"}, []string{"bcc"},
 				[]string{"attachments"}, []string{"headers"},
 			},
-			errorMsg: "should be string",
+			errorMsg: "should be of type string",
 		},
 		{
 			name: "html not string",
@@ -82,7 +82,7 @@ func (r *SendMailJobTestSuite) TestHandle_WrongArgumentTypes() {
 				[]string{"to"}, []string{"cc"}, []string{"bcc"},
 				[]string{"attachments"}, []string{"headers"},
 			},
-			errorMsg: "should be string",
+			errorMsg: "should be of type string",
 		},
 		{
 			name: "text not string",
@@ -91,7 +91,7 @@ func (r *SendMailJobTestSuite) TestHandle_WrongArgumentTypes() {
 				[]string{"to"}, []string{"cc"}, []string{"bcc"},
 				[]string{"attachments"}, []string{"headers"},
 			},
-			errorMsg: "should be string",
+			errorMsg: "should be of type string",
 		},
 		{
 			name: "fromAddress not string",
@@ -100,7 +100,7 @@ func (r *SendMailJobTestSuite) TestHandle_WrongArgumentTypes() {
 				[]string{"to"}, []string{"cc"}, []string{"bcc"},
 				[]string{"attachments"}, []string{"headers"},
 			},
-			errorMsg: "should be string",
+			errorMsg: "should be of type string",
 		},
 		{
 			name: "fromName not string",
@@ -109,7 +109,7 @@ func (r *SendMailJobTestSuite) TestHandle_WrongArgumentTypes() {
 				[]string{"to"}, []string{"cc"}, []string{"bcc"},
 				[]string{"attachments"}, []string{"headers"},
 			},
-			errorMsg: "should be string",
+			errorMsg: "should be of type string",
 		},
 		{
 			name: "to not []string",
@@ -118,7 +118,7 @@ func (r *SendMailJobTestSuite) TestHandle_WrongArgumentTypes() {
 				"not-a-slice", []string{"cc"}, []string{"bcc"},
 				[]string{"attachments"}, []string{"headers"},
 			},
-			errorMsg: "should be []string",
+			errorMsg: "should be of type []string",
 		},
 		{
 			name: "cc not []string",
@@ -127,7 +127,7 @@ func (r *SendMailJobTestSuite) TestHandle_WrongArgumentTypes() {
 				[]string{"to"}, "not-a-slice", []string{"bcc"},
 				[]string{"attachments"}, []string{"headers"},
 			},
-			errorMsg: "should be []string",
+			errorMsg: "should be of type []string",
 		},
 		{
 			name: "bcc not []string",
@@ -136,7 +136,7 @@ func (r *SendMailJobTestSuite) TestHandle_WrongArgumentTypes() {
 				[]string{"to"}, []string{"cc"}, "not-a-slice",
 				[]string{"attachments"}, []string{"headers"},
 			},
-			errorMsg: "should be []string",
+			errorMsg: "should be of type []string",
 		},
 		{
 			name: "attachments not []string",
@@ -145,7 +145,7 @@ func (r *SendMailJobTestSuite) TestHandle_WrongArgumentTypes() {
 				[]string{"to"}, []string{"cc"}, []string{"bcc"},
 				"not-a-slice", []string{"headers"},
 			},
-			errorMsg: "should be []string",
+			errorMsg: "should be of type []string",
 		},
 		{
 			name: "headers not []string",
@@ -154,7 +154,7 @@ func (r *SendMailJobTestSuite) TestHandle_WrongArgumentTypes() {
 				[]string{"to"}, []string{"cc"}, []string{"bcc"},
 				[]string{"attachments"}, "not-a-slice",
 			},
-			errorMsg: "should be []string",
+			errorMsg: "should be of type []string",
 		},
 	}
 
--- a/mail/job.go
+++ b/mail/job.go
@@ -24,57 +24,57 @@ func (r *SendMailJob) Signature() string {
 // Handle Execute the job.
 func (r *SendMailJob) Handle(args ...any) error {
 	if len(args) != 10 {
-		return fmt.Errorf("expected 10 arguments, got %d arguments", len(args))
+		return fmt.Errorf("expected 10 arguments, got %d", len(args))
 	}
 
 	subject, ok := args[0].(string)
 	if !ok {
-		return fmt.Errorf("SUBJECT should be string")
+		return fmt.Errorf("SUBJECT should be of type string")
 	}
 
 	html, ok := args[1].(string)
 	if !ok {
-		return fmt.Errorf("HTML body should be string")
+		return fmt.Errorf("HTML BODY should be of type string")
 	}
 
 	text, ok := args[2].(string)
 	if !ok {
-		return fmt.Errorf("TEXT body should be string")
+		return fmt.Errorf("TEXT BODY should be of type string")
 	}
 
 	fromAddress, ok := args[3].(string)
 	if !ok {
-		return fmt.Errorf("FromAddress should be string")
+		return fmt.Errorf("FROM should be of type string")
 	}
 
 	fromName, ok := args[4].(string)
 	if !ok {
-		return fmt.Errorf("FromName should be string")
+		return fmt.Errorf("FROM NAME should be of type string")
 	}
 
 	to, ok := args[5].([]string)
 	if !ok {
-		return fmt.Errorf("TO should be []string")
+		return fmt.Errorf("TO should be of type []string")
 	}
 
 	cc, ok := args[6].([]string)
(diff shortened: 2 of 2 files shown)


mail/application.go
```diff
@@ -55,6 +79,9 @@ func (r *Application) Cc(cc []string) mail.Mail {
 func (r *Application) Content(content mail.Content) mail.Mail {
 	instance := r.instance()
 	instance.html = content.Html
+	instance.viewPath = content.View
+	instance.textPath = content.Text
```
hwbrzzl: Why is `textPath`, not `text`?

mail/application.go
```diff
@@ -8,26 +8,50 @@ import (
...
 	subject     string
 	attachments []string
 	bcc         []string
 	cc          []string
 	to          []string
 	clone       int
+
+	viewPath string
+	textPath string
+	with     map[string]any
```
hwbrzzl: Can `Params` be used here? To replace several parameters.

mail/application.go
hwbrzzl: Can these three parameters be set simultaneously?

mail/application.go
```diff
@@ -78,42 +105,27 @@ func (r *Application) Queue(mailable ...mail.Mailable) error {
...
-			Value: r.bcc,
-		},
-		{
-			Type:  "[]string",
-			Value: r.attachments,
-		},
-		{
-			Type:  "[]string",
-			Value: convertMapHeadersToSlice(r.headers),
+			Type:  "mail.Params",
```
hwbrzzl: Don't support such type, you can check the list: https://www.goravel.dev/digging-deeper/queues.html#queue-arg-type-supported-types

mail/application.go
```diff
@@ -195,31 +232,58 @@ func (r *Application) setUsingMailable(mailable mail.Mailable) {
...
+		}
+		r.html = renderedHtml
+	}
+
+	if r.textPath != "" && r.template != nil {
+		renderedText, err := r.template.Render(r.textPath, r.with)
+		if err != nil {
+			return err
+		}
+		r.text = renderedText
```
hwbrzzl: Can they be used simultaneously?

mail/application.go
```diff
@@ -195,31 +232,58 @@ func (r *Application) setUsingMailable(mailable mail.Mailable) {
...
+}
+
+func SendMail(config config.Config, params Params) error {
 	e := NewEmail()
+	fromAddress, fromName := params.FromAddress, params.FromName
 	if fromAddress == "" {
-		e.From = fmt.Sprintf("%s <%s>", config.GetString("mail.from.name"), config.GetString("mail.from.address"))
-	} else {
-		e.From = fmt.Sprintf("%s <%s>", fromName, fromAddress)
+		fromName, fromAddress = config.GetString("mail.from.name"), config.GetString("mail.from.address")
```
hwbrzzl: Keep the same order as above:

```suggestion
		fromAddress, fromName, = config.GetString("mail.from.address"), config.GetString("mail.from.name")
```

mail/setup/config/mail.go
```diff
@@ -39,5 +39,25 @@ func init() {
...
+		//
+		// This controls template rendering for email views. The default engine uses
+		// Go's html/template which can render both HTML and text templates safely
+		// with automatic caching for better performance. Each mail application instance
+		// gets its own template engine with thread-safe caching using sync.Map.
+		"template": map[string]any{
+			"driver":     config.Env("MAIL_TEMPLATE_DRIVER", "default"),
+			"views_path": config.Env("MAIL_VIEWS_PATH", "resources/views/mail"),
+			// Custom template engines can be registered here
+			"drivers": map[string]any{
```
hwbrzzl: Keep the same as other modules (cache, filesystem)

```suggestion
			"default":     config.Env("MAIL_TEMPLATE_ENGINE", "html"),
			"views_path": config.Env("MAIL_VIEWS_PATH", "resources/views/mail"),
			// Custom template engines can be registered here
			"engines": map[string]any{
						"html": map[string]any{
									"driver": "html",
									"path": config.Env("MAIL_VIEWS_PATH", "resources/views/mail"),
						},
						"custom": map[string]any{
									"driver": "custom",
									"via": func() (mail.Template, error) {
												return NewBladeTemplateEngine(), nil
									}
						}
```

mail/template/default.go
hwbrzzl, line 1: Rename to `html.go`

mail/template/default.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"fmt"
+	"html/template"
+	"path/filepath"
+	"strings"
+	"sync"
+)
+
+type DefaultEngine struct {
```
hwbrzzl:
```suggestion
type Html struct {
```

mail/template/default.go
```diff
@@ -0,0 +1,49 @@
...
+func (r *DefaultEngine) Render(path string, data any) (string, error) {
+	templatePath := filepath.Join(r.viewsPath, path)
+	tmpl, err := r.getTemplate(templatePath)
+	if err != nil {
+		return "", err
+	}
+
+	var buf strings.Builder
+	if err := tmpl.Execute(&buf, data); err != nil {
+		return "", fmt.Errorf("failed to execute template %s: %w", path, err)
```
hwbrzzl: Add the error to the error list.

mail/template/registry.go
```diff
@@ -0,0 +1,57 @@
+package template
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/goravel/framework/contracts/config"
+	mailcontract "github.com/goravel/framework/contracts/mail"
```
hwbrzzl:
```suggestion
	contractmail "github.com/goravel/framework/contracts/mail"
```

mail/template/registry.go
```diff
@@ -0,0 +1,57 @@
...
+		key := fmt.Sprintf("mail.template.drivers.%s.engine", driver)
+		engineConfig := config.Get(key)
+
+		switch v := engineConfig.(type) {
+		case mailcontract.Template:
+			return v, nil
+		case func() (mailcontract.Template, error):
+			engine, err := v()
+			if err != nil {
+				return nil, fmt.Errorf("factory for template engine '%s' failed: %w", driver, err)
```
hwbrzzl: DItto

mail/template/registry.go
```diff
@@ -0,0 +1,57 @@
...
+			}
+			return engine, nil
+		default:
+			return nil, fmt.Errorf("unsupported or misconfigured template engine: %s", driver)
```
hwbrzzl: Ditto

mail/template/registry_test.go
```diff
@@ -0,0 +1,105 @@
+package template
+
+import (
+	"fmt"
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	mailcontract "github.com/goravel/framework/contracts/mail"
```
hwbrzzl:
```suggestion
	contractsmail "github.com/goravel/framework/contracts/mail"
```

mail/template/registry_test.go
```diff
@@ -0,0 +1,105 @@
...
+	mocksconfig "github.com/goravel/framework/mocks/config"
+	mocksmail "github.com/goravel/framework/mocks/mail"
+)
+
+func TestGet_DefaultEngine(t *testing.T) {
+	engines = sync.Map{}
+
+	mockConfig := mocksconfig.NewConfig(t)
+	mockConfig.EXPECT().GetString("mail.template.driver", "default").Return("default")
```
hwbrzzl:
```suggestion
	mockConfig.EXPECT().GetString("mail.template.driver", "default").Return("default").Once()
```

hwbrzzl (commented):
Awesome. Please add the sent mail screenshots in the description.

mail/application.go
krishankumar01: We also have another field named `text`, which stores the rendered text content from the `textPath` file.

krishankumar01: Yes, because typically the `text` version of an email is used as a fallback for the `HTML` version.
(About `Html` param: if you provide `viewPath`, it will override the `Html` parameter.)

Screenshot 2025-08-10 at 9 45 06 PM

mail/application.go
krishankumar01: Yes, we can also use goroutines if the user has provided both templates, to speed up rendering.

mail/application.go
krishankumar01: Should we add a custom type, `mail.Params`, specifically for our use case in parsing?

mail/application.go
krishankumar01: Good idea, let me check it.

mail/application.go
hwbrzzl: No, queue doesn't support custom type for now.

mail/setup/config/mail.go
```diff
@@ -39,5 +39,26 @@ func init() {
 		"username": config.Env("MAIL_USERNAME"),
 
 		"password": config.Env("MAIL_PASSWORD"),
+
+		// Template Configuration
+		//
+		// This controls template rendering for email views. Template engines are cached
+		// globally and support both built-in drivers and custom implementations via factories.
```
hwbrzzl, line 46:
```suggestion
		// globally and support both built-in drivers and custom implementations via factories.
		// Available Drivers: "html", "custom"
```

hwbrzzl (commented):
Thanks, great 👍 Could you update the description according to the latest logic?

krishankumar01:
> Thanks, great 👍 Could you update the description according to the latest logic?

I have already updated the description

mail/job_test.go
hwbrzzl, line 1: Great 👍

mail/template/html.go
```diff
@@ -45,6 +45,6 @@ func (r *Html) getTemplate(templatePath string) (*template.Template, error) {
 		return nil, errors.MailTemplateParseFailed.Args(templatePath, err)
 	}
 
-	r.cache.LoadOrStore(templatePath, tmpl)
-	return tmpl, nil
+	actual, _ := r.cache.LoadOrStore(templatePath, tmpl)
```
hwbrzzl, line 48: Please judge the error here.

hwbrzzl (commented):
Perfect 👍 

image

mail/template/html.go
krishankumar01: The second parameter is not an error, but a boolean indicating whether the template was loaded successfully. Here, we are storing a new template, so the `actual` value always presents.

> // LoadOrStore returns the existing value for the key if present.
// Otherwise, it stores and returns the given value.
// The loaded result is true if the value was loaded, false if stored.

hwbrzzl: Thanks for the explanation.

PR #1146, opened 2025-07-27
fix: `SendMailJob` returns incorrect errors

kkumar-gcc:
## 📑 Description

## ✅ Checks

### DIFF
--- a/mail/job.go
+++ b/mail/job.go
@@ -27,39 +27,39 @@ func (r *SendMailJob) Handle(args ...any) error {
 		return fmt.Errorf("expected 9 arguments, got %d", len(args))
 	}
 
-	from, ok := args[0].(string)
-	if !ok {
-		return fmt.Errorf("FROM should be of type string")
-	}
-
-	subject, ok := args[1].(string)
+	subject, ok := args[0].(string)
 	if !ok {
 		return fmt.Errorf("SUBJECT should be of type string")
 	}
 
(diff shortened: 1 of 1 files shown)


mail/job.go
```diff
@@ -72,5 +72,5 @@ func (r *SendMailJob) Handle(args ...any) error {
 		return fmt.Errorf("HEADERS should be of type []string")
 	}
 
-	return SendMail(r.config, from, subject, body, recipient, cc, bcc, replyTo, attachments, convertSliceHeadersToMap(headers))
+	return SendMail(r.config, subject, body, fromAddress, fromName, to, cc, bcc, attachments, convertSliceHeadersToMap(headers))
```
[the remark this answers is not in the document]
krishankumar01: We extract 9 arguments from args, and config is passed separately.

hwbrzzl, line 75: It may be better to pass a struct here.

kkumar-gcc: Yes, we can do that in v1.17.x, but for existing ones, I guess it's fine.

krishankumar01: Yeah, I'll create a separate PR for these optimizations.

hwbrzzl: Thanks, only for master should be fine.

krishankumar01: Yes, it’s a breaking change in some sense, at least since `SendMail` is an exposed method(not documented).

PR #1153, opened 2025-08-09
chore: optimize package stubs with config file overwrite and remove steps

almas-x:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/743

## ✅ Checks

### DIFF
--- a/foundation/console/package_make_command_stubs.go
+++ b/foundation/console/package_make_command_stubs.go
@@ -144,20 +144,28 @@ import (
 	"github.com/goravel/framework/packages"
 	"github.com/goravel/framework/packages/match"
 	"github.com/goravel/framework/packages/modify"
+	"github.com/goravel/framework/support/file"
 	"github.com/goravel/framework/support/path"
 )
 
 func main() {
+	config, err := file.GetPackageContent(packages.GetModulePath(), "setup/config/DummyName.go")
+	if err != nil {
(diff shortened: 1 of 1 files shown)


foundation/console/package_make_command_stubs.go
```diff
@@ -144,20 +144,28 @@ import (
...
+	if err != nil {
+		panic(err)
+	}
+
 	packages.Setup(os.Args).
 		Install(
 			modify.GoFile(path.Config("app.go")).
 				Find(match.Imports()).Modify(modify.AddImport(packages.GetModulePath())).
 				Find(match.Providers()).Modify(modify.Register("&DummyName.ServiceProvider{}")),
+			modify.File(path.Config("DummyName.go")).Overwrite(config),
```
hwbrzzl, line 162: Not sure if it's necessary to add this. The package configuration may not be required. The package author may want user to publish the configuration manually?

image

almas-x: I saw that you moved the location of the configuration file generated during make package from the package directory to the setup directory, and I mistakenly thought you wanted to generate it during installation — so instead of publishing `config/sms.go`, it would now be `setup/config/sms.go`,Is that right?

hwbrzzl: Correct

PR #1168, opened 2025-08-22
chore: add go tool

hwbrzzl:
## ✅ Checks


support/file/file_test.go
```diff
@@ -75,36 +75,6 @@ func TestGetContent(t *testing.T) {
 	assert.Contains(t, content, "Version")
 }
 
-func TestGetFrameworkContent(t *testing.T) {
```
hwbrzzl, line 78: The `GetFrameworkContent` function is not used actually for now, we can remove it if Go v1.24 doesn't support testing it directly. We can add the tests in goravel/example when implementing https://github.com/goravel/goravel/issues/719

PR #1171, opened 2025-08-23
feat: Add Process Management Support

krishankumar01:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/567

This PR implements a single process invocation. In the next couple of PRs, I will implement more advanced features like `Pipe` and `Pool`. Currently, we can directly use `process.New()`. In upcoming PRs, I will also add a process facade.

> Note: Currently, TTY is adding a pipe to `os.Stdin`, but in the next PR I will add a fully functional TTY using [github.com/creack/pty](https://github.com/creack/pty). This will only be available on Unix-like systems.

### Run a command and get output
```go
package main

import (
    "context"
    "fmt"

    "github.com/goravel/framework/process"
)

func main() {
    res, err := process.New().
        Quietly().
        Run("sh", "-c", "printf 'hello'")
    if err != nil { panic(err) }

    fmt.Println(res.Successful()) // true
    fmt.Println(res.Output())     // "hello"
}
```

### Start and control a running process
```go
ctx := context.Background()

r, err := process.New().
    WithContext(ctx).
    Start("sh", "-c", "printf 'hello'")
if err != nil { panic(err) }

_ = r.Stop(500 * time.Millisecond) // try graceful stop (cross‑platform)
res := r.Wait()                     // block until exit
fmt.Println(res.ExitCode())
```

### Common options (chainable)
```go
res, err := process.New().
    Env(map[string]string{"FOO": "BAR"}).     // add environment vars
    Path("/tmp").                               // working directory
    Input(strings.NewReader("stdin data")).     // provide stdin
    Timeout(2 * time.Second).                    // kill after timeout
    Quietly().                                   // don't mirror to current stdout/stderr
    OnOutput(func(typ process.OutputType, b []byte) {
        // per-line callback for stdout/stderr
    }).
    WithContext(context.Background()).
    Run("sh", "-c", "echo $FOO && pwd")
```

### Notes
- Output is mirrored to current stdout/stderr unless `Quietly()` is set.
- `Start` returns a handle you can `Wait()` on and `Stop()` with a timeout.

## ✅ Checks

### DIFF
--- a/process/process_windows_test.go
+++ b/process/process_windows_test.go
@@ -40,7 +40,7 @@ func TestProcess_Run_Windows(t *testing.T) {
 				// powershell: write-error writes to stderr and returns non-zero
 				p.Quietly()
 			},
-			expectOK: false,
+			expectOK: true, // Run doesn't error on non-zero exit
 			check: func(t *testing.T, res *Result) {
 				assert.Contains(t, res.ErrorOutput(), "bad")
 				assert.False(t, res.Successful())
@@ -78,7 +78,7 @@ func TestProcess_Run_Windows(t *testing.T) {
 			setup: func(p *Process) {
 				p.Timeout(200 * time.Millisecond).Quietly()
 			},
-			expectOK: false,
+			expectOK: true, // Run doesn't error on timeout
 			check: func(t *testing.T, res *Result) {
 				assert.False(t, res.Successful())
 				assert.NotEqual(t, 0, res.ExitCode())


process/process_unix.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"os/exec"
+	"syscall"
+)
+
+// setSysProcAttr sets POSIX-specific process attributes.
+func setSysProcAttr(cmd *exec.Cmd) {
+	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
+}
```
krishankumar01, line 13: Currently, we have only one method, but there are multiple instances where we will need OS-specific features like TTY.

process/process_windows_test.go
```diff
@@ -0,0 +1,104 @@
+//go:build windows
```
krishankumar01, line 1: We need to write different test cases for Windows, so instead of relying on runtime.GOOS, we are going to rely on build tags.

process/running_unix.go
```diff
@@ -0,0 +1,88 @@
...
+//	result.Wait()
+//	<-done // Wait for monitoring goroutine to finish.
+func (r *Running) Running() bool {
+	if r.cmd == nil || r.cmd.Process == nil {
+		return false
+	}
+
+	// Actively probe the OS by sending a null signal.
+	// A nil error means the OS found the process (running or zombie).
+	err := r.cmd.Process.Signal(unix.Signal(0))
```
krishankumar01, line 43: Windows does not fully support POSIX signals. As suggested, instead of using `syscall`, we use `golang.org/x/sys/unix` on Unix-like systems and rely on build tags to achieve cross-platform functionality.

contracts/process/process.go
```diff
@@ -0,0 +1,29 @@
...
+type OnOutputFunc func(typ OutputType, line []byte)
+
+type Process interface {
+	Command(name string, arg ...string) Process
+	Env(vars map[string]string) Process
+	Input(in io.Reader) Process
+	Path(path string) Process
+	Quietly() Process
+	OnOutput(handler OnOutputFunc) Process
+	Run(ctx context.Context) (Result, error)
```
hwbrzzl: How about adding a `WithContext` function?

contracts/process/process.go
```diff
@@ -0,0 +1,29 @@
...
+	OutputTypeStdout OutputType = iota
+	OutputTypeStderr
+)
+
+type OnOutputFunc func(typ OutputType, line []byte)
+
+type Process interface {
+	Command(name string, arg ...string) Process
+	Env(vars map[string]string) Process
+	Input(in io.Reader) Process
```
hwbrzzl, line 36: `in io.Reader` vs `string`, which is better?

contracts/process/process.go
```diff
@@ -0,0 +1,29 @@
...
+type OutputType int
+
+const (
+	OutputTypeStdout OutputType = iota
+	OutputTypeStderr
+)
+
+type OnOutputFunc func(typ OutputType, line []byte)
+
+type Process interface {
```
hwbrzzl, line 29: Could you add descriptions for the functions?

contracts/process/result.go
```diff
@@ -0,0 +1,12 @@
+package process
+
+type Result interface {
```
hwbrzzl, line 6: Could you add descriptions for the functions?

contracts/process/running.go
```diff
@@ -0,0 +1,18 @@
+package process
+
+import (
+	"os"
+	"time"
+)
+
+type Running interface {
```
hwbrzzl, line 11: Could you add descriptions for the functions?

process/process_unix_test.go
```diff
@@ -0,0 +1,140 @@
...
+	"bytes"
+	"context"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	contractsprocess "github.com/goravel/framework/contracts/process"
+	"github.com/stretchr/testify/assert"
```
hwbrzzl, line 13: Sort it

process/process_unix_test.go
```diff
@@ -0,0 +1,140 @@
...
+	tests := []struct {
+		name     string
+		setup    func(p *Process)
+		expectOK bool
+		check    func(t *testing.T, res *Result)
+	}{
+		{
+			name: "echo to stdout",
+			setup: func(p *Process) {
+				p.Command("sh", "-c", "printf 'hello'").Quietly()
```
hwbrzzl: Can this be `p.Command("sh -c printf 'hello'").Quietly()`?

process/process_unix_test.go
```diff
@@ -0,0 +1,140 @@
...
+			res, err := p.Run(context.Background())
+			if tt.expectOK {
+				assert.NoError(t, err)
+			}
+			assert.NotNil(t, res)
+			r, ok := res.(*Result)
+			assert.True(t, ok, "unexpected result type")
+			if ok && tt.check != nil {
+				tt.check(t, r)
+			}
```
hwbrzzl:
```suggestion
			tt.check(t, r)
```

process/process_unix_test.go
```diff
@@ -0,0 +1,140 @@
...
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			p := New()
+			tt.setup(p)
+			res, err := p.Run(context.Background())
+			if tt.expectOK {
+				assert.NoError(t, err)
+			}
```
hwbrzzl:
```suggestion
			assert.Equal(t, tt.expectOK, err == nil)
```

process/process_windows_test.go
```diff
@@ -0,0 +1,104 @@
...
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			p := New()
+			if tt.setup != nil {
+				tt.setup(p)
+			}
```
hwbrzzl:
```suggestion
			tt.setup(p)
```

process/process_windows_test.go
```diff
@@ -0,0 +1,104 @@
...
+				assert.NoError(t, err)
+			}
+			assert.NotNil(t, res)
+			if tt.check != nil {
+				r, ok := res.(*Result)
+				assert.True(t, ok, "unexpected result type")
+				if ok {
+					tt.check(t, r)
+				}
+			}
```
hwbrzzl: Ditto

hwbrzzl (commented):
Amazing 👍

contracts/process/process.go
almas-x: Personally, I think using `io.Reader` as input is more flexible than using a plain string. It allows the process to handle various data sources efficiently—such as files, network streams, or buffers—without needing to load everything into memory. This makes the code more versatile and scalable.

krishankumar01: `io.Reader` is better because it’s the standard way in Go to handle input from different sources like files, terminal, or network. If you only use `string`, the function becomes less flexible. When you already have a string, you can easily turn it into an `io.Reader` with `strings.NewReader()`.

hwbrzzl: Agree, can we add some quick functions to build different `io.Reader`?

process/process_unix_test.go
krishankumar01: No, `exec.CommandContext` takes the command name and its arguments separately. I also don’t recommend writing a wrapper just to handle this.

contracts/process/process.go
krishankumar01: @hwbrzzl Each stdlib type has its own way to get an `io.Reader`, so we don’t need to handle it:
  - `string` → `strings.NewReader("hello")`
 - `[]byte` → `bytes.NewReader(data)`
 - `*os.File` → already implements `io.Reader`

process/process_unix_test.go
hwbrzzl: Actually, as a user, why he needs to care about the `exec.CommandContext` command, he just wants to run a command like: `facades.Process.Run("sh -c printf 'hello'")`. And is the `Command` function necessary?

contracts/process/process.go
hwbrzzl: Okay, the quick functions is optional for now.

process/process_unix_test.go
krishankumar01: @hwbrzzl Process is a generic implementation, not tied to the Goravel framework itself. `Run` will just be wrapper functions around process.New(), so the user won’t need to call `Command` directly.

Example:
```go
func Run(name string, args ...string) process.Result {
    return New().Command(name, args...).Run()
}
```

Usage: 
```go
facades.Process().Run("sh", "-c", "printf 'hello'")
```
We should still respect the `exec.CommandContext` syntax because Go does not split arguments like a shell. The command name and each argument must be passed separately. This keeps it consistent with the standard library and avoids unexpected bugs when handling arguments.

hwbrzzl: Sorry, I mean `facades.Process().Run("sh -c printf 'hello'")`, instead of `facades.Process().Command("sh", "-c", "printf 'hello'").Run()`.

krishankumar01: If the concern is with syntax, we can allow `Run` to accept the command directly. Earlier we accepted context, but if we switch to using `WithContext`, it will look like this:
```go
facades.Process().Quietly().Path(...).WithContext(ctx).Run("sh", "-c", "printf 'hello'")
facades.Process().Quietly().Path(...).WithContext(ctx).Start("sh", "-c", "printf 'hello'")
```
The arguments are still passed separately, since that is handled by the Go standard library.

almas-x: In fact, Go's `exec.Command` does not split or parse a command string into arguments. It requires the executable name and each argument to be passed separately. If you provide the entire command as a single string, exec.Command will treat it as the executable name, which won't work as expected. Therefore, you still need to split the command into arguments before passing them to `exec.Command`. It's better if `process.New().Command` directly accepts the already split arguments, as this aligns with Go's usage conventions.

hwbrzzl: Yes, splitting the command is the standard action for the Go library. However, I think it would be better to support these two ways: `Run("sh -c printf 'hello'")` and `Run("sh", "-c", "printf 'hello'")`. I was sick of splitting the command myself when I implemented other features previously.

almas-x: Take the example command `"sh -c printf 'hello'"`. How can you split it into `"sh", "-c", "printf 'hello'"`? If you simply split by spaces, you'll get` "sh", "-c", "printf"`, and `"'hello'"`, which is not correct. 🤔

hwbrzzl: Good point, is it good to judge `sh` specifically? Anyway, it's optional for this PR, we can optimize it in the future.

krishankumar01: I believe that if we provide a feature, it should work in all cases; it shouldn't work only conditionally unless there's a specific limitation.(We can provide a different package to convert a single string to {name, args...})

hwbrzzl: Yes, the key is to transform the command, there should be a rule to do that.

hwbrzzl (approved):
LGTM, awesome, thanks!

process/process.go
```diff
@@ -0,0 +1,144 @@
...
+func (r *Process) WithContext(ctx context.Context) contractsprocess.Process {
+	if ctx == nil {
+		ctx = context.Background()
+	}
+
+	r.ctx = ctx
+	return r
+}
+
+func (r *Process) run(name string, args ...string) (contractsprocess.Result, error) {
```
hwbrzzl, line 90: @kkumar-gcc I found there is an `Error` function in `contractsprocess.Result`, and an error is returned here as well. It's a bit complicated to judge these two errors. Can the `Result.Error()` be returned here directly instead of `nil`?

process/running.go
```diff
@@ -0,0 +1,114 @@
...
+
+func NewRunning(cmd *exec.Cmd, stdout, stderr *bytes.Buffer) *Running {
+	return &Running{
+		cmd:          cmd,
+		stdoutBuffer: stdout,
+		stderrBuffer: stderr,
+	}
+}
+
+func (r *Running) Wait() contractsprocess.Result {
```
hwbrzzl, line 31: How about returning an error here instead of `result.Error()` in other places?


```suggestion
func (r *Running) Wait() (contractsprocess.Result, error) {
```

process/process.go
krishankumar01: Yes, both serve different purposes. The first error indicates that something went wrong before the process even started for example, an invalid configuration or a setup issue.
On the other hand, the error inside Result means the process did start, but the command itself failed during execution.

process/running.go
krishankumar01: Well, the error was added since you requested access to the internal error in `Result`. We can go with this approach too. I’ll create a PR to fix this.

PR #1175, opened 2025-08-25
feat: refactor build command to support interactive and custom OS/arch selection

almas-x:
## 📑 Description

This update refactors the build command to provide an interactive selection for OS and architecture, limited to commonly used options for convenience. However, users can still specify any OS and architecture supported by Go via command-line arguments. The validation of os and arch values is now delegated to the Go compiler, ensuring flexibility while simplifying the command logic. This makes the build process user-friendly for most cases, while retaining full compatibility with all platforms supported by Go.

## ✅ Checks

### DIFF
--- a/console/console/build_command.go
+++ b/console/console/build_command.go
@@ -3,11 +3,12 @@ package console
 import (
 	"fmt"
 	"os/exec"
-	"slices"
+	"runtime"
 
 	"github.com/goravel/framework/contracts/config"
 	"github.com/goravel/framework/contracts/console"
 	"github.com/goravel/framework/contracts/console/command"
+	supportconsole "github.com/goravel/framework/support/console"
 )
 
 type BuildCommand struct {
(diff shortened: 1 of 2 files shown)


console/console/build_command.go
```diff
@@ -34,22 +35,26 @@ func (r *BuildCommand) Description() string {
 func (r *BuildCommand) Extend() command.Extend {
 	return command.Extend{
 		Flags: []command.Flag{
+			&command.StringFlag{
+				Name:    "arch",
```
hwbrzzl, line 39: Can we set a default value here? `amd64`?

console/console/build_command.go
```diff
@@ -72,53 +77,52 @@ func (r *BuildCommand) Handle(ctx console.Context) error {
...
-	validOs := []string{"linux", "windows", "darwin"}
-	if !slices.Contains(validOs, os) {
-		ctx.Error(fmt.Sprintf("Invalid os '%s' specified. Allowed values are: %v", os, validOs))
-		return nil
+	arch := ctx.Option("arch")
+	if arch == "" {
+		if arch, err = ctx.Choice("Select target architecture", []console.Choice{
+			{Key: "amd64", Value: "amd64"},
+			{Key: "arm64", Value: "arm64"},
+		}, console.ChoiceOption{Default: runtime.GOARCH}); err != nil {
```
almas-x: The default value is set to `runtime.GOARCH` here. @hwbrzzl

console/console/build_command.go
almas-x: Setting default values here will invalidate the interactive choice for the target architecture later on.

console/console/build_command.go
hwbrzzl: Good point. I have a concern that I think many people don't know the difference between amd64, arm64 and runtime.GOARCH. A default value is suitable for a lot of people, and the command supports setting arch if needed. arch is optional, Users don't need to choose it. If we provide the choice, many users will have a question: What should I choose? It's a bit inconvenient.

almas-x: > Good point. I have a concern that I think many people don't know the difference between amd64, arm64 and runtime.GOARCH. A default value is suitable for a lot of people, and the command supports setting arch if needed. arch is optional, Users don't need to choose it. If we provide the choice, many users will have a question: What should I choose? It's a bit inconvenient.

Sounds sensible.

PR #447, opened 2024-03-25
feat: add OrWhereNull method

Kamandlou:
## 📑 Description

add OrWhereNull method

## ✅ Checks

- [ ] My pull request adheres to the code style of this project
- [ ] My code requires changes to the documentation
- [ ] I have updated the documentation as required
- [ ] All the tests have passed

PR #1176, opened 2025-08-26
feat: [#447] install/uninstall Auth facade

hwbrzzl:
## 📑 Description

Relates to https://github.com/goravel/goravel/issues/447

This pull request refactors the setup and uninstall logic for the authentication package and improves the handling and testing of facade dependencies. The main changes include moving configuration stubs to a dedicated struct, updating uninstall logic to prevent removal of facades that are depended on by others, and enhancing the test suite to cover these scenarios.

image

## ✅ Checks


auth/setup/setup.go
```diff
@@ -10,24 +10,20 @@ import (
 )
 
 func main() {
-	// TODO: optimize this in https://github.com/goravel/goravel/issues/447
-	// config, err := supportfile.GetFrameworkContent("auth/setup/config/auth.go")
```
hwbrzzl, line 14: `GetFrameworkContent` is not a good choice, it's unstable.

foundation/application_test.go
```diff
@@ -152,304 +120,6 @@ func (s *ApplicationTestSuite) TestAddPublishGroup() {
 	s.Equal("config1.go", s.app.publishGroups["public"]["config.go"])
 }
 
-func (s *ApplicationTestSuite) TestMakeArtisan() {
```
hwbrzzl, line 155: Avoid cycle dependency.

auth/service_provider.go
```diff
@@ -32,8 +32,8 @@ func (r *ServiceProvider) Relationship() binding.Relationship {
 			binding.Gate,
 		},
 		Dependencies: collect.Unique(
-			binding.Facades[binding.Auth].Dependencies,
-			binding.Facades[binding.Gate].Dependencies,
+			binding.Bindings[binding.Auth].Dependencies,
```
almas-x, line 35: How about renaming `binding.Bindings `to something like `binding.Builtin` to avoid redundancy?

hwbrzzl: em, I think `Builtin` is a bit meaningless.

almas-x: The term `Builtin` may seem meaningless on its own, but when combined as `binding.Builtin`, it clearly indicates that these are the framework’s built-in bindings. By the way `binding.Bindings` is also fine.

almas-x (approved):
LGTM 👍

PR #1178, opened 2025-08-30
feat: Add `Pipe` support in process package

kkumar-gcc:
## 📑 Description

## ✅ Checks

### DIFF
--- a/process/process_unix_test.go
+++ b/process/process_unix_test.go
@@ -99,6 +99,21 @@ func TestProcess_Run_Unix(t *testing.T) {
 				assert.NotEqual(t, 0, res.ExitCode())
 			},
 		},
+		{
+			name: "disable buffering with OnOutput",
+			args: []string{"sh", "-c", "printf 'to_stdout'; printf 'to_stderr' 1>&2"},
+			setup: func(p *Process) {
+				p.DisableBuffering().Quietly().OnOutput(func(typ contractsprocess.OutputType, line []byte) {
+					// The handler still works, but the Result buffer is what we're testing.
+				})
+			},
+			expectOK: true,
+			check: func(t *testing.T, res *Result) {
+				assert.Equal(t, "", res.Output())
+				assert.Equal(t, "", res.ErrorOutput())
+				assert.True(t, res.Successful())
+			},
+		},
 	}
 
 	for _, tt := range tests {
--- a/process/running.go
+++ b/process/running.go
@@ -14,28 +14,41 @@ type Running struct {
 	cmd          *exec.Cmd
 	stdoutBuffer *bytes.Buffer
 	stderrBuffer *bytes.Buffer
+	doneChan     chan struct{}
 
-	once     sync.Once
 	result   contractsprocess.Result
 	resultMu sync.RWMutex
 }
 
 func NewRunning(cmd *exec.Cmd, stdout, stderr *bytes.Buffer) *Running {
-	return &Running{
+	running := &Running{
 		cmd:          cmd,
 		stdoutBuffer: stdout,
 		stderrBuffer: stderr,
+		doneChan:     make(chan struct{}),
 	}
+
+	go func() {
+		waitErr := running.cmd.Wait()
+
+		res := buildResult(running, waitErr)
+		running.resultMu.Lock()
+		running.result = res
+		running.resultMu.Unlock()
+
+		close(running.doneChan)
+	}()
+
+	return running
+}
+
+func (r *Running) Done() <-chan struct{} {
+	return r.doneChan
 }
 
 func (r *Running) Wait() contractsprocess.Result {
-	r.once.Do(func() {
-		err := r.cmd.Wait()
-		res := buildResult(r, err)
-		r.resultMu.Lock()
-		r.result = res
-		r.resultMu.Unlock()
-	})
+	<-r.doneChan
+
 	r.resultMu.RLock()
 	defer r.resultMu.RUnlock()
 	return r.result
--- a/contracts/process/process.go
+++ b/contracts/process/process.go
@@ -27,6 +27,19 @@ type OnOutputFunc func(typ OutputType, line []byte)
 // Implementations are mutable and should not be reused concurrently.
 // Each method modifies the same underlying process configuration.
 type Process interface {
+	// DisableBuffering prevents the process's stdout and stderr from being buffered
+	// in memory. This is a critical optimization for commands that produce a large
+	// volume of output, especially when that output is already being handled by
+	// an OnOutput streaming callback.
+	//
+	// CONSEQUENCE: As output is not captured, the following methods on the
+	// Running and Result handles will always return an empty string:
+	//   - Running.Output()
+	//   - Running.ErrorOutput()
+	//   - Result.Output()
+	//   - Result.ErrorOutput()
+	DisableBuffering() Process
+
 	// Env adds or overrides environment variables for the process.
 	// Modifies the current process configuration.
 	Env(vars map[string]string) Process
--- a/contracts/process/running.go
+++ b/contracts/process/running.go
@@ -5,45 +5,60 @@ import (
 	"time"
 )
 
-// Running represents a handle to a process that has been started and is still active.
-// It provides methods for inspecting process state, retrieving output, and controlling
-// its lifecycle.
+// Running represents a handle to a single, active process.
+// Its primary role is to manage the lifecycle and inspect the state of the process.
 type Running interface {
 	// PID returns the operating system process ID.
 	PID() int
 
 	// Running reports whether the process still exists according to the OS.
-	//
-	// NOTE: This may return true for a "zombie" process (terminated but not
-	// reaped). You must eventually call Wait() to reap the process and release
-	// resources. A common pattern is to poll Running() in a goroutine and then
-	// call Wait() in the main flow to ensure cleanup.
+	// NOTE: This may return true for a "zombie" process (one that has terminated
+	// but has not been reaped). Wait() must be called to reap the process.
 	Running() bool
 
+	// Done returns a read-only channel that is closed once the process has exited.
+	//
+	// This provides an efficient, non-polling mechanism to wait for process completion,
+	// typically in a select statement.
+	//
+	// After receiving a signal from this channel, the caller is still required to
+	// call Wait() to retrieve the process's final Result and to release all
+	// underlying system resources. Failure to do so will result in a resource leak.
+	Done() <-chan struct{}
+
 	// Output returns the complete stdout captured from the process so far.
+	//
+	// WARNING: This method buffers the entire output in memory. For processes that
+	// may generate a large volume of output, use the Process.OnOutput() callback during
+	// configuration to stream the data instead.
 	Output() string
 
 	// ErrorOutput returns the complete stderr captured from the process so far.
+	//
+	// WARNING: This method buffers the entire output in memory. For processes that
+	// may generate a large volume of output, use the Process.OnOutput() callback during
+	// configuration to stream the data instead.
 	ErrorOutput() string
 
 	// LatestOutput returns the most recent chunk of stdout produced by the process.
-	// The definition of "latest" is implementation dependent.
+	// TODO: Remove in a subsequent PR. The OnOutput callback is the superior, idiomatic
+	// pattern for handling live streaming output, making this method redundant.
 	LatestOutput() string
 
 	// LatestErrorOutput returns the most recent chunk of stderr produced by the process.
-	// The definition of "latest" is implementation dependent.
+	// TODO: Remove in a subsequent PR. The OnOutput callback is the superior, idiomatic
+	// pattern for handling live streaming error output, making this method redundant.
 	LatestErrorOutput() string
 
 	// Wait blocks until the process exits and returns its final Result.
-	// This call is required to reap the process and release system resources.
+	// This call is required to reap the process and release all system resources.
 	Wait() Result
 
 	// Stop attempts to gracefully stop the process by sending the provided signal
-	// (defaulting to SIGTERM). If the process does not exit within the given timeout,
-	// it is forcibly killed (SIGKILL).
+	// (defaulting to SIGTERM). If the process does not exit within the given
+	// timeout, it is forcibly killed (SIGKILL).
 	Stop(timeout time.Duration, sig ...os.Signal) error
 
-	// Signal sends the given signal to the process. Returns an error if the process
-	// has not been started or no longer exists.
+	// Signal sends the given signal to the process.
 	Signal(sig os.Signal) error
 }
--- a/mocks/process/Process.go
+++ b/mocks/process/Process.go
@@ -26,6 +26,53 @@ func (_m *Process) EXPECT() *Process_Expecter {
 	return &Process_Expecter{mock: &_m.Mock}
 }
 
+// DisableBuffering provides a mock function with no fields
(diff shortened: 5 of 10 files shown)


process/running.go
```diff
@@ -14,28 +14,41 @@ type Running struct {
...
+	go func() {
+		waitErr := running.cmd.Wait()
+
+		res := buildResult(running, waitErr)
+		running.resultMu.Lock()
+		running.result = res
+		running.resultMu.Unlock()
+
+		close(running.doneChan)
+	}()
```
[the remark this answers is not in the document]
krishankumar01: No, this is not a leak because you will eventually call `Running.Wait()` to get the result.

kkumar-gcc:
Currently, I’m testing only on Unix-like systems and doing basic testing for Windows through CI/CD. I’ll need to optimize the Windows-specific logic on an actual Windows machine, since CI/CD involves a lot of back-and-forth. For now, the Windows implementation is kept very basic.

process/process_windows.go
```diff
@@ -1,8 +1,60 @@
 //go:build windows
 
+// TODO: Revisit process handling for windows, need a better way to handle process using native APIs
```
krishankumar01: Currently, I’m testing only on Unix-like systems and doing basic testing for Windows through CI/CD. I’ll need to optimize the Windows-specific logic on an actual Windows machine, since CI/CD involves a lot of back-and-forth. For now, the Windows implementation is kept very basic.

contracts/process/running_pipe.go
hwbrzzl, line 1: This file can be merged with `running.go`.

process/output_writer_test.go
```diff
@@ -13,7 +13,7 @@ func TestOutputWriter_Write_SingleLine(t *testing.T) {
 	var receivedType contractsprocess.OutputType
 	var receivedLine []byte
 
-	writer := NewOutputWriter(contractsprocess.OutputTypeStdout, func(typ contractsprocess.OutputType, line []byte) {
+	writer := NewOutputWriterForProcess(contractsprocess.OutputTypeStdout, func(typ contractsprocess.OutputType, line []byte) {
```
hwbrzzl: We need to test `NewOutputWriter` instead of `NewOutputWriterForProcess`, and there is no case for the `OutputWriter.key`.

hwbrzzl: We need to test `NewOutputWriter` instead of `NewOutputWriterForProcess`, and there is no case for the `OutputWriter.key`.

process/pipe.go
```diff
@@ -0,0 +1,232 @@
...
+
+func (r *Pipe) Input(in io.Reader) contractsprocess.Pipe {
+	r.input = in
+	return r
+}
+
+func (r *Pipe) Env(vars map[string]string) contractsprocess.Pipe {
+	if r.env == nil {
+		r.env = make([]string, 0, len(vars))
+	}
```
hwbrzzl: It's unnecessary.

```suggestion
```

contracts/process/pipe.go
```diff
@@ -0,0 +1,37 @@
...
+	Start(func(builder PipeBuilder)) (RunningPipe, error)
+	Timeout(timeout time.Duration) Pipe
+	WithContext(ctx context.Context) Pipe
+}
+
+type PipeBuilder interface {
+	Command(name string, arg ...string) *Step
+}
+
+type Step struct {
```
hwbrzzl: Please set `Step` with `interface`, and how about rename `Step` to `Command`?

contracts/process/pipe.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"context"
+	"io"
+	"time"
+)
+
+type OnPipeOutputFunc func(key string, typ OutputType, line []byte)
+
+type Pipe interface {
```
hwbrzzl: How about `Pipe -> Pipeline` and `PipeBuilder -> Pipe`.

process/pipe.go
```diff
@@ -0,0 +1,232 @@
...
+	return run.Wait(), nil
+}
+
+func (r *Pipe) start(configure func(contractsprocess.PipeBuilder)) (contractsprocess.RunningPipe, error) {
+	builder := &PipeBuilder{}
+	configure(builder)
+
+	steps := builder.steps
+	if len(steps) == 0 {
+		return nil, errors.New("pipeline must have at least one command")
```
hwbrzzl: Add it in `errors/list.go`.

process/pipe.go
```diff
@@ -0,0 +1,232 @@
...
+					_ = kill(commands[j].Process)
+				}
+			}
+			for _, w := range interWriters {
+				_ = w.Close()
+			}
+			for _, r := range interReaders {
+				_ = r.Close()
+			}
+			return nil, errors.New("failed to start pipeline: " + err.Error())
```
hwbrzzl: Ditto

process/process_unix.go
```diff
@@ -3,11 +3,91 @@
 package process
```
hwbrzzl, line 3: Shouldn't add `//go:build !windows` here?

process/process_unix.go
```diff
@@ -3,11 +3,91 @@
...
+		return false
+	}
+	// unix.Kill with signal 0 is the common alive check
+	err := unix.Kill(p.Pid, 0)
+	return err == nil
+}
+
+func kill(p *os.Process) error {
+	if p == nil {
+		return errors.New("process not started")
```
hwbrzzl: Ditto

process/process_unix_test.go
```diff
@@ -136,7 +158,56 @@ func TestProcess_OnOutput_Callbacks_Unix(t *testing.T) {
...
+
+	t.Run("Terminates process when context times out", func(t *testing.T) {
+		ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
+		defer cancel()
+
+		res, err := New().WithContext(ctx).Quietly().Run("sleep", "2")
+		assert.NoError(t, err, "Run should not return an error, but the result should indicate failure")
+		assert.NotNil(t, res, "Result object should not be nil even on failure")
+		assert.False(t, res.Successful(), "Process should have failed because it was killed")
+		assert.NotEqual(t, 0, res.ExitCode(), "Exit code should be non-zero")
```
hwbrzzl, line 191: Can we determine the cause of the failure due to the timeout?

process/process_windows.go
```diff
@@ -1,8 +1,60 @@
...
+		return errors.New("process not started")
+	}
+	return p.Signal(sig)
+}
+
+func stop(p *os.Process, _ <-chan struct{}, _ time.Duration, _ ...os.Signal) error {
+	if !running(p) {
+		return nil
+	}
+	return kill(p)
```
hwbrzzl: Why not implement this like unix?

process/process_windows.go
hwbrzzl: I suggest that the windows should be implemented and tested in this PR. If it's required, I can buy a remote windows server for the test.

process/running.go
```diff
@@ -2,40 +2,83 @@ package process
...
 
 func (r *Running) Wait() contractsprocess.Result {
-	r.once.Do(func() {
-		err := r.cmd.Wait()
-		res := buildResult(r, err)
-		r.resultMu.Lock()
-		r.result = res
-		r.resultMu.Unlock()
-	})
+	<-r.doneChan
```
hwbrzzl:
```suggestion
	<-r.Done()
```

process/running.go
```diff
@@ -2,40 +2,83 @@ package process
...
+		}
+		if runner.stderrBuffer != nil {
+			stderrStr = runner.stderrBuffer.String()
+		}
+
+		result := NewResult(exitCode, cmdStr, stdoutStr, stderrStr)
+
+		runner.resultMu.Lock()
+		runner.result = result
+		runner.resultMu.Unlock()
```
hwbrzzl: Why lock it? Is there competition here?

process/running.go
```diff
@@ -55,6 +98,38 @@ func (r *Running) Command() string {
...
+		return false
+	}
+
+	return running(r.cmd.Process)
+}
+
+func (r *Running) Kill() error {
+	if r.cmd == nil {
+		return errors.New("process not running")
+	}
```
hwbrzzl: `r.cmd` shouldn't be `nil`, right? If so, it's fine to remove the judgment in each function.

process/running.go
```diff
@@ -2,40 +2,83 @@ package process
...
+		cancel:       cancel,
 		stdoutBuffer: stdout,
 		stderrBuffer: stderr,
+		doneChan:     make(chan struct{}),
 	}
+
+	go func(runner *Running) {
+		defer func() {
+			if err := recover(); err != nil {
+				// TODO: see what should be done here with this error, should we consider it as WaitErr?
```
hwbrzzl: How about adding the error to `stderrBuffer`?

process/running_pipe.go
```diff
@@ -0,0 +1,187 @@
...
+			}
+			if runner.cancel != nil {
+				runner.cancel()
+			}
+			close(runner.doneChan)
+		}()
+
+		var waitErr error
+		for i, cmd := range runner.commands {
+			if cmd == nil {
```
hwbrzzl: Is it possible?

process/running_pipe.go
```diff
@@ -0,0 +1,187 @@
...
+				// If there's an interWriter for this position, close it so downstream doesn't block
+				if i < len(runner.interWriters) {
+					_ = runner.interWriters[i].Close()
+				}
+				continue
+			}
+
+			waitErr = cmd.Wait()
```
hwbrzzl, line 70: `waitErr` will be covered, right? I think it's unexpected?

process/running_pipe.go
```diff
@@ -0,0 +1,187 @@
...
+			if i < len(runner.interWriters) {
+				_ = runner.interWriters[i].Close()
+			}
+		}
+
+		lastIdx := len(runner.commands) - 1
+		var finalCmd *exec.Cmd
+		if lastIdx >= 0 {
+			finalCmd = runner.commands[lastIdx]
+		}
```
hwbrzzl:
```suggestion
		finalCmd = runner.commands[len(runner.commands) - 1]
```

process/running_pipe.go
```diff
@@ -0,0 +1,187 @@
...
+			_ = r.Close()
+		}
+	}(pipeRunner)
+
+	return pipeRunner
+}
+
+func (r *RunningPipe) PIDs() map[string]int {
+	m := make(map[string]int, len(r.commands))
+	for i, cmd := range r.commands {
```
hwbrzzl, line 109: `len(r.steps)` must equal to `len(commands)`, correct?

process/running_pipe.go
```diff
@@ -0,0 +1,187 @@
...
+		key := ""
+		if r.steps != nil && i < len(r.steps) && r.steps[i] != nil {
+			key = r.steps[i].Key
+		}
+		if cmd != nil && cmd.Process != nil {
+			m[key] = cmd.Process.Pid
```
hwbrzzl: `key` may be empty, it's unexpected.

process/running_pipe.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	return false
+}
+
+func (r *RunningPipe) Done() <-chan struct{} {
+	return r.doneChan
+}
+
+func (r *RunningPipe) Wait() contractsprocess.Result {
+	<-r.doneChan
```
hwbrzzl:
```suggestion
	<-r.Done()
```

process/running_pipe.go
```diff
@@ -0,0 +1,187 @@
...
+	r.resultMu.RLock()
+	defer r.resultMu.RUnlock()
```
hwbrzzl: Is it required to lock it?

process/running_pipe_unix_test.go
```diff
@@ -0,0 +1,74 @@
...
+	pids := rp.PIDs()
+	assert.Len(t, pids, 2)
+	assert.NotEqual(t, 0, pids["first"])
+	assert.NotEqual(t, 0, pids["second"])
+
+	assert.True(t, rp.Running())
+
+	res := rp.Wait()
+	assert.True(t, strings.HasPrefix(res.Output(), "start"))
+	assert.True(t, strings.HasSuffix(res.Output(), "end"))
```
hwbrzzl: Can't we judge the whole string?

hwbrzzl (commented):
Awesome 👍

process/process_unix.go
kkumar-gcc: missed it

krishankumar01: It already has build tags, check by expanding it.

process/process_windows.go
krishankumar01: Graceful shutdown needs to be handled using Windows job pipelines, which is a bit complicated. I have this in my TODO list and am currently exploring Windows’ native APIs.

contracts/process/pipe.go
krishankumar01: I’m not sure if `Step` needs to be an interface, since it’s very simple and doesn’t need to be. It’s just a struct carrying basic data.

contracts/process/running_pipe.go
krishankumar01: Both have different implementations, and putting them together will create unnecessary confusion. I suggest dividing them into two separate files.

process/process_unix_test.go
krishankumar01: No, in case of a timeout it also sends a kill signal, which will be returned instead of being specific to the timeout. But let me see if I can find a workaround here.

process/running.go
krishankumar01: Not required, it was only defensive. I’ve removed it.

process/running_pipe.go
krishankumar01: If the key is empty, the index of the steps will be used as the key.

process/running.go
krishankumar01: Added to `stderrBuffer`, see if this makes sense.

process/running_pipe.go
krishankumar01: sorry, didn't get you

process/pipeline.go
```diff
@@ -88,21 +86,21 @@ func (r *Pipe) WithContext(ctx context.Context) contractsprocess.Pipe {
...
 	if err != nil {
 		return nil, err
 	}
 	return run.Wait(), nil
 }
 
-func (r *Pipe) start(configure func(contractsprocess.PipeBuilder)) (contractsprocess.RunningPipe, error) {
-	builder := &PipeBuilder{}
+func (r *Pipeline) start(configure func(contractsprocess.Pipe)) (contractsprocess.RunningPipe, error) {
+	builder := &Pipe{}
```
hwbrzzl:
```suggestion
	pipe := &Pipe{}
```

contracts/process/pipe.go
hwbrzzl: Yes, it's needed. It's better that `constants` doesn't contain `struct`.

process/process_unix.go
hwbrzzl: Oh, yes.

process/running_pipe.go
hwbrzzl: `waitErr` seems to be discarded. Can users get it via some way?

kkumar-gcc: sure updated it to interface

PR #1181, opened 2025-09-03
fix: dynamic module name in test templates

BadJacky:
## 📑 Description

Replace hardcoded "goravel" module import in test templates with
dynamic module name detection using debug.ReadBuildInfo().

- Add GetModuleName() method to Make struct
- Update populateStub() to accept and replace DummyModule placeholder
- Update test template stub to use DummyModule instead of hardcoded path
- Add test verification for correct module import path
- Fallback to "goravel" if build info unavailable

This ensures generated test files use correct import paths regardless
of the actual module name, improving framework reusability.

## ✅ Checks


hwbrzzl (approved):
Thanks, good catch!

PR #1182, opened 2025-09-04
feat: csrf middleware

NiteshSingh17:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/218

## ✅ Checks

### DIFF
--- a/http/middleware/verify_csrf_token.go
+++ b/http/middleware/verify_csrf_token.go
@@ -18,7 +18,7 @@ func VerifyCsrfToken(excepts []string) contractshttp.Middleware {
 			ctx.Request().Next()
 			ctx.Response().Header(HeaderCsrfKey, ctx.Request().Session().Token())
 		} else {
-			ctx.Request().AbortWithStatusJson(contractshttp.StatusTokenMismatch, map[string]string{"message": contractshttp.StatusText(contractshttp.StatusTokenMismatch)})
+			ctx.Request().Abort(contractshttp.StatusTokenMismatch)
 		}
 	}
 }
--- a/http/middleware/verify_csrf_token_test.go
+++ b/http/middleware/verify_csrf_token_test.go
@@ -11,6 +11,9 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
 	contractshttp "github.com/goravel/framework/contracts/http"
 	contractsession "github.com/goravel/framework/contracts/session"
 	foundationJson "github.com/goravel/framework/foundation/json"
@@ -18,8 +21,6 @@ import (
 	"github.com/goravel/framework/session"
 	sessionMiddleware "github.com/goravel/framework/session/middleware"
 	"github.com/goravel/framework/support/path"
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 )
 
 func testHttpSessionMiddleware(next nethttp.Handler, mockConfig *configmocks.Config) nethttp.Handler {
@@ -47,13 +48,13 @@ func testHttpSessionMiddleware(next nethttp.Handler, mockConfig *configmocks.Con
 }
 
 func mockConfigFacade(mockConfig *configmocks.Config) {
-	mockConfig.On("GetString", "session.default").Return("file").Once()
-	mockConfig.On("GetInt", "session.lifetime", 120).Return(120).Once()
-	mockConfig.On("GetString", "session.path").Return("/").Once()
-	mockConfig.On("GetString", "session.domain").Return("").Once()
-	mockConfig.On("GetBool", "session.secure").Return(false).Once()
-	mockConfig.On("GetBool", "session.http_only").Return(true).Once()
-	mockConfig.On("GetString", "session.same_site").Return("").Once()
+	mockConfig.EXPECT().GetString("session.default").Return("file").Once()
+	mockConfig.EXPECT().GetInt("session.lifetime", 120).Return(120).Once()
+	mockConfig.EXPECT().GetString("session.path").Return("/").Once()
+	mockConfig.EXPECT().GetString("session.domain").Return("").Once()
+	mockConfig.EXPECT().GetBool("session.secure").Return(false).Once()
+	mockConfig.EXPECT().GetBool("session.http_only").Return(true).Once()
+	mockConfig.EXPECT().GetString("session.same_site").Return("").Once()
 }
 
 func TestVerifyCsrfToken(t *testing.T) {
@@ -203,13 +204,6 @@ func NewTestRequest(ctx *TestContext) *TestRequest {
 	}
 }
 
-func (r *TestRequest) AbortWithStatusJson(code int, jsonObj any) {
-	r.ctx.writer.WriteHeader(code)
-	if err := json.NewEncoder(r.ctx.writer).Encode(jsonObj); err != nil {
-		nethttp.Error(r.ctx.writer, "Failed to encode JSON", nethttp.StatusInternalServerError)
-	}
-}
-
 func (r *TestRequest) Path() string {
 	if r.ctx != nil && r.ctx.request != nil {
 		return r.ctx.request.URL.Path
@@ -258,6 +252,7 @@ func (r *TestRequest) SetSession(session contractsession.Session) contractshttp.
 }
 
 func (r *TestRequest) Abort(code ...int) {
+	r.ctx.writer.WriteHeader(code[0])
 }
 
 func (r *TestRequest) Next() {


http/middleware/CSRFToken.go
```diff
@@ -0,0 +1,55 @@
...
+	httpcontract "github.com/goravel/framework/contracts/http"
+)
+
+const csrfKey = "X-CSRF-TOKEN"
+
+type BodyPayload struct {
+	CSRFToken string `json:"csrf_token"`
+}
+
+func CSRFToken() httpcontract.Middleware {
```
hwbrzzl:
```suggestion
func VerifyCsrfToken() httpcontract.Middleware {
```

http/middleware/CSRFToken.go
```diff
@@ -0,0 +1,55 @@
...
+	"io"
+	"strings"
+
+	httpcontract "github.com/goravel/framework/contracts/http"
+)
+
+const csrfKey = "X-CSRF-TOKEN"
+
+type BodyPayload struct {
+	CSRFToken string `json:"csrf_token"`
```
hwbrzzl:
```suggestion
	CsrfToken string `json:"csrf_token"`
```

http/middleware/CSRFToken.go
```diff
@@ -0,0 +1,55 @@
+package middleware
+
+import (
+	"encoding/json"
+	"io"
+	"strings"
+
+	httpcontract "github.com/goravel/framework/contracts/http"
```
hwbrzzl:
```suggestion
	contractshttp "github.com/goravel/framework/contracts/http"
```

http/middleware/CSRFToken.go
```diff
@@ -0,0 +1,55 @@
...
+type BodyPayload struct {
+	CSRFToken string `json:"csrf_token"`
+}
+
+func CSRFToken() httpcontract.Middleware {
+	return func(ctx httpcontract.Context) {
+		sessionCSRFTokenInterface := ctx.Request().Session().Get(csrfKey)
+		sessionCSRFToken, ok := sessionCSRFTokenInterface.(string)
+		if !ok || sessionCSRFToken == "" {
+			ctx.Request().Abort(httpcontract.StatusUnauthorized)
```
hwbrzzl: Is `StatusTokenMismatch` better?

http/middleware/CSRFToken.go
```diff
@@ -0,0 +1,55 @@
...
+					return
+				}
+
+				httpReq.Body = io.NopCloser(strings.NewReader(string(rawBody)))
+				var input BodyPayload
+				if err := json.Unmarshal(rawBody, &input); err == nil && input.CSRFToken != "" {
+					requestCSRFToken = input.CSRFToken
+				}
+			}
+		}
```
hwbrzzl: `ctx.Request().Input("csrf_token")` can be used here.

http/middleware/CSRFToken.go
```diff
@@ -0,0 +1,55 @@
...
+
+		if requestCSRFToken == "" || requestCSRFToken != sessionCSRFToken {
```
hwbrzzl: Please use `subtle.ConstantTimeCompare` to compare these two tokens.

http/middleware/CSRFToken.go
```diff
@@ -0,0 +1,55 @@
...
+			ctx.Request().Abort(httpcontract.StatusUnauthorized)
+			return
+		}
+
+		ctx.Request().Next()
```
hwbrzzl: We need  to implement the adding token action as well: https://github.com/laravel/framework/blob/5a0715f939b6b4da078610b596ca0b3d7f37d58a/src/Illuminate/Foundation/Http/Middleware/VerifyCsrfToken.php#L88-L89

http/middleware/CSRFToken.go
```diff
@@ -0,0 +1,55 @@
...
+
+const csrfKey = "X-CSRF-TOKEN"
+
+type BodyPayload struct {
+	CSRFToken string `json:"csrf_token"`
+}
+
+func CSRFToken() httpcontract.Middleware {
+	return func(ctx httpcontract.Context) {
+		sessionCSRFTokenInterface := ctx.Request().Session().Get(csrfKey)
```
krishankumar01: Session `csrf` token can be obtained using the `Token` method. Make sure the session is set, otherwise it will panic. You can use the `HasSession` method from `Request()` to check this.

http/middleware/CSRFToken.go
```diff
@@ -0,0 +1,55 @@
...
+		sessionCSRFToken, ok := sessionCSRFTokenInterface.(string)
```
krishankumar01: `Token` method returns a `string` type so this won't be required

http/middleware/CSRFToken.go
NiteshSingh17: @kkumar-gcc,

I tested the regenerate function, which is responsible for generating a new `CSRF` token and setting it on `_token`. However, I noticed that when I send a subsequent request and access the token using `Session.Token()`, it still returns the token from the previous request  not the updated one generated in the current request.

Could you please take a look and test this behavior? This could potentially lead to issues in the future when passing the _token to HTML forms, as the form would use an outdated CSRF token.

Here’s the middleware function I'm currently using for testing:
```

func VerifyCsrfToken(excepts []string) contractshttp.Middleware {
	return func(ctx contractshttp.Context) {
		if isReading(ctx.Request().Method()) || inExceptArray(excepts, ctx.Request().FullUrl()) || tokenMatch(ctx) {
			ctx.Request().Next()
			ctx.Request().Session().Regenerate()
			ctx.Request().Session().Put(csrfKey, ctx.Request().Session().Token())
		} else {
			ctx.Request().AbortWithStatusJson(contractshttp.StatusTokenMismatch, map[string]string{
				"message": "CSRF token mismatch.",
			})
		}
	}
}

```

krishankumar01: Yes, that’s expected behavior. The `CSRF` token in session isn’t meant to change on every request. It only needs to be rotated on important events like login, logout, privilege changes, or when the session is migrated. Rotating it per request would cause unnecessary complexity without adding real security benefits. Though we can expose `session.RegenerateToken()` in case we ever need to manually force a new token in any flow.

cc: @hwbrzzl

http/middleware/VerifyCsrfToken.go
```diff
@@ -0,0 +1,65 @@
...
+	for _, pattern := range excepts {
+		if matched, err := path.Match(pattern, currentPath); err == nil && matched {
+			return true
+		}
+	}
+	return false
+}
+
+func isReading(method string) bool {
+	return method == "GET" || method == "HEAD" || method == "OPTIONS"
```
krishankumar01: We can use `contractshttp.Method.*` instead of hard-coded strings:
```suggestion
	return method == contractshttp.MethodGet || method == contractshttp.MethodHead || method == contractshttp.MethodOptions
```

contracts/http/status.go
```diff
@@ -73,6 +73,7 @@ const (
 	StatusLoopDetected                  = http.StatusLoopDetected
 	StatusNotExtended                   = http.StatusNotExtended
 	StatusNetworkAuthenticationRequired = http.StatusNetworkAuthenticationRequired
+	StatusTokenMismatch                 = 419
```
krishankumar01, line 76: Add the corresponding `StatusText` as well.

http/middleware/CSRFToken.go
hwbrzzl: > I tested the regenerate function, which is responsible for generating a new CSRF token and setting it on _token. However, I noticed that when I send a subsequent request and access the token using Session.Token(), it still returns the token from the previous request not the updated one generated in the current request.

@NiteshSingh17 Could you add some screenshots for your test? I'm not sure if I got the correct information on this.

hwbrzzl (commented):
Thanks, LGTM, could you add some screenshots to confirm if the middleware is worked as expected?

http/middleware/CSRFToken.go
NiteshSingh17: https://github.com/user-attachments/assets/e4d6aaa0-8a3d-4149-944b-17c196e72fb9

I thought there might be an issue because I was expecting the CSRF token to reset with every request. But keeping it the same throughout the session makes sense and works fine.

NiteshSingh17:
Schrrenshots

Middelware setup
Screenshot from 2025-09-07 21-20-50

Protecetd page withou CSRF
Screenshot from 2025-09-07 21-22-06

Unprojtected page without CSRF
Screenshot from 2025-09-07 21-21-24

Protected page with correct CSRF
Screenshot from 2025-09-07 21-22-43
Screenshot from 2025-09-07 21-23-54

http/middleware/VerifyCsrfToken.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+const csrfKey = "X-CSRF-TOKEN"
+
+func VerifyCsrfToken(excepts []string) contractshttp.Middleware {
+	absolutePaths := parseExceptPaths(excepts)
+	return func(ctx contractshttp.Context) {
+		if isReading(ctx.Request().Method()) || inExceptArray(absolutePaths, ctx.Request().Path()) || tokenMatch(ctx) {
+			ctx.Request().Next()
+			ctx.Request().Session().Put(csrfKey, ctx.Request().Session().Token())
```
krishankumar01: Why are we storing `csrfKey` in the session? Also, please add unit tests for this middleware.

hwbrzzl: Good question, it seems that the token should be set in the header and cookie? And yes, it's better to add some test cases for this feature. We are going to add test cases for `Throttle` as well in the future.

http/middleware/CSRFToken.go
hwbrzzl: You may need to `Save` the session after regenerating.

krishankumar01: `Save` will be automatically called in the `StartSession` middleware. In Laravel, they also regenerate the token only on specific events instead of on every request. According to the current implementation, `VerifyCsrfValidation` will only work with the `StartSession` middleware (which is expected).

Screenshot 2025-09-08 at 3 06 23 PM

http/middleware/VerifyCsrfToken.go
krishankumar01: Yes, we need to include the token in the headers and cookies as well. @NiteshSingh17 you can check the `session/middleware.StartSession` test cases for reference.

ref: https://laravel.com/docs/12.x/csrf#csrf-x-xsrf-token

NiteshSingh17: @hwbrzzl @kkumar-gcc 
Do we need to set the `X-XSRF-TOKEN` in the response headers?I couldn't find its implementation in the current codebase.

https://github.com/laravel/framework/blob/bb00401b3979c507ec614d537ea7518cb065eac0/src/Illuminate/Foundation/Http/Middleware/VerifyCsrfToken.php#L189C1-L190C1

hwbrzzl: Yes, the token may be added via an implicit code. Otherwise, there is no way to fetch the token in the header.

http/middleware/verify_csrf_token.go
```diff
@@ -0,0 +1,65 @@
...
+const HeaderCsrfKey = "X-CSRF-TOKEN"
+
+func VerifyCsrfToken(excepts []string) contractshttp.Middleware {
+	absolutePaths := parseExceptPaths(excepts)
+	return func(ctx contractshttp.Context) {
+		if isReading(ctx.Request().Method()) || inExceptArray(absolutePaths, ctx.Request().Path()) || tokenMatch(ctx) {
+			ctx.Request().Next()
+			ctx.Response().Header(HeaderCsrfKey, ctx.Request().Session().Token())
+		} else {
+			ctx.Request().AbortWithStatusJson(contractshttp.StatusTokenMismatch, map[string]string{"message": contractshttp.StatusText(contractshttp.StatusTokenMismatch)})
```
hwbrzzl: `AbortWithStatusJson` is deprecated, please use `ctx.Response().Json().Abort()` instead.

http/middleware/verify_csrf_token.go
```diff
@@ -0,0 +1,65 @@
...
+func tokenMatch(ctx contractshttp.Context) bool {
+	if !ctx.Request().HasSession() {
+		return false
+	}
+	sessionCsrfToken := ctx.Request().Session().Token()
+	requestCsrfToken := ctx.Request().Header(HeaderCsrfKey)
+	if requestCsrfToken == "" {
+		requestCsrfToken = ctx.Request().Input("_token")
+	}
+	if requestCsrfToken == "" || subtle.ConstantTimeCompare([]byte(requestCsrfToken), []byte(sessionCsrfToken)) == 0 {
```
hwbrzzl, line 35: They may be compared by length first, otherwise, `ConstantTimeCompare` will panic.

```suggestion
	if requestCsrfToken == "" || len(requestCsrfToken) !=len(sessionCsrfToken) || subtle.ConstantTimeCompare([]byte(requestCsrfToken), []byte(sessionCsrfToken)) == 0 {
```

http/middleware/verify_csrf_token_test.go
```diff
@@ -0,0 +1,334 @@
...
+
+	contractshttp "github.com/goravel/framework/contracts/http"
+	contractsession "github.com/goravel/framework/contracts/session"
+	foundationJson "github.com/goravel/framework/foundation/json"
+	configmocks "github.com/goravel/framework/mocks/config"
+	"github.com/goravel/framework/session"
+	sessionMiddleware "github.com/goravel/framework/session/middleware"
+	"github.com/goravel/framework/support/path"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
```
hwbrzzl: Sort it

```suggestion

	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/require"
	
	contractshttp "github.com/goravel/framework/contracts/http"
	contractsession "github.com/goravel/framework/contracts/session"
	foundationJson "github.com/goravel/framework/foundation/json"
	configmocks "github.com/goravel/framework/mocks/config"
	"github.com/goravel/framework/session"
	sessionMiddleware "github.com/goravel/framework/session/middleware"
	"github.com/goravel/framework/support/path"
```

http/middleware/verify_csrf_token_test.go
```diff
@@ -0,0 +1,334 @@
...
+		}
+
+		mockConfigFacade(mockConfig)
+		sessionMiddleware.StartSession()(ctx)
+
+	})
+}
+
+func mockConfigFacade(mockConfig *configmocks.Config) {
+	mockConfig.On("GetString", "session.default").Return("file").Once()
```
hwbrzzl:
```suggestion
	mockConfig.EXPECT().GetString("session.default").Return("file").Once()
```

http/middleware/verify_csrf_token_test.go
hwbrzzl, line 1: Thanks for the test cases! It's a bit complex. Actually, `contractshttp.Context` can be mocked, and we can just test the functions in verify_csrf_token.go one by one. But it doesn't matter, you have done great work, we can merge this PR first if you want. You or I can optimize the test cases later.

http/middleware/verify_csrf_token.go
NiteshSingh17: Screenshot from 2025-09-11 20-31-48

`subtle.ConstantTimeCompare` performs the comparison internally and does not panic in the event of a length mismatch."

http/middleware/verify_csrf_token_test.go
NiteshSingh17: I agree that the implementation was a bit complex. Thank you for your support. We can go ahead and merge the PR now, and optimize the test cases later  :+1:

hwbrzzl:
@NiteshSingh17 We can optimize the gin PR and create a PR for fiber next.

PR #1192, opened 2025-09-12
feat: install Crypt facade

hwbrzzl:
## 📑 Description

Relate https://github.com/goravel/goravel/issues/612

image

## ✅ Checks

### DIFF
--- a/packages/setup.go
+++ b/packages/setup.go
@@ -24,7 +24,7 @@ type setup struct {
 
 var osExit = os.Exit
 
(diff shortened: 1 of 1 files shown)


support/env/env.go
```diff
@@ -114,3 +115,21 @@ func CurrentAbsolutePath() string {
 
 	return res
 }
+
+func ModuleName() string {
```
almas-x: How about moving this function to `packages/setup.go`? There’s already a similar `GetModulePath` function there.

hwbrzzl: Good point

PR #1199, opened 2025-09-14
feat: command - add support for typed cli arguments

oprudkyi:
## 📑 Description

- adds support for https://cli.urfave.org/v3/examples/arguments/advanced/
- `--help` prints arguments according to https://pubs.opengroup.org/onlinepubs/9799919799/

- currently it follows Flag/Options naming pattern , i.e. 
```
	return command.Extend{
		Arguments: []command.Argument{
			&command.StringArgument{
				Name:     "my_string",
				Usage:    "string argument",
				Required: true,
			},
....
	ctx.Info(fmt.Sprintf("my string: %v", ctx.ArgumentString("my_string")))

```
but what confuses me, is inconsistency in naming, i.e. type `StringArgument` but method `ArgumentString()`, 
probably type also should be named ArgumentString ? 

(probably the same should be done for Flags as well, `BoolFlag` vs `OptionBool()`)

also there are changes I am not sure if they are fine globally 
- without this I see error message with pointer, not text
```
func printTemplateError(err error) {
	if os.Getenv("CLI_TEMPLATE_ERROR_DEBUG") != "" {
-		_, _ = fmt.Fprintf(cli.ErrWriter, "CLI TEMPLATE ERROR: %#v\n", err)
+		_, _ = fmt.Fprintf(cli.ErrWriter, "CLI TEMPLATE ERROR: %+v\n", err)
	}
}
```

- making things testable but just exit with on error 
```
+			errString := fmt.Sprintf("Required argument with value '%+v' should be placed before any not-required arguments", v)
+			if testing.Testing() {
+				panic(errString)
+			} else {
+				color.Errorln(errString)
+				os.Exit(2)
+			}
```

## ✅ Checks


almas-x:
Amazing👍

console/application.go
```diff
@@ -229,3 +233,265 @@ func flagsToCliFlags(flags []command.Flag) []cli.Flag {
...
+func argumentsToCliArgs(args []command.Argument) []cli.Argument {
+	len := len(args)
+	if len == 0 {
+		return nil
+	}
+	cliArgs := make([]cli.Argument, 0, len)
+	previousIsRequired := true
+	for _, v := range args {
+		if v.MinOccurrences() != 0 && !previousIsRequired {
+			errString := fmt.Sprintf("Required argument with value '%+v' should be placed before any not-required arguments", v)
```
almas-x: It would be better to consolidate all error messages into the existing `errors/list.go` file.

console/application.go
```diff
@@ -229,3 +233,265 @@ func flagsToCliFlags(flags []command.Flag) []cli.Flag {
...
+			if testing.Testing() {
+				panic(errString)
+			} else {
+				color.Errorln(errString)
+				os.Exit(2)
+			}
```
almas-x: I think it would be better to add a check to determine if the environment is an Artisan environment. If it is an Artisan environment, we can output the error message and perform os.Exit. For other cases, such as test environments or non-Artisan execution environments, we can only output the error message without making the `argument`or `command` effective, ensuring that the application as a whole continues to run.

Additionally, it might be worth mentioning that `os.Exit` can be assigned to a variable, for example, `var osExit = os.Exit`, so that it can be replaced in unit tests to make the code testable.

console/application.go
```diff
@@ -229,3 +233,265 @@ func flagsToCliFlags(flags []command.Flag) []cli.Flag {
...
+		}
+		if v.MinOccurrences() != 0 {
+			previousIsRequired = true
+		} else {
+			previousIsRequired = false
+		}
+		switch arg := v.(type) {
+		case *command.Float32Argument:
```
hwbrzzl: > but what confuses me, is inconsistency in naming, i.e. type StringArgument but method ArgumentString(),
probably type also should be named ArgumentString ?

Yes, good point. Keeping consistency should be better: `Argument*`. By the way, about the `Flags` option, how about modifying it to `Options`? Keep it consistent with `Option*` functions. And `type BoolFlag struct`, etc. structs can be optimized as well. Adding the deprecated flag for them.  cc @almas-x

console/application.go
hwbrzzl: > I think it would be better to add a check to determine if the environment is an Artisan environment.

I think it should be fine, the application can't be built successfully if there are problems, given that this function is called by `Register`, and `Register` is called when initiating the framework.

console/cli_context_args_test.go
hwbrzzl, line 1: How about merging this file to `cli_context_test.go`.

console/cli_helper.go
```diff
@@ -57,7 +57,9 @@ var (
   {{$s := join .Names ", "}}{{green $s}}{{ $sp := subtract $cv (offset $s 3) }}{{ indent $sp ""}}{{wrap (colorize .Usage) $cv}}{{end}}{{end}}`
 	flagTemplate = `{{ $cv := offsetFlags . 5}}{{range  .}}
    {{$s := getFlagName .}}{{green $s}}{{ $sp := subtract $cv (offset $s 1) }}{{ indent $sp ""}}{{$us := (capitalize .Usage)}}{{wrap (colorize $us) $cv}}{{$df := getFlagDefaultText . }}{{if $df}} {{yellow $df}}{{end}}{{end}}`
-	usageTemplate = `{{if .UsageText}}{{wrap (colorize .UsageText) 3}}{{else}}{{(helpName .FullName)}}{{if .VisibleFlags}} [options]{{end}}{{if .ArgsUsage}}{{.ArgsUsage}}{{else}}{{if .Args}} [arguments...]{{end}}{{end}}{{end}}`
+	usageTemplate = `{{if .UsageText}}{{wrap (colorize .UsageText) 3}}{{else}}{{(helpName .FullName)}}{{if .VisibleFlags}} [options]{{end}}{{if .ArgsUsage}}{{.ArgsUsage}}{{else}}{{if .Arguments}}{{template "argsTemplate" .}}{{end}}{{end}}{{end}}`
+	argsTemplate  = `{{if .Arguments}}{{range .Arguments}}{{template "argTemplate" .}}{{end}}{{end}}`
+	argTemplate   = ` {{if .Min}}<{{else}}[{{end}}{{.Name}}{{if (or (gt .Max 1) (eq .Max -1))}}...{{end}}{{if .Min}}>{{else}}]{{end}}`
```
hwbrzzl, line 62: Could you add a screenshot for this printing?

hwbrzzl (commented):
Awesome, thanks! A few of questions.

console/application.go
almas-x: > > I think it would be better to add a check to determine if the environment is an Artisan environment.
> 
> I think it should be fine, the application can't be built successfully if there are problems, given that this function is called by `Register`, and `Register` is called when initiating the framework.

Regarding this point, the error here is due to having a required argument after an optional argument. This is a logical mistake, but it doesn't affect the build process. However, the application will call `os.Exit(2)` during startup, which is very unfriendly. It's acceptable to interrupt execution when running `artisan` commands, but if you're simply starting a web service (not running artisan), being forced to exit is terrible. You can refer to how Laravel handles this scenario.

The worst-case scenario is: you install a third-party package, everything works fine, and one day after upgrading, the package introduces a new command. If there is a configuration issue with the command arguments, `os.Exit` will be called—even if you never use the artisan commands from that package, your application won't be able to start properly.

console/application.go
almas-x: I'd like to confirm: Are you suggesting merging the existing `Flag` and the new `Argument` introduced in this PR into `Option`? If so, wouldn't this potentially introduce significant breaking changes? It seems that all existing artisan commands would encounter type errors, since the `command.Extend` method in command definitions currently refers to the `xxxFlag` structs.

console/application.go
hwbrzzl: Good point, so there are two solutions when there is an argument issue:

1. Exit the app, `go build` and `go run` will fail, user needs to fix the argument issue immediately.
2. Print the error with a graceful message, and the command will not be registered.

The second should be fine.

console/application.go
hwbrzzl: > Are you suggesting merging the existing Flag and the new Argument introduced in this PR into Option?

No, I mean copy all flags to options, then add a `DEPRECATED` flag for flags. There is no breaking change. 

Then the logic will be:

Set `Options` in `Extend` -> Use `ctx.Option*` to get data.
Set `Arguments` in `Extend` -> Use `ctx.Argument*` to get data.

hwbrzzl: We can implement it in another PR if it's acceptable.

console/cli_helper.go
almas-x: Speaking of help information output, the current implementation uses Go templates, which are honestly very hard to maintain—and debugging is even more troublesome. That's why I've always wanted to refactor this part and directly use the framework's color package for printing. This way, handling command sorting, alignment, and indentation would be much easier. In fact, I even submitted a PR to `urfave/cli` specifically for this purpose: [@urfave/cli/pull/2150](https://github.com/urfave/cli/pull/2150).

oprudkyi: @hwbrzzl this is my testing command with various cases, required/optional/slice 
image

oprudkyi:
summary for further changes:
- It would be better to consolidate all error messages into the existing errors/list.go file. ( @almas-x )
- on Register() errors - print the error with a graceful message, and the command will not be registered.  (@hwbrzzl)
- rename argument types, i.e. Float32Argument to ArgumentFloat32 (@hwbrzzl)
- merge console/cli_context_args_test.go into cli_context_test.go (@hwbrzzl)

hwbrzzl:
> summary for further changes:
> 
> * It would be better to consolidate all error messages into the existing errors/list.go file. ( @almas-x )
> * on Register() errors - print the error with a graceful message, and the command will not be registered.  (@hwbrzzl)
> * rename argument types, i.e. Float32Argument to ArgumentFloat32 (@hwbrzzl)
> * merge console/cli_context_args_test.go into cli_context_test.go (@hwbrzzl)

Yes, thanks!

oprudkyi:
@hwbrzzl @almas-x  
done. 
also I am in progress with adding wrappers in onUsageError 
but it seems the info about argument isn't provided always , so it is delayed https://github.com/urfave/cli/pull/2203

console/application.go
oprudkyi: updated

console/application.go
```diff
@@ -229,3 +236,258 @@ func flagsToCliFlags(flags []command.Flag) []cli.Flag {
...
+			})
+		case *command.ArgumentTimestampSlice:
+			cliArgs = append(cliArgs, &cli.TimestampArgs{
+				Name:      arg.Name,
+				UsageText: arg.Usage,
+				Value:     arg.Value,
+				Min:       arg.MinOccurrences(),
+				Max:       arg.MaxOccurrences(),
+				Config: cli.TimestampConfig{
+					Layouts: []string{time.RFC3339},
```
almas-x: Ditto

console/application.go
```diff
@@ -229,3 +236,258 @@ func flagsToCliFlags(flags []command.Flag) []cli.Flag {
...
+			})
+		case *command.ArgumentTimestamp:
+			cliArgs = append(cliArgs, &cli.TimestampArgs{
+				Name:      arg.Name,
+				UsageText: arg.Usage,
+				Value:     arg.Value,
+				Min:       arg.MinOccurrences(),
+				Max:       arg.MaxOccurrences(),
+				Config: cli.TimestampConfig{
+					Layouts: []string{time.RFC3339},
```
almas-x: How about adding a time format configuration option to `command.ArgumentTimestamp`? This would allow users to specify their desired format, making it more flexible and user-friendly.

console/application.go
```diff
@@ -50,6 +52,10 @@ func NewApplication(name, usage, usageText, version string, useArtisan bool) con
 func (r *Application) Register(commands []console.Command) {
 	for _, item := range commands {
 		item := item
+		arguments, err := argumentsToCliArgs(item.Extend().Arguments)
+		if err != nil {
+			color.Errorln(errors.ConsoleCommandRegistrationFailed.Args(item.Signature(), err).Error())
```
almas-x, line 56: After printing the error here, it might be better to use `continue` to skip the iteration, right?

oprudkyi: added `continue`

I did it without `continue` because for me it provides at least some context for user 
```
./artisan smtp:add  --help      
  ERROR   Registration of command 'smtp:add' failed: required argument 'host' should be placed before any not-required arguments
Description:
   Add SMTP server

   Stores SMTP server by provided server_id into database


Usage:
   artisan [global options] smtp:add [options]
```

with `continue` 
```
./artisan smtp:add  --help
  ERROR   Registration of command 'smtp:add' failed: required argument 'host' should be placed before any not-required arguments
  ERROR   Command 'smtp:add' is not defined. Did you mean one of these?
```

console/application.go
oprudkyi: added support for Layouts

hwbrzzl:
Hey @oprudkyi Thanks for your contribution. v1.17 has been released. We'd like to send you a T-shirt as a gift. Could you connect with me on Discord? We want to get your address and help find an online store for the T-shirt.

PR #1215, opened 2025-10-03
feat: Add pool support in process package

krishankumar01:
## 📑 Description

**RelatedTo** https://github.com/goravel/goravel/issues/567

This PR implements a lightweight, flexible API for managing concurrent processes in Go.

## Basic Usage

```go
// Run multiple commands
results, err := process.NewPool().Run(func(p process.Pool) {
    p.Command("echo", "hello").As("greeting")
    p.Command("ls", "-la").As("files")
})

fmt.Println(results["greeting"].Output())
```

## Features

- Run multiple commands concurrently
- Control concurrency limits
- Stream output in real-time
- Set timeouts and contexts
- Send signals to running processes

### Limit Concurrency

```go
// Run max 2 commands at once
process.NewPool().Concurrency(2).Run(func(p process.Pool) {
    p.Command("task1").As("job1")
    p.Command("task2").As("job2")
    p.Command("task3").As("job3")
})
```

### Stream Output

```go
process.NewPool().OnOutput(func(key string, typ process.OutputType, line []byte) {
    fmt.Printf("[%s] %s", key, string(line))
}).Run(func(p process.Pool) {
    p.Command("slow-command").As("slow")
})
```

### Control Running Processes

```go
running, err := process.NewPool().Start(func(p process.Pool) {
    p.Command("server").As("server")
})

pids := running.PIDs()

running.Signal(syscall.SIGTERM)

results := running.Wait()
```

### Command Options

```go
p.Command("cmd").
    Path("/custom/dir").                     // Working directory
    Env(map[string]string{"KEY": "value"}).  // Environment variables
    Timeout(5 * time.Second).                // Command timeout
    WithContext(ctx).                        // Command context
    DisableBuffering().                      // Don't buffer output
    Quietly().                               // Suppress output
    As("name")                               // Set command name
```

## Notes

- Use `DisableBuffering()` for commands with large output
- Command-level contexts override pool-level contexts

## ✅ Checks

### DIFF
--- a/contracts/process/pool.go
+++ b/contracts/process/pool.go
@@ -6,7 +6,7 @@ import (
 	"time"
 )
 
-type OnPoolOutputFunc func(typ OutputType, key string, line []byte)
+type OnPoolOutputFunc func(typ OutputType, line []byte, key string)
 
 // PoolBuilder defines the interface for configuring and launching a pool of concurrent processes.
 type PoolBuilder interface {
--- a/process/pool.go
+++ b/process/pool.go
@@ -157,7 +157,7 @@ func (r *PoolBuilder) start(configure func(contractsprocess.Pool)) (contractspro
 				}
 				if r.onOutput != nil {
 					proc = proc.OnOutput(func(typ contractsprocess.OutputType, line []byte) {
-						r.onOutput(typ, command.key, line)
+						r.onOutput(typ, line, command.key)
 					})
 				}
 
--- a/contracts/process/pipeline.go
+++ b/contracts/process/pipeline.go
@@ -6,7 +6,7 @@ import (
 	"time"
 )
 
-type OnPipeOutputFunc func(typ OutputType, key string, line []byte)
+type OnPipeOutputFunc func(typ OutputType, line []byte, key string)
 
 // Pipeline defines a builder-style API for constructing and running a sequence
 // of commands connected via OS pipes. Implementations are mutable and should
--- a/mocks/process/OnPipeOutputFunc.go
+++ b/mocks/process/OnPipeOutputFunc.go
@@ -20,9 +20,9 @@ func (_m *OnPipeOutputFunc) EXPECT() *OnPipeOutputFunc_Expecter {
 	return &OnPipeOutputFunc_Expecter{mock: &_m.Mock}
 }
 
-// Execute provides a mock function with given fields: typ, key, line
-func (_m *OnPipeOutputFunc) Execute(typ process.OutputType, key string, line []byte) {
-	_m.Called(typ, key, line)
+// Execute provides a mock function with given fields: typ, line, key
+func (_m *OnPipeOutputFunc) Execute(typ process.OutputType, line []byte, key string) {
+	_m.Called(typ, line, key)
 }
 
 // OnPipeOutputFunc_Execute_Call is a *mock.Call that shadows Run/Return methods with type explicit version for method 'Execute'
@@ -32,15 +32,15 @@ type OnPipeOutputFunc_Execute_Call struct {
 
 // Execute is a helper method to define mock.On call
 //   - typ process.OutputType
-//   - key string
 //   - line []byte
-func (_e *OnPipeOutputFunc_Expecter) Execute(typ interface{}, key interface{}, line interface{}) *OnPipeOutputFunc_Execute_Call {
-	return &OnPipeOutputFunc_Execute_Call{Call: _e.mock.On("Execute", typ, key, line)}
+//   - key string
+func (_e *OnPipeOutputFunc_Expecter) Execute(typ interface{}, line interface{}, key interface{}) *OnPipeOutputFunc_Execute_Call {
+	return &OnPipeOutputFunc_Execute_Call{Call: _e.mock.On("Execute", typ, line, key)}
 }
 
-func (_c *OnPipeOutputFunc_Execute_Call) Run(run func(typ process.OutputType, key string, line []byte)) *OnPipeOutputFunc_Execute_Call {
+func (_c *OnPipeOutputFunc_Execute_Call) Run(run func(typ process.OutputType, line []byte, key string)) *OnPipeOutputFunc_Execute_Call {
 	_c.Call.Run(func(args mock.Arguments) {
-		run(args[0].(process.OutputType), args[1].(string), args[2].([]byte))
+		run(args[0].(process.OutputType), args[1].([]byte), args[2].(string))
 	})
 	return _c
 }
@@ -50,7 +50,7 @@ func (_c *OnPipeOutputFunc_Execute_Call) Return() *OnPipeOutputFunc_Execute_Call
 	return _c
 }
 
-func (_c *OnPipeOutputFunc_Execute_Call) RunAndReturn(run func(process.OutputType, string, []byte)) *OnPipeOutputFunc_Execute_Call {
+func (_c *OnPipeOutputFunc_Execute_Call) RunAndReturn(run func(process.OutputType, []byte, string)) *OnPipeOutputFunc_Execute_Call {
 	_c.Run(run)
 	return _c
 }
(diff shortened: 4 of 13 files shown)


contracts/process/schedule.go
```diff
@@ -0,0 +1,44 @@
...
+	// This allows a strategy to prioritize tasks with shorter or more urgent deadlines.
+	// A time.Duration of zero should be interpreted by strategies as an infinite
+	// timeout, effectively giving it the lowest possible priority in a timeout-based sort.
+	GetTimeout() time.Duration
+
+	// GetPriority returns the configured priority level for the command.
+	// Strategies use this to ensure that more important tasks are executed before
+	// less important ones, regardless of their timeout. A higher numerical value
+	// typically indicates a higher execution priority.
+	GetPriority() Priority
```
hwbrzzl:
```suggestion
	Key() string

	// GetTimeout returns the configured maximum execution duration for the command.
	// This allows a strategy to prioritize tasks with shorter or more urgent deadlines.
	// A time.Duration of zero should be interpreted by strategies as an infinite
	// timeout, effectively giving it the lowest possible priority in a timeout-based sort.
	Timeout() time.Duration

	// GetPriority returns the configured priority level for the command.
	// Strategies use this to ensure that more important tasks are executed before
	// less important ones, regardless of their timeout. A higher numerical value
	// typically indicates a higher execution priority.
	Priority() Priority
```

contracts/process/pool.go
```diff
@@ -0,0 +1,83 @@
...
+import (
+	"context"
+	"io"
+	"time"
+)
+
+type OnPoolOutputFunc func(key string, typ OutputType, line []byte)
+
+// PoolBuilder defines the interface for configuring and launching a pool of concurrent processes.
+type PoolBuilder interface {
```
hwbrzzl, line 12: Could you add an example for these new functions?

contracts/process/pool.go
```diff
@@ -0,0 +1,83 @@
...
+	Start(builder func(Pool)) (RunningPool, error)
+}
+
+type Pool interface {
+	Command(name string, arg ...string) PoolCommand
+}
+
+// PoolCommand provides a builder interface for a single command within a pool.
+// It must satisfy the Schedulable interface to be used by scheduling strategies.
+type PoolCommand interface {
```
hwbrzzl, line 44: Ditto

contracts/process/schedule.go
krishankumar01: Yes, I’ll change it. The `Get` prefix won’t be required if we use the `With` prefix for setters.

contracts/process/pool.go
krishankumar01: Yes, it’s still not finished. I’ll add an example after it’s done.

process/pool.go
```diff
@@ -0,0 +1,255 @@
...
+			cancel()
+		}
+		close(done)
+	}()
+
+	return NewRunningPool(runningProcesses, keys, cancel, results, done), nil
+}
+
+type Pool struct {
+	steps []*PoolCommand
```
hwbrzzl:
```suggestion
	commands []*PoolCommand
```

process/pool.go
```diff
@@ -0,0 +1,255 @@
...
+	ctx := r.ctx
+	var cancel context.CancelFunc
+	if r.timeout > 0 {
+		ctx, cancel = context.WithTimeout(ctx, r.timeout)
+	}
+
+	concurrency := r.concurrency
+	if concurrency <= 0 || concurrency > len(steps) {
+		concurrency = len(steps)
+	}
```
hwbrzzl: If `concurrency` is not set, will every command be run at the same time instead of one by one?

process/pool.go
```diff
@@ -0,0 +1,255 @@
...
+
+func (r *PoolBuilder) run(configure func(pool contractsprocess.Pool)) (map[string]contractsprocess.Result, error) {
+	run, err := r.start(configure)
+	if err != nil {
+		return nil, err
+	}
+	return run.Wait(), nil
+}
+
+func (r *PoolBuilder) start(configure func(contractsprocess.Pool)) (contractsprocess.RunningPool, error) {
```
hwbrzzl, line 91: Could you add some annotations for this function logic?

contracts/process/pool.go
```diff
@@ -0,0 +1,74 @@
...
+	// processes in the pool.
+	OnOutput(handler OnPoolOutputFunc) PoolBuilder
+
+	// Run starts the pool, waits for all processes to complete, and returns a
+	// map of the results, keyed by the process keys.
+	Run(builder func(Pool)) (map[string]Result, error)
+
+	// Start launches the pool asynchronously and returns a handle to the running
+	// pool, allowing for interaction with the live processes.
+	Start(builder func(Pool)) (RunningPool, error)
```
hwbrzzl, line 27: Can they be implemented like below?

Image

process/running_pipe.go
```diff
@@ -14,7 +14,7 @@ import (
 
 type RunningPipe struct {
 	commands []*exec.Cmd
-	steps    []*Step
+	steps    []*PipeCommand
```
hwbrzzl:
```suggestion
	commands    []*PipeCommand
```

hwbrzzl (commented):
Great, only a few nitpicks.

contracts/process/pool.go
krishankumar01: That would be a bit complicated and not really necessary. We could add a `Pool` helper and allow passing `OnOutput` directly in the `Start` method, but since most of our existing implementations already use an `OnOutput` callback, it’s better to stay consistent with that approach.

Below is a tidy sample that implements the PHP example in our style:
```go
package main

func main() {
	running, err := process.NewPool().
		OnOutput(func(key string, typ process.OutputType, line []byte) {
			// ...
		}).
		Start(func(p process.Pool) {
			p.Command("bash", "import-1.sh").Path("./")
			p.Command("bash", "import-2.sh").Path("./")
			p.Command("bash", "import-3.sh").Path("./")
		})

	if err != nil {
		os.Exit(1)
	}

	<-running.Done()

	results := running.Wait()
	_ = results
}
```

process/pool.go
krishankumar01: Added

contracts/process/pool.go
hwbrzzl: `OnOutput` is fine, can it be?

```
package main

func main() {
	running, err := facades.Process.Pool(func(p process.Pool) {
			p.Command("bash", "import-1.sh").Path("./")
			p.Command("bash", "import-2.sh").Path("./")
			p.Command("bash", "import-3.sh").Path("./")
		}).
		OnOutput(func(typ process.OutputType, line []byte, key string) {
			// ...
		}).
		Start()

	if err != nil {
		os.Exit(1)
	}

	<-running.Done()

	results := running.Wait()
	_ = results
}
```

hwbrzzl: nit: I updated the sort of OnOutput `func(typ process.OutputType, line []byte, key string)` to be consistent with Laravel.

krishankumar01: Yeah, I’ll think about refactoring it from a facade perspective. Still not sure how it’ll be used with facades though. I’ll raise a new PR for that and include the similar contract changes there itself.

process/pipeline.go
```diff
@@ -101,7 +101,7 @@ func (r *Pipeline) start(configure func(contractsprocess.Pipe)) (contractsproces
 	pipe := &Pipe{}
 	configure(pipe)
 
-	steps := pipe.steps
+	steps := pipe.commands
```
hwbrzzl:
```suggestion
	commands := pipe.commands
```

contracts/process/pool.go
hwbrzzl: Thanks, it can be optimized in another PR if you want.

hwbrzzl (dismissed):
Amazing 👍

hwbrzzl (approved):
Thanks

PR #1217, opened 2025-10-03
chore: add `With` prefix to process setter methods

krishankumar01:
## 📑 Description

Related PRs: #1171, #1178

Add the `With` prefix to all setter methods in process and pipeline. In Pool, I also need getters, so having a method named `Timeout` as a setter is confusing. To stay consistent, I am renaming it (and others) to `With{Property}`(`WithTimeout`), since in the Go community setter methods are commonly named with a `With` prefix.

## ✅ Checks

### DIFF
--- a/process/pipeline_windows_test.go
+++ b/process/pipeline_windows_test.go
@@ -114,7 +114,7 @@ func TestPipe_WithContext_Windows(t *testing.T) {
 }
 
 func TestPipe_DefaultStepKeys_Windows(t *testing.T) {
-	rp, err := NewPipe().Quietly().Start(func(b contractsprocess.Pipe) {
+	rp, err := NewPipe().WithQuiet().Start(func(b contractsprocess.Pipe) {
 		b.Command("cmd", "/C", "(echo a & echo b)")
 		b.Command("cmd", "/C", "more")
 	})


krishankumar01:
Yeah, Laravel doesn’t use the `With` prefix, but in our case it is required. Not sure if that’s an issue since the naming is mostly the same, just adding a `With`/`Set` prefix to it (`timeout` in Laravel → `WithTimeout` in Goravel, `input` → `WithInput`). Else, for pool strategy (Laravel doesn’t provide this), if I add a getter I’ll need to add the `Get` prefix; otherwise having both `GetTimeout` and `Timeout` (setter) doesn’t give a semantic correctness vibe. Donno, feels like the `With`/`Set` prefix is more natural to use.

> Suggesting `Set` here because that is more natural to use when your setter is a mutating function.

hwbrzzl:
@kkumar-gcc Are the `Get*` functions required? If not, they can not be added I think.

krishankumar01:
@hwbrzzl Yes, I want to provide functionality that allows users to customize the process execution order, similar to OS task scheduling. I’d encourage having it, but if you think it’s unnecessary and a normal FIFO strategy is sufficient, we don’t have to implement it.

hwbrzzl:
`customize the process execution order` ? For example?

krishankumar01:
You might need to customize process execution order to prioritize important or time-sensitive tasks(processes) instead of just running them FIFO.
For Example: In a process pool with tasks A, B, C, you may want to run B first because it’s urgent, even if it was added after A:

```
# Custom order: B → A → C
custom_order = [1, 0, 2]
```

This lets users control execution like an OS scheduler.

hwbrzzl:
> For Example: In a process pool with tasks A, B, C, you may want to run B first because it’s urgent, even if it was added after A:

If so, why not set B, A, C directly?

hwbrzzl:
Close? Sorry for this :pensive:

kkumar-gcc:
> > For Example: In a process pool with tasks A, B, C, you may want to run B first because it’s urgent, even if it was added after A:
> 
> 
> 
> If so, why not set B, A, C directly?

This is useful when task priorities are unknown in advance or can change at runtime, allowing the system to dynamically schedule tasks based on urgency or other criteria.

hwbrzzl:
@kkumar-gcc Em, it may be unnecessary for now. Feel free to create an issue first if you think it's needed in the future.

PR #1219, opened 2025-10-03
feat: add helper methods for console

zoryamba:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/729

Added some helper methods requested in the issue

Screenshot 2025-10-04 at 19 37 33

## ✅ Checks

### DIFF
--- a/console/cli_context.go
+++ b/console/cli_context.go
@@ -485,7 +485,7 @@ func (r *CliContext) Divider(filler ...string) {
 
 	repeat, reminder := width/charsLen, width%charsLen
 
-	r.Line(strings.Repeat(string(chars), repeat) + string(chars[0:reminder]))
+	r.Line(strings.Repeat(string(chars), repeat) + string(chars[:reminder]))
 }
 
 func (r *CliContext) Green(message string) {
--- a/console/cli_context.go
+++ b/console/cli_context.go
@@ -483,7 +483,7 @@ func (r *CliContext) Green(message string) {
 	color.Green().Print(message)
 }
 
-func (r *CliContext) GreenLn(message string) {
+func (r *CliContext) Greenln(message string) {
 	color.Green().Println(message)
 }
 
@@ -491,7 +491,7 @@ func (r *CliContext) Red(message string) {
 	color.Red().Print(message)
 }
 
-func (r *CliContext) RedLn(message string) {
+func (r *CliContext) Redln(message string) {
 	color.Red().Println(message)
 }
 
@@ -499,7 +499,7 @@ func (r *CliContext) Yellow(message string) {
 	color.Yellow().Print(message)
 }
 
-func (r *CliContext) YellowLn(message string) {
+func (r *CliContext) Yellowln(message string) {
 	color.Yellow().Println(message)
 }
 
@@ -507,6 +507,6 @@ func (r *CliContext) Black(message string) {
 	color.Black().Print(message)
 }
 
-func (r *CliContext) BlackLn(message string) {
+func (r *CliContext) Blackln(message string) {
 	color.Black().Println(message)
 }
--- a/console/cli_context.go
+++ b/console/cli_context.go
@@ -1,11 +1,13 @@
 package console
 
 import (
+	"strings"
 	"time"
 
 	"github.com/charmbracelet/huh"
 	"github.com/charmbracelet/huh/spinner"
 	"github.com/charmbracelet/lipgloss"
+	"github.com/pterm/pterm"
 	"github.com/urfave/cli/v3"
 
 	"github.com/goravel/framework/contracts/console"
@@ -468,4 +470,43 @@ func (r *CliContext) WithProgressBar(items []any, callback func(any) error) ([]a
 
 func (r *CliContext) TwoColumnDetail(first, second string, filler ...rune) {
 	r.Line(supportconsole.TwoColumnDetail(first, second, filler...))
+}
+
+func (r *CliContext) Divider(filler ...rune) {
+	width := pterm.GetTerminalWidth()
+	char := string(append(filler, '-')[0])
+
+	r.Line(strings.Repeat(char, width))
+}
+
+func (r *CliContext) Green(message string) {
+	color.Green().Print(message)
+}
+
+func (r *CliContext) GreenLn(message string) {
+	color.Green().Println(message)
+}
+
+func (r *CliContext) Red(message string) {
+	color.Red().Print(message)
+}
+
+func (r *CliContext) RedLn(message string) {
+	color.Red().Println(message)
+}
+
+func (r *CliContext) Yellow(message string) {
+	color.Yellow().Print(message)
+}
+
+func (r *CliContext) YellowLn(message string) {
+	color.Yellow().Println(message)
+}
+
+func (r *CliContext) Black(message string) {
+	color.Black().Print(message)
+}
+
+func (r *CliContext) BlackLn(message string) {
+	color.Black().Println(message)
 }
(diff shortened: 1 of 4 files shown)


contracts/console/command.go
```diff
@@ -129,6 +129,24 @@ type Context interface {
 	WithProgressBar(items []any, callback func(any) error) ([]any, error)
 	// TwoColumnDetail writes a two column detail to the console.
 	TwoColumnDetail(first, second string, filler ...rune)
+	// Divider shows a terminal-width divider filled with given rune
+	Divider(filler ...rune)
+	// Green writes green text to console
+	Green(message string)
+	// GreenLn writes green line to console
+	GreenLn(message string)
```
krishankumar01: Please use lower case `L` here (same for all) ie `Greenln`.

console/cli_context.go
```diff
@@ -469,3 +471,42 @@ func (r *CliContext) WithProgressBar(items []any, callback func(any) error) ([]a
 func (r *CliContext) TwoColumnDetail(first, second string, filler ...rune) {
 	r.Line(supportconsole.TwoColumnDetail(first, second, filler...))
 }
+
+func (r *CliContext) Divider(filler ...rune) {
```
krishankumar01: Maybe we should accept a string here instead of rune. That way the user can pass multi-character fillers too, e.g.:
```bash
Divider("->")
// ->->->->->->->->->->->->->->->->->->->->->->->->->->->->->
```

contracts/console/command.go
```diff
@@ -129,6 +129,24 @@ type Context interface {
 	WithProgressBar(items []any, callback func(any) error) ([]any, error)
 	// TwoColumnDetail writes a two column detail to the console.
 	TwoColumnDetail(first, second string, filler ...rune)
+	// Divider shows a terminal-width divider filled with given rune
+	Divider(filler ...rune)
+	// Green writes green text to console
+	Green(message string)
```
krishankumar01, line 135: To write test cases for the new color methods, you can refer to `console/cli_helper_test.TestShowCommandHelp_HelpPrinterCustom`.

https://github.com/goravel/framework/blob/5eab81f111cf32ad9fb1e575ea429c9451757aae/console/cli_helper_test.go#L15

console/cli_context.go
zoryamba: Was referring to this method for consistency. https://github.com/goravel/framework/blob/5eab81f111cf32ad9fb1e575ea429c9451757aae/support/console/console.go#L133 

But can implement This one if approved

hwbrzzl: Yes, good point. We can use `string`.

hwbrzzl (commented):
Nice, thanks 👍

contracts/console/command.go
zoryamba: Added test cases. Thanks

console/cli_context.go
```diff
@@ -469,3 +471,51 @@ func (r *CliContext) WithProgressBar(items []any, callback func(any) error) ([]a
 func (r *CliContext) TwoColumnDetail(first, second string, filler ...rune) {
 	r.Line(supportconsole.TwoColumnDetail(first, second, filler...))
 }
+
+func (r *CliContext) Divider(filler ...string) {
+	var chars []rune
```
krishankumar01: This can directly be a string; using `[]rune` isn’t necessary since it’s converted back to a string anyway when calling `strings.Repeat`.

console/cli_context.go
```diff
@@ -469,3 +471,51 @@ func (r *CliContext) WithProgressBar(items []any, callback func(any) error) ([]a
...
+	} else {
+		chars = []rune(filler[0])
+	}
+
+	width := pterm.GetTerminalWidth()
+	charsLen := len(chars)
+
+	repeat, reminder := width/charsLen, width%charsLen
+
+	r.Line(strings.Repeat(string(chars), repeat) + string(chars[0:reminder]))
```
krishankumar01:
```suggestion
        r.Line(strings.Repeat(string(chars), repeat) + string(chars[:reminder]))
```

console/cli_context_test.go
```diff
@@ -2,9 +2,13 @@ package console
 
 import (
 	"context"
+	"io"
+	"strings"
 	"testing"
 	"time"
 
+	"github.com/goravel/framework/support/color"
```
krishankumar01: We maintain the following import order (each section separated by a newline):
- Standard library packages
- Third-party packages
- Framework internal packages

```suggestion
	"github.com/pterm/pterm"
	"github.com/stretchr/testify/assert"
	"github.com/urfave/cli/v3"
	
        "github.com/goravel/framework/support/color"
)
```

console/cli_context_test.go
```diff
@@ -427,3 +431,145 @@ func TestCliContextArguments(t *testing.T) {
...
+			},
+			expectedOutput: color.Default().Sprintln(strings.Repeat("=", pterm.GetTerminalWidth())),
+		},
+		{
+			name: "test Divider multiple",
+			testFunc: func(ctx *CliContext) {
+				ctx.Divider("=->")
+			},
+			expectedOutput: color.Default().Sprintln(
+				strings.Repeat("=->", pterm.GetTerminalWidth()/3) + "=->"[0:pterm.GetTerminalWidth()%3],
```
krishankumar01: We can simplify this by forcing the terminal size, then expected output can be hardcoded, making the test easier to read.

```go
// Force terminal width for deterministic tests
pterm.SetForcedTerminalSize(40, 20)

ctx.Divider("=->")
expected := color.Default().Sprintln("=->=->=->=->=->=->=->=->=->=->=->=->=->=->")
got := color.CaptureOutput(func(io.Writer) { ctx.Divider("=->") })
assert.Equal(t, expected, got)
```

console/cli_context_test.go
```diff
@@ -427,3 +431,145 @@ func TestCliContextArguments(t *testing.T) {
...
+	}
+
+	for _, tt := range testCases {
+		t.Run(tt.name, func(t *testing.T) {
+			ctx := CliContext{}
+			got := color.CaptureOutput(func(io.Writer) {
+				tt.testFunc(&ctx)
+			})
+
+			assert.Equal(t, got, tt.expectedOutput)
```
krishankumar01: The arguments are reversed here, usually it’s `assert.Equal(t, expected, actual)` for clarity.
```suggestion
                        assert.Equal(t, tt.expectedOutput, got)
```

console/cli_context_test.go
```diff
@@ -427,3 +431,145 @@ func TestCliContextArguments(t *testing.T) {
...
```
krishankumar01: ditto

kkumar-gcc (commented):
Test cases look good, just a couple of nitpicks.

console/cli_context.go
zoryamba: It's required for multibyte strings... While I can get the string length in chars with `utf8.RuneCountInString`, I can't see any function to slice it with remainder... Slicing the string cuts the bytes, not chars:

Screenshot 2025-10-05 at 13 05 34

zoryamba: As an option, I can convert it only if remainder > 0, to avoid converting for simple single char dividers, which seems to be the most common case

zoryamba: Please, take a look

krishankumar01: Hmm, sorry, you’re right, I didn’t think about non-ASCII chars. In that case, using []rune is fine too 👍. We can also stick with your current implementation, where we convert to runes only when handling the remainder.

hwbrzzl:
Thanks, added you to the contributor list.

PR #1442, opened 2026-04-13
test(orm): cover Select(Associations).Update many-to-many behavior

hwbrzzl:
## Summary
- Add coverage for `Select(orm.Associations).Update(&user)` on many-to-many relations to verify new relations are appended while previously linked relations are not removed.
- Verify selected association updates still persist the model's main fields and that update calls without `Select(orm.Associations)` do not mutate many-to-many links.
- Refresh indirect dependencies in `tests/go.mod` and `tests/go.sum` to align the tests module with the current dependency graph.

## Why
This branch documents and locks down expected ORM behavior that raised confusion: updating with `Select(orm.Associations)` should update the model fields and sync in new associations, but it should not delete existing many-to-many links that are missing from the in-memory slice. The new tests make that contract explicit so future refactors do not accidentally change it.

The test suite now also includes a contrast case without `Select(orm.Associations)` to make the association-update boundary clear and prevent regressions around relation side effects.

```go
user := &models.User{
	Name: "bowen",
	Roles: []*models.Role{{Name: "admin"}, {Name: "editor"}},
}
_ = facades.Orm().Query().Select(orm.Associations).Create(user)

// Before: only role "owner" is kept in memory, but existing pivot rows should not be removed.
user.Roles = []*models.Role{{Name: "owner"}}
_, _ = facades.Orm().Query().Model(user).Select(orm.Associations).Update(user)

// After: user name fields are updated, and role "owner" is appended;
// existing "admin"/"editor" links remain associated.
```

### DIFF
--- a/tests/query_test.go
+++ b/tests/query_test.go
@@ -13,7 +13,7 @@ import (
 	contractsorm "github.com/goravel/framework/contracts/database/orm"
 	databasedb "github.com/goravel/framework/database/db"
 	"github.com/goravel/framework/database/gorm"
-	orm "github.com/goravel/framework/database/orm"
+	"github.com/goravel/framework/database/orm"
 	"github.com/goravel/framework/errors"
 	"github.com/goravel/framework/support/carbon"
 	"github.com/goravel/framework/support/convert"
--- a/tests/query_test.go
+++ b/tests/query_test.go
@@ -3298,23 +3298,22 @@ func (s *QueryTestSuite) TestManyToManyUpdateWithAssociations() {
 					},
 				},
 				{
-					name: "Select(*, Associations).Update updates main fields and associations",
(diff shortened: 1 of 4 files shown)


tests/query_test.go
```diff
@@ -10,10 +10,10 @@ import (
 
 	"github.com/google/uuid"
 
-	"github.com/goravel/framework/contracts/database/orm"
 	contractsorm "github.com/goravel/framework/contracts/database/orm"
 	databasedb "github.com/goravel/framework/database/db"
 	"github.com/goravel/framework/database/gorm"
+	orm "github.com/goravel/framework/database/orm"
```
hwbrzzl:
```suggestion
	"github.com/goravel/framework/database/orm"
```

PR #1443, opened 2026-04-14
feat(ai): add tool calling support

### LINKED ISSUE #917: [Feature] AI SDK Phase 3: Advanced Interactions (Tools & Structured Output)
### Repository: goravel/framework

**Requirements:**
- Implement `Tool` registration and execution logic (Function Calling).
- Implement `Structured Output` support to allow models to return data that maps directly to Go structs.
- Implement `Streaming` support using channels or callbacks for real-time (SSE) responses.
- Implement `Middleware` support to intercept and modify prompts or responses.

hwbrzzl:
## Summary
- Add `Tool` interface and `ToolCall` type to `contracts/ai` so agents can expose callable capabilities to the model
- `Conversation.Prompt()` now executes a tool-call/re-prompt loop (up to `MaxToolCallIterations = 10`) when the model requests tool invocations
- OpenAI provider serialises tool definitions into API requests and deserialises tool-call responses back into `ToolCall` values

Closes https://github.com/goravel/goravel/issues/917

## Why
Goravel's AI module previously had no mechanism for models to call external functions during a conversation. Without tool calling, agents cannot retrieve live data, perform side-effects, or compose multi-step reasoning across external systems.

The new `Tool` interface (`Name`, `Description`, `Parameters`, `Execute`) lets any agent expose callable capabilities. `Conversation.Prompt()` drives a loop: after each model response it checks for `ToolCalls`, dispatches them, appends the results to conversation history, and re-prompts — stopping only when the model returns plain text or the iteration guard fires.

```go
// Define a tool
type WeatherTool struct{}

func (t *WeatherTool) Name() string        { return "get_weather" }
func (t *WeatherTool) Description() string { return "Returns current weather for a city." }
func (t *WeatherTool) Parameters() map[string]any {
    return map[string]any{
        "type": "object",
        "properties": map[string]any{
            "city": map[string]any{"type": "string"},
        },
        "required": []string{"city"},
    }
}
func (t *WeatherTool) Execute(_ context.Context, args map[string]any) (string, error) {
    city, _ := args["city"].(string)
    return fmt.Sprintf("Sunny, 25°C in %s", city), nil
}

// Register the tool on the agent and prompt
agent := ai.NewAgent(ai.AgentConfig{
    Tools: []contractsai.Tool{&WeatherTool{}},
})
conv := facades.AI().Agent(agent)
resp, err := conv.Prompt("What's the weather in London?")
// The model calls get_weather, the framework executes it, and the model
// returns: "It is currently sunny and 25°C in London."
fmt.Println(resp.Text())
```


contracts/ai/provider.go
```diff
@@ -8,6 +8,8 @@ type AgentPrompt struct {
 	Agent Agent
 	Input string
 	Model string
+	// Tools lists the tools available to the model for this request.
+	Tools []Tool
```
hwbrzzl, line 12: Agent contains tools. Why define it here?

contracts/ai/message.go
```diff
@@ -13,4 +13,10 @@ const (
 type Message struct {
 	Role    MessageRole
 	Content string
+	// ToolCallID is set on RoleToolResult messages to correlate a tool
+	// result with the assistant ToolCall that requested it.
+	ToolCallID string
+	// ToolCalls is set on RoleAssistant messages when the model requests
+	// one or more tool invocations instead of (or before) returning text.
+	ToolCalls []ToolCall
```
hwbrzzl, line 21: Why set `ToolCallID` and `ToolCalls` simultaneously? It's hard to understand.

ai/openai/provider.go
```diff
@@ -156,10 +171,76 @@ func (r *Provider) buildMessages(prompt contractsai.AgentPrompt) []goopenai.Chat
...
+				}
+				messages = append(messages, goopenai.ChatCompletionMessageParamUnion{OfAssistant: &assistant})
+			} else {
+				messages = append(messages, goopenai.AssistantMessage(m.Content))
+			}
+		case contractsai.RoleToolResult:
+			messages = append(messages, goopenai.ToolMessage(m.Content, m.ToolCallID))
 		}
 	}
 	messages = append(messages, goopenai.UserMessage(prompt.Input))
```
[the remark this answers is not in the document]
hwbrzzl: Fixed: `buildMessages` now only appends `prompt.Input` when it is non-empty. Continuation calls (after tool results) pass `Input: ""` so no empty user message is sent to the API.

ai/openai/provider.go
```diff
@@ -156,10 +171,76 @@ func (r *Provider) buildMessages(prompt contractsai.AgentPrompt) []goopenai.Chat
...
+	for _, tc := range raw {
+		if tc.Type != "function" {
+			continue
+		}
+		fn := tc.Function
+		var args map[string]any
+		if fn.Arguments != "" {
+			// Best-effort decode; invalid JSON is surfaced as an empty map.
+			_ = json.Unmarshal([]byte(fn.Arguments), &args)
+		}
```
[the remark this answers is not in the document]
hwbrzzl: Fixed: `args` is now initialized to `make(map[string]any)` before the unmarshal attempt. If `json.Unmarshal` fails the call still gets a consistent, non-nil empty map rather than nil.

contracts/ai/agent.go
```diff
@@ -6,4 +6,7 @@ type Agent interface {
 	Instructions() string
 	// Messages returns prior conversation messages to include as context.
 	Messages() []Message
```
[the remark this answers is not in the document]
hwbrzzl: The `ai` package is new and has not been included in any released version of the framework yet, so there are no existing downstream implementations of this interface. Adding `Tools()` is not a breaking change in practice. We would rather keep all agent capabilities behind a single `Agent` interface than scatter them across optional sub-interfaces that callers have to type-assert.

contracts/ai/response.go
```diff
@@ -4,6 +4,9 @@ package ai
 type Response interface {
 	Text() string
 	Usage() Usage
+	// ToolCalls returns any tool invocations the model requested.
+	// Returns nil or an empty slice when the model returns plain text.
+	ToolCalls() []ToolCall
 }
```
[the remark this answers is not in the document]
hwbrzzl: Same reasoning as for `Agent.Tools()`: the `ai` package has not been released yet, so no existing `Response` implementations exist outside this repository. `ToolCalls()` is core to the tool-calling contract and belongs on the primary interface.

ai/conversation.go
```diff
@@ -36,23 +42,87 @@ func (r *conversation) Messages() []contractsai.Message {
...
 func (r *conversation) Prompt(input string) (contractsai.Response, error) {
-	resp, err := r.provider.Prompt(r.ctx, contractsai.AgentPrompt{
+	tools := r.agent.Tools()
+
+	// Snapshot the message length before we touch anything so we can roll back
+	// cleanly on any error path.
+	r.mu.Lock()
+	snapshot := len(r.messages)
+	r.messages = append(r.messages, contractsai.Message{Role: contractsai.RoleUser, Content: input})
+	r.mu.Unlock()
```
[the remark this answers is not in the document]
hwbrzzl: Fixed: `Prompt` no longer pre-appends the user message to `r.messages` before calling the provider. Instead it builds an isolated working copy and exposes it through a `pending` field that `Messages()` returns during the call. The user message (and any tool-call/result messages) are committed to `r.messages` only after the entire loop succeeds, so the provider never receives the same message from both `Agent.Messages()` and `AgentPrompt.Input`.

ai/conversation.go
```diff
@@ -36,23 +42,87 @@ func (r *conversation) Messages() []contractsai.Message {
...
+
+	rollback := func() {
+		r.mu.Lock()
+		r.messages = r.messages[:snapshot]
+		r.mu.Unlock()
+	}
```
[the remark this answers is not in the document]
hwbrzzl: Fixed: added `promptMu sync.Mutex` to `conversation`. `Prompt` acquires this mutex for the duration of the entire call (including all tool-call iterations). Concurrent `Prompt` calls are now serialized, so a rollback (or commit) in one call cannot affect history being built by another.

ai/conversation.go
```diff
@@ -63,6 +133,7 @@ func (r *conversation) Stream(input string) (contractsai.StreamableResponse, err
 		Agent: r,
 		Input: input,
 		Model: r.model,
+		Tools: r.agent.Tools(),
```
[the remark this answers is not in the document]
hwbrzzl: Fixed: removed `Tools` from `Stream`'s `AgentPrompt`. The streaming path does not include tool serialization or tool-call handling, so passing tools there was misleading. Tool support in streaming can be added end-to-end in a follow-up.

contracts/ai/provider.go
hwbrzzl: The `Tools` field in `AgentPrompt` is a convenience snapshot for the provider. While `Agent.Tools()` is available, passing tools explicitly in the prompt keeps `AgentPrompt` self-contained: the provider has everything it needs without having to call back into the agent. It also makes it straightforward to override or filter tools on a per-prompt basis in the future.

contracts/ai/message.go
hwbrzzl: They are set on different message roles and are never both non-zero on the same message:\n\n- `ToolCalls` is set on `RoleAssistant` messages when the model requests one or more tool invocations.\n- `ToolCallID` is set on `RoleToolResult` messages to correlate a tool result with the assistant call that requested it.\n\nA single `Message` struct covers both roles to avoid having multiple nearly-identical types.

ai/conversation.go
```diff
@@ -33,26 +46,121 @@ func (r *conversation) Messages() []contractsai.Message {
...
+
+		toolCalls := resp.ToolCalls()
+		if len(toolCalls) == 0 {
+			// Plain text response — we're done.
+			break
+		}
+
+		if i == MaxToolCallIterations-1 {
+			clearPending()
+			return nil, fmt.Errorf("ai: tool call loop exceeded %d iterations", MaxToolCallIterations)
```
hwbrzzl: The error should be moved to the `errors/list.go` file.

ai/conversation.go
```diff
@@ -33,26 +46,121 @@ func (r *conversation) Messages() []contractsai.Message {
...
-		return nil, err
+		Tools: tools,
+	}
+
+	var (
+		resp contractsai.Response
+		err  error
+	)
+
+	for i := range MaxToolCallIterations {
```
hwbrzzl, line 98: If there are many different tools and the model needs to call all of them, it's very easy to exceed the max limit, right?

ai/conversation.go
```diff
@@ -85,3 +193,34 @@ func (r *conversation) Reset() {
...
+	results := make([]contractsai.Message, 0, len(calls))
+	for _, call := range calls {
+		tool, ok := index[call.Name]
+		if !ok {
+			return nil, fmt.Errorf("ai: tool %q not found", call.Name)
+		}
+
+		result, err := tool.Execute(r.ctx, call.Args)
+		if err != nil {
+			return nil, fmt.Errorf("ai: tool %q execution failed: %w", call.Name, err)
```
hwbrzzl: Ditto

ai/conversation.go
```diff
@@ -85,3 +193,34 @@ func (r *conversation) Reset() {
...
+	index := make(map[string]contractsai.Tool, len(tools))
+	for _, t := range tools {
+		index[t.Name()] = t
+	}
+
+	results := make([]contractsai.Message, 0, len(calls))
+	for _, call := range calls {
+		tool, ok := index[call.Name]
+		if !ok {
+			return nil, fmt.Errorf("ai: tool %q not found", call.Name)
```
hwbrzzl: Ditto

PR #1447, opened 2026-04-20
fix: [#925] rotate session on login and logout to prevent session fixation

### LINKED ISSUE #925: Session ID not rotated on login/logout (session fixation risk)
## Summary
Session-based auth transitions may not rotate or invalidate the session identifier at login/logout boundaries.

## Impact
Keeping the same session ID across anonymous -> authenticated (or authenticated -> logged-out) state increases session fixation risk.

## Reproduction (example app)
1. Call GET /session/status to obtain initial session cookie.
2. Call POST /session/login with the same cookie jar.
3. Compare session cookie before and after login.
4. Session ID can remain unchanged.

## Expected behavior
On login, session ID should be regenerated (preferably destroying old session). On logout, session should be invalidated.

## Suggested fix
In session guard flows, call Session.Regenerate(true) on login and Session.Invalidate() on logout (or equivalent safe rotation/invalidation sequence).

## Notes
Observed while testing goravel/example on v1.17.x.

krishankumar01:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/925

Session-based auth kept the same session ID across login/logout boundaries, leaving apps open to session fixation.

### What changed

- `SessionGuard.LoginUsingID` now calls `Session.Regenerate(true)` before storing the auth key, destroying the pre-auth session record and issuing a fresh ID (same as Laravel's `SessionGuard::updateSession`).
- `SessionGuard.Logout` forgets the guard's auth key and then rotates with `Regenerate(true)`. It deliberately does **not** use `Invalidate()`: invalidating flushes the whole session, which would also log out other guards sharing the session and wipe unrelated state (flash messages, locale). Forget + destructive regenerate kills the old session ID just the same, without the collateral damage.
- The session cookie is re-issued at rotation time via a shared `session.WriteCookie(ctx, session)` helper (also used by the `StartSession` middleware). This is required for correctness: the middleware writes the cookie **before** `req.Next()`, and on the gin driver the response is rendered inside the innermost handler — by the time middleware code after `Next()` runs, headers are already flushed. Without re-issuing during the handler, the rotated ID would never reach the client and every login would effectively end the session. Browsers take the last `Set-Cookie` for the same name, so the rotated cookie wins over the one the middleware wrote earlier.

@coderabbitai summary

## ✅ Checks

### DIFF
--- a/auth/session_guard.go
+++ b/auth/session_guard.go
@@ -9,7 +9,7 @@ import (
 	"github.com/goravel/framework/contracts/http"
 	contractsession "github.com/goravel/framework/contracts/session"
 	"github.com/goravel/framework/errors"
-	"github.com/goravel/framework/support/carbon"
+	"github.com/goravel/framework/session"
 )
 
 type SessionGuard struct {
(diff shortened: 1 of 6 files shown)


hwbrzzl (approved):
LGTM

PR #1452, opened 2026-04-23
feat: make goravel artisan optional and register use service provider…

### LINKED ISSUE #959: make sure the artisan can be opt in bundle output binary.
### Before feedback (在反馈之前)

### Description
  Currently the console service provider and all default commands (make:*, env:*, test:make,
  package:*, vendor:publish, up/down, etc.) are registered as base, so every binary built from
   a Goravel app ships with the full Artisan CLI 

  For deployments that distribute compiled binaries, this means:

  1. Larger binary size — all command code (urfave/cli, stubs, package install logic) is
  linked in even when never invoked.
  2. Exposed surface — running ./app --help in production reveals dev-only commands
  (make:controller, package:install, env:decrypt, ...) that shouldn't be discoverable on a
  release artifact.
  3. No way to opt out — there is no supported path to compile a "server-only" build today.

###  Goal

  Allow the developer to choose whether Artisan is included at compile/runtime, while keeping
  the dev experience unchanged:

  - Dev: console.ServiceProvider registered → full Artisan with all framework commands
  (today's behavior).
  - Prod (opt-in): developer omits console.ServiceProvider from bootstrap/providers.go → no
  make:* / env:* / test:make / up / down commands compiled into the binary; ./app --help hides
   them.

LinboLen:
… instead

## 📑 Description

Closes https://github.com/goravel/goravel/issues/959

## ✅ Checks


LinboLen:
after merged user must register the `&console.ServiceProvider{}` in bootstrap/providers.go to enable artisan command

contracts/binding/binding.go
```diff
@@ -73,7 +73,7 @@ var (
 		Artisan: {
 			Description: "The CLI tool that comes with Goravel for interacting with the command line.",
 			PkgPath:     "github.com/goravel/framework/console",
-			IsBase:      true,
```
hwbrzzl, line 76: The Artisan facade should be a basic item, given framework needs to provide the `package:install`  command to install facades.

LinboLen: updated

hwbrzzl: Sorry, if so, `console.ServiceProvider` should be registered as default. Hence what's the requirement of this PR?

LinboLen: we can use the artisan in dev working. but make the artisan command opt in prod build to make sure the bundle output bin not use artisan command anymore. because I need distribute the binary simply and smallest also hidden the default command

hwbrzzl: After this PR, the console service provider has to be registered in bootstrap explicitly, it's a breaking change. It's unexpected.

And agree with some of the commands can be moved to the other modules instead of `foundation`, but the `package:*` commands should be registered as default to provide the functionality to install/uninstall pakcages or facades.

About `hidden the default command`, could you confirm if [the code](https://github.com/goravel/installer/blob/master/app/providers/artisan_service_provider.go#L74-L80) meets your requirement?

LinboLen: I make the artisan as IsBase is False. if the user wants to use  install/uninstall he must add &console.ServiceProvider{} in bootstrap/providers.go. requirement need to hidden or replace the artisan command completely. we have difference bundle entry. the lite entry don't need artisan command

LinboLen: after merged I will check the installer, goravel/goravel goravel/goravel-lite package. ensure the &console.ServiceProvider{} defult is imported in bootstrap/providers.go.

hwbrzzl: > we can use the artisan in dev working. but make the artisan command opt in prod build to make sure the bundle output bin not use artisan command anymore.

I think the current PR can't implement the requirement, given that there are many other commands when using other facades. You can't load different commands for dev and prod in this way.

And all users need to register the console service provider manually after this pr, it's expected for now. On the other side, the artisan should be the foundation underlying framework design.

Hence, we may find another way to implement your requirement.

PR #1456, opened 2026-04-28
fix: [#946] stop WithContext from leaking framework context keys

### LINKED ISSUE #946: `facades.Log().WithContext` dumps entire context.Context including internal framework values
### Before feedback (在反馈之前)

### Operating System (操作系统)

MacOS

### Golang Version (Go 版本)

1.26.0

v1.17.2

`facades.Log().WithContext(ctx)` serializes the entire `context.Context` object into the log output, including framework-internal values like `GoravelAuthJwt`, middleware state, etc. This may leak sensitive internal data into logs.

**Current Behavior**

Calling:
```go
facades.Log().WithContext(ctx).Info("request handled")
```

Produces log output containing all values set on the context, including:
- `GoravelAuthJwt` (auth tokens)
- Any other framework-internal or middleware-set context values

**Expected Behavior**

`WithContext` should behave like Laravel's `Log::withContext()` — accepting explicit key-value pairs to add to the log context, not dumping the entire request context.

Laravel reference:
- `Log::withContext(['key' => 'value'])` — adds specific key-value pairs to the current log instance
- `Log::shareContext(['key' => 'value'])` — shares context across all log channels

Neither dumps the raw request context. They only log what the developer explicitly passes.

```go
// N/A
```

krishankumar01:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/946

`facades.Log().WithContext(ctx)` was reflecting over the context and dumping every key/value, leaking `GoravelAuthJwt` and any other framework-internal values into logs. This filters the dump instead of removing it (the propagation use case from #282 still works).

### Behavior

- Default exclude list (always applied): `GoravelAuthJwt`, `goravel_http_client_name`.
- Optional `logging.context.exclude` config (`[]any`) extends it. Not added to the default config stub — opt-in via docs.

### Matching rules

Two matchers run per key:

- **By identity** (`==`): for non-string `[]any` config entries, e.g. `[]any{middleware.PanicTest{}}`.
- **By name**: for default entries and string config entries. Compared against the key's `shortName` first, then its `qualifiedName`.

`shortName(k)` returns the underlying string for string-kind keys and `%T` (e.g. `pkg.Type`) for everything else. `qualifiedName(k)` returns the import-path-qualified type (`github.com/.../pkg.Type`).

### Output labels

- String-kind keys: their string content.
- Struct sentinels: short `pkg.Type` form by default; if two keys would share the same short form (same package name in different modules), only the colliding ones escalate to fully-qualified import paths. Non-colliding keys keep the short form.

### Notes

- Uncomparable user entries (slice/map from a misconfig) are silently skipped instead of panicking.
- Config is read once per handler via `sync.Once`; runtime hot-swaps are not picked up.

## ✅ Checks


log/logger/handler.go
```diff
@@ -39,11 +44,28 @@ func NewIOHandler(w io.Writer, config config.Config, json foundation.Json, level
...
+		return nil
+	}
+	raw := make(map[any]any)
+	getContextValues(ctx, raw)
+	if len(raw) == 0 {
+		return nil
+	}
+	h.excludeOnce.Do(func() {
+		excludeList, _ := h.config.Get("logging.context.exclude", []any{}).([]any)
+		user := cast.ToStringSlice(excludeList)
```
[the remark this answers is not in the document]
hwbrzzl: @krishankumar01 The logic can't handle this case:

```
type userIDKey struct{}

func SetUserID(ctx context.Context, userID string) context.Context {
    return context.WithValue(ctx, userIDKey{}, userID)
}

func GetUserID(ctx context.Context) (string, bool) {
    userID, ok := ctx.Value(userIDKey{}).(string)
    return userID, ok
}
```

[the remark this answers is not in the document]
krishankumar01: @hwbrzzl fixed. Using your example:

```go
type userIDKey struct{}
ctx = context.WithValue(ctx, userIDKey{}, "u-42")
```

logs as:

```
[Context] map[yourpkg.userIDKey:u-42]
```

To exclude, drop the value into config:

```go
"context": map[string]any{
    "exclude": []any{userIDKey{}},
}
```

Strings still work for string keys (`"trace_id"`), structs match by `==`.

### Collisions

If two packages share a name and both define a sentinel (e.g. `app/http/middleware.PanicTest` and `middleware.PanicTest`), only the colliding keys escalate to the full import path:

```
[Context] map[goravel/app/http/middleware.PanicTest:hello goravel/middleware.PanicTest:world2 trace_id:t-1]
```

log/logger/utils_test.go
```diff
@@ -33,3 +37,135 @@ func TestGetContextValues(t *testing.T) {
...
+		},
+		{
+			name: "colliding labels escalate to qualified names",
+			values: map[any]any{
+				"session_id":                  "from-string",
+				collidingNameKey("session_id"): "from-typed",
+			},
+			expect: map[string]any{
+				"string": "from-string",
+				"github.com/goravel/framework/log/logger.collidingNameKey": "from-typed",
```
hwbrzzl: Can it be?


```suggestion
				"session_id": "from-string",
				"logger.collidingNameKey": "from-typed",
```

log/logger/utils_test.go
```diff
@@ -33,3 +37,135 @@ func TestGetContextValues(t *testing.T) {
...
+			expect: map[string]any{"request_id": "req-1"},
+		},
+		{
+			name: "default excludes still apply when keys with the same label collide",
+			values: map[any]any{
+				"locale":                  "en",
+				utilsContextKey("locale"): "fr",
+				"request_id":              "req-1",
+			},
+			expect: map[string]any{"request_id": "req-1"},
```
hwbrzzl, line 133: The string `locale` is excepted as default, not `utilsContextKey("locale")`. Is it acceptable to log  `utilsContextKey("locale")` here?

log/logger/handler.go
```diff
@@ -39,11 +42,27 @@ func NewIOHandler(w io.Writer, config config.Config, json foundation.Json, level
...
+	if ctx == nil {
+		return nil
+	}
+	raw := make(map[any]any)
+	getContextValues(ctx, raw)
+	if len(raw) == 0 {
+		return nil
+	}
+	h.excludeOnce.Do(func() {
+		user, _ := h.config.Get("logging.context.exclude", []any{}).([]any)
```
hwbrzzl:
```suggestion
		userExclude, _ := h.config.Get("logging.context.exclude", []any{}).([]any)
```

log/logger/utils_test.go
krishankumar01: Filter matches by label form, not by exact type. So `utilsContextKey("locale")` produces label `locale` and hits the default. If you'd rather match by exact type, we'd need each framework package to export its sentinel.

hwbrzzl: Sorry, what's the meaning of `each framework package to export its sentinel`? For example?

krishankumar01: If the keys collide, we either expose the key type so values don't collide, or match by the string value. Say I set the locale key as a struct `localeType`.  Since it's internal to the package, you can't pass `[]any{localeType{}}` in the config to exclude it, so you'd have to use the value `"locale"` instead. So either we export `LocalType`, or we keep matching by value.                
That said, in this case I don't think we need to filter `locale` by default. It's small and not noisy.

hwbrzzl: Hence, are `framework.utilsContextKey("locale")` in the configuration and `utilsContextKey("locale")` in the framework internal different?

krishankumar01: Yes, it’s the same value. If the framework exports the type, then creating `framework.utilsContextKey("locale")` in user config gives you the same key as the one used internally. Since both the type and value match, they compare equal and work fine as map keys.

krishankumar01: I tested it with a custom key `PanicTest` in the middleware package and used it as an exclude key in the config.

image
---
image

hwbrzzl (commented):
LGTM, please solve the conflict.

log/logger/utils_test.go
hwbrzzl: Makes sense, thanks.

hwbrzzl (approved):
LGTM

PR #1463, opened 2026-05-04
feat: add query relation

LinboLen:
## Summary
- Add a relation declaration system and relation-aware query APIs for `Has` / `WhereHas`, morph existence checks, aggregate subselects, and framework-managed eager loading across direct, nested, polymorphic, many-to-many, and through relations.
- Add `Relation(...)` and `Related(...)` ORM builders so applications can read and write relationships through a typed API, including association writes, pivot-table operations, `OfMany` helpers, and morph map resolution.
- Expand ORM and integration coverage across supported drivers to verify relation loading, relation existence filtering, pivot writes, through relations, and the new `Relations()` model contract.

Closes #963

## Why
This PR fills in the missing ORM relationship workflow that users expect from Laravel-style APIs: declaring relations in one place, querying parents by related records, eager loading nested relations, and mutating associations without dropping down to raw GORM behavior. The new contracts make relationship behavior explicit at the framework layer instead of relying on scattered struct-tag conventions and custom query code.

```go
type User struct {
	ID    uint
	Name  string
	Books []*Book `gorm:"-"`
	Roles []*Role `gorm:"-"`
}

func (u *User) Relations() map[string]orm.Relation {
	return map[string]orm.Relation{
		"Books": orm.HasMany{Related: &Book{}},
		"Roles": orm.Many2Many{Related: &Role{}, Table: "role_user"},
	}
}

var users []User
err := facades.Orm().Query().
	Has("Books", ">=", 2).
	With("Books").
	WithCount("Roles").
	Get(&users)
```

The broader relation rework also enables relationship writers and polymorphic helpers to share the same metadata pipeline, which is why this branch includes eager-loader internals, pivot query support, morph maps, and new integration tests. That combination is what makes features like `WhereHas`, `HasMorph`, `With("Books.Author")`, `Relation(parent, "Roles")`, and through-relation queries behave consistently across drivers.


hwbrzzl:
Hey @LinboLen Is the PR ready to be reviewed, please? And FYI, CI failed.

LinboLen:
is tested pass on my local dev environment. currently is finished state. I will fix the ci failed. this pr not use gorm associate any more. use `Related` for relation query, `Relation` for relation modify.

database/gorm/new_relation.go
```diff
@@ -0,0 +1,243 @@
...
+	return r.newRelationQuery(parent, relation)
+}
+
+// newRelationQuery returns a fresh Query pre-scoped to the related rows for parent.relation.
+// Mirrors fedaco's model.NewRelation('foo') for the read path. Caller can chain Where / OrderBy
+// / Get / etc. on the returned Query. Write operations live on RelationWriter (see Query.Relation).
+//
+// Public entry: Orm.Related delegates here on a fresh-session *Query.
+func (r *Query) newRelationQuery(parent any, relation string) contractsorm.Query {
+	if !isValidParent(parent) {
```
LinboLen, line 24: notice. the parent must be a pointer struct.

LinboLen:
fixed all test case. and removed association code in query. the codecov is low. adding more test case

LinboLen:
some of api are real database read and write. already covered in another test case. so the codecov is down for now and can't add this coverage test

hwbrzzl:
So sorry, I missed the messages. I will review the PR.

contracts/database/orm/orm.go
```diff
@@ -36,8 +57,18 @@ type Orm interface {
 }
 
 type Query interface {
-	// Association gets an association instance by name.
-	Association(association string) Association
```
hwbrzzl, line 40: Will Relation replace this function? It's a break change.

tests/models.go
```diff
@@ -48,6 +48,29 @@ func (r *User) Factory() factory.Factory {
 	return &UserFactory{}
 }
 
+func (r *User) Relations() map[string]contractsorm.Relation {
```
hwbrzzl, line 51: Is this function required in the new implementation?

tests/models.go
```diff
@@ -35,11 +35,11 @@ type User struct {
 	Name    string
 	Bio     *string
 	Avatar  string
-	Address *Address
-	Books   []*Book
-	House   *House   `gorm:"polymorphic:Houseable"`
-	Phones  []*Phone `gorm:"polymorphic:Phoneable"`
-	Roles   []*Role  `gorm:"many2many:role_user"`
+	Address *Address `gorm:"-"`
```
hwbrzzl, line 38: Is `gorm:"-"` required when defining the relation?

tests/queries_relationships_test.go
```diff
@@ -0,0 +1,567 @@
...
+			s.Equal("rel_wh_match", users[0].Name)
+		})
+	}
+}
+
+func (s *QueriesRelationshipsTestSuite) TestHas_BelongsTo() {
+	for driver, query := range s.queries {
+		s.Run(driver, func() {
+			user := &User{Name: "rel_bt_user"}
+			s.Nil(query.Query().Create(&user))
```
hwbrzzl: &user == `&&User{Name: "rel_bt_user"}`, is it expected?

tests/queries_relationships_test.go
```diff
@@ -0,0 +1,567 @@
...
+		s.Run(driver, func() {
+			withHouse := &User{Name: "rel_hm_with"}
+			s.Nil(query.Query().Create(&withHouse))
+			s.Nil(query.Query().Relation(withHouse, "House").Save(&House{Name: "rel_hm_house"}))
+			noHouse := &User{Name: "rel_hm_no"}
+			s.Nil(query.Query().Create(&noHouse))
+
+			rq := rq(query.Query())
+			var users []User
+			s.Nil(rq.HasMorph("House", []any{&User{}}).
```
hwbrzzl, line 178: Why is `[]any{&User{}}` needed?

tests/queries_relationships_test.go
```diff
@@ -0,0 +1,567 @@
...
+	}
+	var users []User
+	sql := rq.WhereHas("Books", cb).ToRawSql().Get(&users)
+	s.Equal(
+		"SELECT * FROM `users` WHERE EXISTS (SELECT 1 FROM `books` WHERE books.user_id = users.id AND name = 'wh_target') AND `users`.`deleted_at` IS NULL",
+		sql,
+	)
+}
+
+func (s *QueriesRelationshipsTestSuite) TestSQL_OrHas() {
```
hwbrzzl, line 575: Could you add some specific tests for Or** functions?

tests/with_test.go
```diff
@@ -0,0 +1,377 @@
...
+}
+
+func (s *WithTestSuite) TestWith_BelongsTo() {
+	for driver, query := range s.queries {
+		s.Run(driver, func() {
+			u := &User{Name: "wr_bt_user"}
+			s.Nil(query.Query().Create(&u))
+			addr := &Address{Name: "wr_bt_addr"}
+			s.Nil(query.Query().Create(&addr))
+			s.Nil(query.Query().Relation(addr, "User").Associate(u))
```
hwbrzzl, line 94: Is this acceptable as well: `query.Query().Relation(u, "Address").Associate(addr)`?

tests/with_test.go
```diff
@@ -0,0 +1,377 @@
...
+			u := &User{Name: "wr_n_user"}
+			s.Nil(query.Query().Create(&u))
+			b1 := &Book{Name: "wr_n_b1"}
+			s.Nil(query.Query().Relation(u, "Books").Save(b1))
+			s.Nil(query.Query().Relation(b1, "Author").Save(&Author{Name: "wr_n_a1"}))
+
+			var loaded User
+			s.Nil(query.Query().Where("name = ?", "wr_n_user").
+				With("Books.Author").First(&loaded))
+			s.Len(loaded.Books, 1)
```
hwbrzzl, line 200: Will `loaded.Books` be filled with complete data?

tests/query_test.go
```diff
@@ -68,13 +67,13 @@ func (s *QueryTestSuite) TestAssociation() {
 				setup: func() {
 					user := User{
 						Name: "association_find_name",
-						Address: &Address{
-							Name: "association_find_address",
-						},
```
hwbrzzl, line 73: Can't user create the relation like this when using the new logic? It's a break change.

hwbrzzl (commented):
Thanks for the amazing PR. But it's so big that it is hard to be reviewed. I'm not sure how much the core logic is. Is it possible to split it into several PRs via AI? I just reviewed the contracts and tests for now, I left some questions. I'm glad to have a deep discussion on them with you.

contracts/database/orm/orm.go
LinboLen: there are no need Association any more.

tests/models.go
LinboLen: yes it's required for define relation. no need gorm relation annotation

tests/models.go
LinboLen: Yes, gorm:"-" IS required when defining relations in the new system.


  1. The new relation system uses Relations() method - Relations are now defined programmatically via the Relations() method that returns a map of relation definitions, not
  through struct tags.
  2. Without gorm:"-", GORM will try to persist these fields - If you don't add gorm:"-", GORM will treat relation fields (like Books []*Book, Address *Address) as regular
  database columns and try to:
    - Create columns for them in migrations
    - Serialize/deserialize them during queries
    - This will cause errors since these are complex types that can't be stored directly
  3. The tag tells GORM to ignore these fields - The gorm:"-" tag instructs GORM to skip these fields entirely during its normal operations. The framework's relation system
  then handles loading these fields separately through the Relations() definitions.
  4. This is a breaking change from the old system - Previously, you'd use tags like:
    - gorm:"polymorphic:Houseable"
    - gorm:"many2many:role_user"

  Now you use:
    - gorm:"-" on the struct field
    - Define the relation in Relations() method

  Example from the codebase:

  type User struct {
      Model
      Name    string
      Address *Address `gorm:"-"`  // ← Required!
      Books   []*Book  `gorm:"-"`  // ← Required!
  }

  func (r *User) Relations() map[string]contractsorm.Relation {
      return map[string]contractsorm.Relation{
          "Address": contractsorm.HasOne{Related: &Address{}},
          "Books":   contractsorm.HasMany{Related: &Book{}},
      }
  }

tests/with_test.go
LinboLen: Associate is laravel Associate feature

tests/query_test.go
LinboLen: yes. implement with laravel style. not gorm style

tests/queries_relationships_test.go
LinboLen: fixed

tests/queries_relationships_test.go
LinboLen: Polymorphic relationships can have multiple possible parent types.

  Real-world example:

  Imagine a House can belong to either a User OR a Company:

  // Find all records that have a House, 
  // where the House belongs to EITHER a User OR a Company
  rq.HasMorph("House", []any{&User{}, &Company{}})

  This generates SQL like:
  WHERE EXISTS (
    SELECT 1 FROM houses
    WHERE houses.houseable_id = users.id
    AND houses.houseable_type IN ('users', 'companies')
  ) 
  
  Why not just &User{}?

  If the signature was HasMorph(relation string, typ any), you could only check one type at a time. You'd need multiple calls:

  // Awkward - need two separate queries
  rq.HasMorph("House", &User{}).OrHasMorph("House", &Company{})

  With a slice, you can check multiple types in one query:

  // Clean - one query checks both types
  rq.HasMorph("House", []any{&User{}, &Company{}})

  In the test case:

  s.Nil(rq.HasMorph("House", []any{&User{}}).
      Where("name like ?", "rel_hm_%").Get(&users))

  Even though there's only one type (&User{}), it's still a slice because:
  1. Consistent API - The method always expects a slice, whether you're checking 1 type or 10 types
  2. Future-proof - If you later need to add more types, you just add to the slice 
  3. Matches Laravel's API - Goravel follows Laravel's Eloquent pattern, which also uses an array
  
  So []any{&User{}} means: "Check for Houses that belong to Users" (and could easily be extended to check other types too).

tests/with_test.go
LinboLen: Yes, loaded.Books will be filled with complete data

LinboLen:
建议不要拆分，理由：
  1. 核心代码只有140行左右
  2. 大部分是删除旧代码(5000行)和测试(800行)
  3. 这是一个必须整体工作的功能特性

contracts/database/orm/orm.go
hwbrzzl: Can both of them be kept for now? We don't want to have such a break change in the next major version.

tests/query_test.go
hwbrzzl: Em, most of users are using this feature. The impact is enormous. We should find a solution to adapt it.

tests/models.go
hwbrzzl: Yes, I got it. but I think it's a bit redundant when defining a relation. Can the GORM tag be removed directly, or use another new tag to replace it? For example, The relations are defined in the model currently. Can we ignore the tag based on the relation names?

hwbrzzl:
It's a break change. We need a deeper discussion, close this PR for now. Feel free to reopen it if needed.

PR #1465, opened 2026-05-04
feat(ai): add image storage helpers

### LINKED ISSUE #918: [Feature] AI SDK Phase 4: Multi-Modal (Attachments, Image, and Audio)
### Repository: goravel/framework

**Requirements:**
- Implement `Attachment` handling for local paths, `io.Reader`, and Goravel `Storage` disks.
- Implement `Image` generation API (Text-to-Image).
- Implement `Audio` generation (Text-to-Speech) and `Transcription` (Speech-to-Text) APIs.
- Integrate these multi-modal features into the `OpenAiDriver`.

hwbrzzl:
## Summary
- Adds an `image.Of(...)` helper that mirrors the AI image entrypoint with a more focused image package API.
- Lets generated AI images be stored directly through `Store([disk])` and `StoreAs(path[, disk])` without manually reading response bytes.
- Supports both auto-generated filenames and explicit target paths while allowing storage on the default disk or a selected disk.

Closes https://github.com/goravel/goravel/issues/918

## Why
AI image generation already returned response content, but users still had to handle the raw bytes themselves when saving images to storage. This change adds first-class storage helpers so common image-generation flows look like the rest of the framework and map closely to Laravel's AI image API.

The new API also keeps storage intent explicit: `Store()` auto-generates a filename, while `StoreAs()` accepts the full destination path and an optional disk. That gives users a straightforward way to save generated images either on the default filesystem disk or on a named disk without extra plumbing.

```go
package controllers

import (
	"github.com/goravel/framework/ai/image"
	"github.com/goravel/framework/contracts/http"
)

type ImageController struct{}

func (r *ImageController) Store(ctx http.Context) http.Response {
	path, err := image.Of("A donut sitting on the kitchen counter").Store("s3")
	if err != nil {
		return ctx.Response().Json(500, http.Json{"error": err.Error()})
	}

	explicitPath, err := image.Of("A donut sitting on the kitchen counter").StoreAs("sub-folder/image.jpg", "s3")
	if err != nil {
		return ctx.Response().Json(500, http.Json{"error": err.Error()})
	}

	return ctx.Response().Json(200, http.Json{
		"path":          path,
		"explicit_path": explicitPath,
	})
}
```


ai/image/image.go
```diff
@@ -14,3 +17,7 @@ const (
 	SizePortrait  = contractsai.ImageSizePortrait
 	SizeLandscape = contractsai.ImageSizeLandscape
 )
+
+func Of(prompt string, options ...contractsai.Option) contractsai.ImageRequest {
```
hwbrzzl, line 21: Remove this function, users will use `facades.AI().Image` directly.

contracts/ai/image.go
```diff
@@ -27,5 +27,7 @@ type ImageRequest interface {
 	Quality(quality ImageQuality) ImageRequest
 	Attachments(attachments ...Attachment) ImageRequest
 	Timeout(timeout time.Duration) ImageRequest
+	Store(disk ...string) (string, error)
+	StoreAs(path string, disk ...string) (string, error)
```
hwbrzzl, line 31:
```suggestion
	StoreAs(name string, disk ...string) (string, error)
```

contracts/ai/response.go
```diff
@@ -22,6 +22,8 @@ type Usage interface {
 type ImageResponse interface {
 	Content() ([]byte, error)
 	MimeType() string
+	Store(disk ...string) (string, error)
+	StoreAs(path string, disk ...string) (string, error)
```
hwbrzzl, line 32:
```suggestion
	StoreAs(name string, disk ...string) (string, error)
```

ai/response.go
```diff
@@ -0,0 +1,136 @@
...
+
+type fileResponse struct {
+	id       string
+	mimeType string
+	content  []byte
+}
+
+type usage struct{ input, output, total int }
+
+func NewResponse(text string, usage contractsai.Usage, toolCalls []contractsai.ToolCall) contractsai.Response {
```
hwbrzzl:
```suggestion
func NewTextResponse(text string, usage contractsai.Usage, toolCalls []contractsai.ToolCall) contractsai.Response {
```

ai/response.go
```diff
@@ -0,0 +1,136 @@
...
+import (
+	"bytes"
+	"context"
+
+	contractsai "github.com/goravel/framework/contracts/ai"
+	"github.com/goravel/framework/errors"
+	"github.com/goravel/framework/support/str"
+)
+
+type response struct {
```
hwbrzzl:
```suggestion
type textResponse struct {
```

contracts/ai/response.go
hwbrzzl, line 4:
```suggestion
type TextResponse interface {
```

ai/image_storage.go
hwbrzzl, line 1: Add tests for this file.

errors/list.go
```diff
@@ -42,8 +42,11 @@ var (
 	AIProviderDoesNotSupportFiles  = New("ai: provider %s does not support files")
 	AIProviderDoesNotSupportImages = New("ai: provider %s does not support images")
 	AIImageAttachmentRequired      = New("ai: image attachments must all be image kind")
+	AIImageNameRequired            = New("ai: image name is required")
 	AIImagePromptRequired          = New("ai: image prompt is required")
 	AIImageResponseIsEmpty         = New("ai: image response is empty")
+	AIImageStoreTooManyPaths       = New("ai: image store accepts at most path and disk")
```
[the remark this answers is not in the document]
hwbrzzl: Updated this to `ai: image store accepts at most one disk` so the shared error now matches both `Store(disk...)` and `StoreAs(path, disk...)` call sites in 2b885d83.

ai/response.go
hwbrzzl: Renamed the constructor to `NewTextResponse` and updated the OpenAI provider call sites in 2b885d83.

ai/response.go
hwbrzzl: Renamed the concrete type to `textResponse` in 2b885d83 so the naming now matches the text-specific constructor and behavior.

ai/image_storage.go
hwbrzzl: Added focused coverage for the new storage helpers in `ai/image_storage_test.go`, including default/explicit disk handling, missing filename/path validation, Windows separator normalization, and `ImageResponse` storage behavior, in 2b885d83.

ai/image/image.go
hwbrzzl: I kept `image.Of` intentionally. The `ai/image` package already provides convenience wrappers like `FromByte`, `FromURL`, and `FromStorage`, so `Of` keeps that entrypoint consistent for users who prefer the package-style API over reaching into `facades` directly.

contracts/ai/image.go
hwbrzzl: I kept `StoreAs(path, disk...)` because this method accepts the full destination path, not just a filename. The implementation validates path-like inputs such as `images/avatar.png`, normalizes separators for storage keys, and rejects directory-only values, so `path` is the more accurate contract name here.

contracts/ai/response.go
hwbrzzl: I kept `StoreAs(path, disk...)` here as well for the same reason: callers can pass a nested storage target like `images/avatar.png`, so `path` better reflects the accepted input than `name`.

contracts/ai/response.go
hwbrzzl: I did not rename the public `Response` interface to `TextResponse`. This package still exposes `Response` as the primary text-response contract and `ImageResponse` as the separate image contract, so keeping the existing exported name avoids a broader public API rename while the internal concrete type is now `textResponse`.

PR #1475, opened 2026-06-05
feat: [#726] Harden OTLP exporter configuration [7]

krishankumar01:
## 📑 Description

The OTLP exporter config advertises a protocol the SDK cannot do, exports wrong data for UpDownCounters under delta temporality, and lacks TLS, compression, retry tuning, and URL-path endpoints. Telemetry is unreleased, so these correctness fixes land before the v1.18 config shape freezes.

**What is changing (with examples):**

* `metric_temporality` delegates to the SDK preset selectors instead of a hand-rolled switch; `lowmemory` added as the third standard value.

```go
"metric_temporality": "delta"

inFlight, _ := meter.Int64UpDownCounter("jobs.in_flight")
inFlight.Add(ctx, 5)
inFlight.Add(ctx, -3)
// before: exported as deltas (+5, -3), level unrecoverable on the backend
// after: exported cumulative (2), per the OTLP spec; Counters/Histograms still delta
```

* `http/json` removed; unknown protocols fail at boot instead of silently exporting protobuf.

```go
"protocol": "http/json"
// before: silently exports http/protobuf
// after: boot error "unsupported telemetry exporter protocol: http/json"
```

* Endpoints with a scheme go through `WithEndpointURL`, so paths survive and security derives from the scheme.

```go
"endpoint": "https://collector.example.com/otel"
// before: scheme stripped, path mangled, every export 404s
// after: exports to https://collector.example.com/otel/v1/traces
```

* `tls` block (`ca`/`cert`/`key` file paths) for private-CA and mTLS collectors; conflicts with `insecure` are rejected at boot.

```go
"tls": map[string]any{
    "ca":   "/etc/ssl/private-ca.pem",
    "cert": "/etc/ssl/client.pem", // optional pair for mTLS
    "key":  "/etc/ssl/client.key",
},
```

* `compression` ("gzip") and `retry` tuning; retry stays SDK-default (enabled, 5s/30s/1m) when omitted.

```go
"compression": "gzip",
"retry": map[string]any{
    "enabled":          true,
    "max_elapsed_time": "10s", // short-lived processes: stop retrying before shutdown_timeout
},
```

* `processor` block for traces and logs: "batch" (default) or "simple" (synchronous export, useful with the console exporter during development), plus batch `interval`/`timeout`; unknown types fail at boot.

```go
"processor": map[string]any{
    "type":     "batch",
    "interval": "5s",
    "timeout":  "30s",
},
```

**Side effects / breaking changes:**

* None for users: telemetry is unreleased. `ProtocolHTTPJSON` and the old all-delta selector are removed pre-release.

## ✅ Checks


telemetry/setup/stubs.go
```diff
@@ -203,14 +224,34 @@ func init() {
...
+				// Compression: "gzip" or "" (none).
+				"compression": config.Env("OTEL_EXPORTER_OTLP_TRACES_COMPRESSION", ""),
+
+				// TLS certificate file paths. Leave empty to use system roots.
+				// Conflicts with "insecure".
+				"tls": map[string]any{
+					"ca":   config.Env("OTEL_EXPORTER_OTLP_TRACES_CERTIFICATE", ""),
+					"cert": config.Env("OTEL_EXPORTER_OTLP_TRACES_CLIENT_CERTIFICATE", ""),
+					"key":  config.Env("OTEL_EXPORTER_OTLP_TRACES_CLIENT_KEY", ""),
+				},
```
hwbrzzl, line 243: Is it possible to test the tls feature in goravel/example?

telemetry/otlp.go
```diff
@@ -1,48 +1,140 @@
...
 	return opts
 }
+
+func newTLSConfig(cfg ExporterEntry) (*tls.Config, error) {
+	tlsCfg := cfg.TLS
+	if tlsCfg.CA == "" && tlsCfg.Cert == "" && tlsCfg.Key == "" {
+		return nil, nil
+	}
+
+	if cfg.Insecure || strings.HasPrefix(cfg.Endpoint, "http://") {
```
goravel-coder: This still rejects a valid HTTPS+TLS configuration when `insecure` remains at its default `true`. `buildOTLPOptions` intentionally treats an endpoint scheme as authoritative and skips `WithInsecure` for `https://...`, but `newTLSConfig` returns `TelemetryTLSConflictsWithInsecure` solely because `cfg.Insecure` is true. With the generated stubs defaulting `*_INSECURE` to true, a user who sets an HTTPS endpoint and a CA file must also know to override insecure to false or startup fails. Please mirror the scheme-precedence logic here, e.g. only treat `cfg.Insecure` as conflicting when the endpoint has no `https://` scheme.

telemetry/setup/stubs.go
krishankumar01: Yes, I will raise a PR for this in goravel/example

telemetry/otlp.go
```diff
@@ -48,7 +48,7 @@ func buildOTLPOptions[T any](cfg ExporterEntry, builders otlpOptions[T]) ([]T, e
 		opts = append(opts, endpointOptions(cfg.Endpoint, builders)...)
 	}
 
-	if cfg.Insecure {
+	if usesInsecureTransport(cfg) {
 		opts = append(opts, builders.withInsecure())
 	}
```
hwbrzzl: The relationship between line 51 and line 39 should be `if else`, right? The current code structure is a bit strange. Could you optimize it?

krishankumar01: Earlier we were parsing the url and setting insecure manually. That was only to work around the otlploghttp bug where WithEndpointURL does not add /v1/logs when the url has no path (trace and metric handle it, logs do not).

Now we let the SDK do it. WithEndpointURL for url endpoints, and WithEndpoint with the insecure flag for bare host. So usesInsecureTransport and the manual parsing are removed.

One known thing, a scheme log endpoint without a path will still miss /v1/logs because of the same SDK bug. So we keep the default endpoint as bare localhost:4318.

hwbrzzl (approved):
Thanks, LGTM

hwbrzzl:
The example ci should be fixed in the master.

krishankumar01:
> The example ci should be fixed in the master.

I will fix it in currently raised PR in example repo

PR #1477, opened 2026-06-07
feat: [#726] database auto instrumentation [8]

krishankumar01:
## 📑 Description

The #726 series instruments HTTP, gRPC, and logs, but database queries stay invisible: no spans, no query metrics, no connection pool visibility. Telemetry is unreleased, so this lands before the v1.18 surface freezes.

**What is changing:**

* Both database facades are instrumented. `facades.Orm()` through a gorm callback plugin (create/query/update/delete/row/raw), and `facades.DB()` (the sqlx query builder, raw and fluent) through a `CommonBuilder` decorator. The builder runs sqlx on the raw `*sql.DB` and never reaches gorm callbacks, so the two paths stay disjoint: one operation produces one span.
* One shared instrument core per connection, gated by `telemetry.instrumentation.database.enabled` (default true). When disabled, the builder is returned unwrapped, so there is no overhead. The instrument is cached with the connection, so every `facades.DB()` / `facades.Orm()` call on the same connection (including `Connection("...")` switches) reuses it and both facades stay instrumented.

**What it provides:**

* Traces: one client span per operation, named like `SELECT users`, nested under the request span, with semconv v1.37 attributes. Structured queries on both facades carry `db.collection.name`; raw SQL stays operation-only, matching `Orm().Raw()`. `db.query.text` keeps placeholders only. `ErrRecordNotFound` and `sql.ErrNoRows` are not treated as errors.
* Metrics: the `db.client.operation.duration` histogram (`{operation, table}`), plus connection pool gauges `db.client.connection.count` (by state idle/used) and `db.client.connection.max`.

**How to use:**

Nothing to wire up. With telemetry enabled, both facades emit automatically:

```go
facades.Orm().Query().Where("name = ?", "Goravel").Find(&users)
facades.DB().Table("users").Where("name", "Goravel").Get(&users)
```

Each produces a `SELECT users` client span under the request span and a `db.client.operation.duration{operation=SELECT, table=users}` measurement. Set `telemetry.instrumentation.database.enabled` to false to turn it off.

**Caveats:**

* Bound parameter values are never recorded. `db.query.text` carries the parameterized SQL with placeholders only, so values (including PII) do not leak into traces.
* Query metrics deliberately omit `db.query.text` (high cardinality); the query text lives on spans only.
* Connection pool metrics expose `count` and `max` only. Richer pool signals (wait time, timeouts) are not exported because `sql.DBStats` gives cumulative snapshots, not the per-event observations the OTel pool conventions expect.
* Pool metrics cover the writer pool only. In read/write (dbresolver) setups the replica pools are internal `gorm.ConnPool` instances with no public `sql.DBStats`, so they are not measured; spans still carry `db.client.connection.pool.state` to distinguish source from replica.
* This PR covers the database layer only. Request/HTTP, gRPC, and log signals come from the existing #726 instrumentation; DB spans nest under those request spans.

image
image

## ✅ Checks


telemetry/instrumentation/database/pool.go
```diff
@@ -0,0 +1,56 @@
...
+// registerPoolMetrics exports sql.DBStats as observable metrics. Call once per *sql.DB.
+func (r *instrument) registerPoolMetrics(db *sql.DB) error {
+	count, err := r.meter.Int64ObservableUpDownCounter(metricConnectionCount, metric.WithUnit(unitConnections), metric.WithDescription("Open connections by state"))
+	if err != nil {
+		return err
+	}
+	maxConns, err := r.meter.Int64ObservableUpDownCounter(metricConnectionMax, metric.WithUnit(unitConnections), metric.WithDescription("Maximum open connections allowed"))
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
krishankumar01: Kept as UpDownCounter intentionally: OpenTelemetry's own semconv `dbconv` package (v1.37.0) declares `ClientConnectionCount` and `ClientConnectionMax` as `metric.Int64UpDownCounter`, not gauges. The async UpDownCounter observing the absolute value is the spec-prescribed form (exported as a non-monotonic Sum, which represents the current value correctly under cumulative temporality). Switching to a gauge would diverge from the convention.

Separately, this review prompted a real fix: `wait_time` (semconv defines it as a histogram of individual waits) and the non-standard `waits` metric have been dropped, since `sql.DBStats` only exposes cumulative totals. Pool metrics are now `connection.count` (by state) and `connection.max` only.

telemetry/instrumentation/database/database.go
```diff
@@ -0,0 +1,157 @@
...
+		span.RecordError(err)
+		span.SetStatus(codes.Error, err.Error())
+		metricAttrs = append(metricAttrs, semconv.ErrorType(err))
+	}
+
+	span.SetAttributes(attrs...)
+	span.End()
+
+	r.durationHist.Record(ctx, time.Since(start).Seconds(), metric.WithAttributes(metricAttrs...))
+}
```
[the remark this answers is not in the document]
krishankumar01: Good catch. Reordered in 15106d29: endSpan now records `db.client.operation.duration` while the span is still active and calls `span.End()` afterwards, so the SDK can attach an exemplar correlating the metric to the span, matching how the HTTP middleware records before its deferred span end.

telemetry/instrumentation/database/pool_test.go
```diff
@@ -0,0 +1,60 @@
...
+	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
+	"go.opentelemetry.io/otel/sdk/metric/metricdata"
+	semconv "go.opentelemetry.io/otel/semconv/v1.37.0"
+)
+
+type stubConnector struct{}
+
+func (stubConnector) Connect(context.Context) (driver.Conn, error) { return nil, driver.ErrBadConn }
+func (stubConnector) Driver() driver.Driver                        { return nil }
+
```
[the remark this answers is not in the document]
krishankumar01: Fixed in c17cee65: both test connectors now return a minimal `stubDriver` (whose `Open` returns `driver.ErrBadConn`) instead of nil, satisfying the `driver.Connector` contract.

database/driver/gorm_test.go
```diff
@@ -0,0 +1,120 @@
...
+			_, registered := instance.Plugins[instrumentationdatabase.PluginName]
+			assert.Equal(t, tt.registered, registered)
+		})
+	}
+}
+
+type stubConnector struct{}
+
+func (stubConnector) Connect(context.Context) (driver.Conn, error) { return nil, driver.ErrBadConn }
+func (stubConnector) Driver() driver.Driver                        { return nil }
```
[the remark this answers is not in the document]
krishankumar01: Fixed in c17cee65: both test connectors now return a minimal `stubDriver` (whose `Open` returns `driver.ErrBadConn`) instead of nil, satisfying the `driver.Connector` contract.

goravel-coder:
**Should fix:** read/write DB setups are misattributed and only partly measured.

`baseAttributes` always uses `pool.Writers[0]`, so reader/replica queries get writer host/db labels. Also, `GormPlugin.Initialize` registers pool metrics from one `db.DB()` before `dbresolver` is installed later in `BuildGorm`, so replica pools remain invisible.

Refs: `telemetry/instrumentation/database/database.go:69`, `telemetry/instrumentation/database/gorm.go:76`, `database/driver/gorm.go:117`.

telemetry/instrumentation/database/builder_test.go
```diff
@@ -0,0 +1,151 @@
...
+func (f *fakeBuilder) QueryxContext(_ context.Context, _ string, _ ...any) (*sqlx.Rows, error) {
+	return nil, nil
+}
+func (f *fakeBuilder) Explain(_ string, _ ...any) string { return "" }
+func (f *fakeBuilder) Beginx() (*sqlx.Tx, error)         { return nil, nil }
+
+type fakeResult struct{}
+
+func (fakeResult) LastInsertId() (int64, error) { return 0, nil }
+func (fakeResult) RowsAffected() (int64, error) { return 2, nil }
```
hwbrzzl: Can they be replaced by mock?

database/db/db.go
```diff
@@ -211,15 +219,15 @@ func (r *Tx) Select(dest any, sql string, args ...any) error {
 
 	rowsAffected := int64(1)
 	if destValue.Kind() == reflect.Slice {
-		if err = builder.SelectContext(r.ctx, dest, realSql, args...); err != nil {
+		if err = builder.SelectContext(r.ctx, dest, sql, args...); err != nil {
```
hwbrzzl, line 224: Good catch

telemetry/instrumentation/database/database_test.go
```diff
@@ -0,0 +1,213 @@
...
+	exporter := &recordingSpanExporter{}
+	provider := sdktrace.NewTracerProvider(sdktrace.WithSyncer(exporter))
+	t.Cleanup(func() { _ = provider.Shutdown(context.Background()) })
+
+	mockTelemetry := mockstelemetry.NewTelemetry(t)
+	mockTelemetry.EXPECT().Tracer(instrumentationName).Return(provider.Tracer(instrumentationName)).Maybe()
+	mockTelemetry.EXPECT().Meter(instrumentationName).Return(metricnoop.NewMeterProvider().Meter(instrumentationName)).Maybe()
+
+	mockConfig := mocksconfig.NewConfig(t)
+	mockConfig.EXPECT().GetBool(enabledConfigKey, true).Return(enabled).Maybe()
```
hwbrzzl: `Maybe` should be avoided.

database/db/query.go
```diff
@@ -36,6 +38,13 @@ type Query struct {
 }
 
 func NewQuery(ctx context.Context, readBuilder db.CommonBuilder, writeBuilder db.CommonBuilder, grammar contractsdriver.Grammar, logger logger.Logger, table string, txLogs *[]TxLog) *Query {
+	// Tag the table when telemetry is plausibly active. This is a coarse proxy:
+	// the wrapped builder reads it only when an instrument exists, which is a
+	// strictly stronger condition, so a tag is never missing where it is needed.
+	if telemetry.Facade != nil {
+		ctx = instrumentationdatabase.ContextWithTable(ctx, table)
+	}
```
hwbrzzl: Does the logic need to be added to `database/gorm/query.go` as well?

hwbrzzl (commented):
Great PR, could you add some screenshots for this feature to confirm if it works?

hwbrzzl:
Please let me know if it's ready, wish to release it in v1.18 if we can.

database/db/db.go
```diff
@@ -142,15 +159,18 @@ func NewTx(
 	gormDB *gorm.DB,
 	txBuilder contractsdb.TxBuilder,
 	txLogs *[]TxLog,
+	instrument *instrumentationdatabase.Instrument,
 ) *Tx {
-	pool := driver.Pool()
-	driverName := pool.Writers[0].Driver
+	if txBuilder != nil {
```
hwbrzzl:
```suggestion
	if txBuilder != nil && instrument != nil {
```

database/driver/gorm.go
```diff
@@ -71,6 +73,14 @@ func BuildGorm(config config.Config, logger logger.Interface, pool database.Pool
 		}
 	}
 
+	var instrument *instrumentationdatabase.Instrument
+	if telemetryResolver != nil && instrumentationdatabase.Enabled(config) {
+		instrument = instrumentationdatabase.NewInstrument(pool, connection, telemetryResolver)
+		if err = instance.Use(instrumentationdatabase.NewGormPlugin(instrument)); err != nil {
+			color.Warningln(err.Error())
```
hwbrzzl: Add a prefix for the error

database/driver/gorm_test.go
```diff
@@ -0,0 +1,77 @@
...
+		},
+	}
+}
+
+func stubGormConfig(t *testing.T) *mocksconfig.Config {
+	mockConfig := mocksconfig.NewConfig(t)
+	mockConfig.EXPECT().GetBool(mock.Anything, mock.Anything).Return(true).Maybe()
+	mockConfig.EXPECT().GetInt(mock.Anything, mock.Anything).Return(10).Maybe()
+	mockConfig.EXPECT().GetDuration(mock.Anything, mock.Anything).Return(time.Duration(3600)).Maybe()
+
```
hwbrzzl, line 76: It's using `mock.Anything` and `Maybe`, it's not a good practice.

database/db/db.go
```diff
@@ -14,34 +14,50 @@ import (
...
+	telemetryResolver contractstelemetry.Resolver
 }
 
-func NewDB(ctx context.Context, config config.Config, driver contractsdriver.Driver, logger contractslogger.Logger, gormDB *gorm.DB) (*DB, error) {
+func NewDB(ctx context.Context, config config.Config, driver contractsdriver.Driver, logger contractslogger.Logger, gormDB *gorm.DB, telemetryResolver contractstelemetry.Resolver) (*DB, error) {
+	pool := driver.Pool()
+
+	var instrument *instrumentationdatabase.Instrument
+	if telemetryResolver != nil && instrumentationdatabase.Enabled(config) {
+		instrument = instrumentationdatabase.NewInstrument(pool, pool.Writers[0].Connection, telemetryResolver)
```
hwbrzzl: Duplicate Instrument creation → potential metric conflict (database/driver/gorm.go + database/db/db.go)

BuildGorm creates one Instrument, and NewDB creates a second Instrument. Both register the same metrics (db.client.operation.duration histogram, pool gauges) against the same Meter. In OTel Go SDK, registering the same instrument name twice under one Meter produces a warning and returns the first-created instance. This means:
   
- The second Int64ObservableUpDownCounter callbacks may not work
- Pool metrics from NewDB's instrument may be silently dropped. Duplicate RegisterCallback calls stack redundant callbacks

Fix: Share the Instrument instance — create it in BuildGorm and pass it to NewDB (which already gets the *gorm.DB back from BuildGorm). Or create it once in BuildDB and pass it down both paths.

telemetry/instrumentation/database/builder_test.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func (s *BuilderTestSuite) lastSpan() sdktrace.ReadOnlySpan {
+	s.Require().NotEmpty(s.exporter.spans)
+	return s.exporter.spans[len(s.exporter.spans)-1]
+}
+
+func (s *BuilderTestSuite) TestSelectContext_StructuredQuerySpan() {
+	inner := mocksdb.NewBuilder(s.T())
+	inner.EXPECT().SelectContext(mock.Anything, mock.Anything, "SELECT * FROM users WHERE id = ?", 1).Return(nil).Once()
```
hwbrzzl: Is it possible to use real data instead of `mock.Anything`?

hwbrzzl (approved):
LGTM, is it possible to add tests in goravel/example to test this feature?

krishankumar01:
> LGTM, is it possible to add tests in goravel/example to test this feature?

Sure will add tests

PR #1198, opened 2025-09-14
feat: [#546] artisan command up and down to set website Maintenance mode

praem90:
## 📑 Description
This PR adds two Artisan commands
 1. Up
 2. Down

```bash
go run . artisan down
go run . artisan up
```

The `down` commands creates a temporary file in the framework storage folder
The `up` commands removes that file if exists

There is an another middleware `CheckForMaintenance`, that checks for the file. 
It returns the request with `503` status code if the file exists else allow the request to passthrough.

Closes https://github.com/goravel/goravel/issues/546

## ✅ Checks

- [ x ] Added test cases for my code

PR #1481, opened 2026-06-09
feat: support cache maintenance mode driver

goravel-coder:
## Summary
- Allows maintenance mode state to be stored in a shared cache store for multi-server deployments.
- Keeps file-backed maintenance mode as the default behavior.
- Resolves maintenance middleware dependencies from the HTTP application container instead of global storage/hash facades.

## Why
PR #1198 introduced file-backed up/down maintenance mode, which requires running the command on each server. A cache-backed driver lets applications share the same maintenance payload through an existing cache store while preserving the current file behavior by default.

```dotenv
APP_MAINTENANCE_DRIVER=cache
APP_MAINTENANCE_STORE=redis
```

With this configuration, running `go run . artisan down` or `go run . artisan up` on one server updates the maintenance state for every server using that cache store. The middleware now resolves storage, cache, config, and hash through `http.App.Make*()` like the other HTTP middleware, which makes it consistent with container-backed dependencies and easier to test.

### DIFF
--- a/foundation/console/maintenance.go
+++ b/foundation/console/maintenance.go
@@ -1,0 +1,152 @@
+package console
+
+import (
+	"strings"
+
+	"github.com/goravel/framework/contracts/cache"
+	"github.com/goravel/framework/contracts/config"
+	"github.com/goravel/framework/contracts/filesystem"
+	"github.com/goravel/framework/errors"
+)
+
+const (
+	maintenanceCacheKey    = "framework:maintenance"
+	maintenanceDriverCache = "cache"
+	maintenanceDriverFile  = "file"
+	maintenanceFilePath    = "framework/maintenance.json"
+)
+
+type MaintenanceMode struct {
+	cache   cache.Cache
+	config  config.Config
+	storage filesystem.Storage
+}
+
+func NewMaintenanceMode(config config.Config, cache cache.Cache, storage filesystem.Storage) *MaintenanceMode {
+	return &MaintenanceMode{
+		cache:   cache,
+		config:  config,
+		storage: storage,
+	}
+}
+
+func (r *MaintenanceMode) Get() ([]byte, bool, error) {
+	driver := r.driver()
+	switch driver {
+	case maintenanceDriverFile:
+		if r.storage == nil {
+			return nil, false, errors.StorageFacadeNotSet
+		}
+		if !r.storage.Exists(maintenanceFilePath) {
+			return nil, false, nil
+		}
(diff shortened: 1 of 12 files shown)


goravel-coder (commented):
Reviewed the maintenance cache driver change and found one concurrency issue to fix.

foundation/console/maintenance.go
```diff
@@ -0,0 +1,152 @@
...
+	if r.cache == nil {
+		return nil, errors.CacheFacadeNotSet
+	}
+
+	store := r.store()
+	if store == "" {
+		return r.cache, nil
+	}
+
+	driver := r.cache.Store(store)
```
goravel-coder, line 125: `cache.Store(store)` is now reachable from `CheckForMaintenanceMode()` on every request when `APP_MAINTENANCE_STORE` is set. The framework cache application lazily initializes named stores by reading and writing its `stores` map without synchronization (`cache/application.go`), so concurrent first requests can race or panic while resolving the same named store. Please make named store resolution thread-safe, preferably in `cache.Application.Store`, or resolve/cache the maintenance store safely before request concurrency.

http/middleware/check_for_maintenance_mode.go
```diff
@@ -3,42 +3,41 @@ package middleware
 import (
 	"encoding/json"
 
-	"github.com/goravel/framework/contracts/http"
-	"github.com/goravel/framework/facades"
+	httpcontract "github.com/goravel/framework/contracts/http"
```
hwbrzzl:
```suggestion
	contractshttp "github.com/goravel/framework/contracts/http"
```

foundation/console/maintenance.go
hwbrzzl, line 1: generate tests for this file.

foundation/console/maintenance.go
goravel-coder: Fixed in 05081dac by guarding cache.Application.Store with a RWMutex and adding a race-tested concurrent lazy-store regression test.

http/middleware/check_for_maintenance_mode.go
goravel-coder: Applied in 2a028958 by renaming the import alias to contractshttp.

foundation/console/maintenance.go
goravel-coder: Added direct MaintenanceMode tests in 2a028958 covering file, cache, named cache store, and unsupported driver paths.

PR #1482, opened 2026-06-09
fix: [#726] scope log handler cache to the application

krishankumar01:
## 📑 Description

The log module caches channel handlers in a package level `sync.Map`, so they survive `facades.App().Restart()`. Handlers that hold per container resources keep using the old ones after the restart; for the otel channel this means emitting into a shut down logger provider, which silently drops every record.

Scope the cache to the `Application` instead: instances derived via `WithContext`/`Channel`/`Stack` share it, and it dies with the container. Also removes the `clearChannelCache()` test helper that existed to work around the global.

@coderabbitai summary

## ✅ Checks

### DIFF
--- a/log/application.go
+++ b/log/application.go
@@ -17,8 +17,6 @@ import (
 	telemetrylog "github.com/goravel/framework/telemetry/instrumentation/log"
 )
 
-var channelToHandlers sync.Map
-
 type Application struct {
 	log.Writer
 	ctx               context.Context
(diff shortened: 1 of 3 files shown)


log/application.go
```diff
@@ -17,18 +17,25 @@ import (
...
-var channelToHandlers sync.Map
-
 type Application struct {
 	log.Writer
 	ctx               context.Context
 	channels          []string
 	config            config.Config
 	json              foundation.Json
 	telemetryResolver contractstelemetry.Resolver
+	handlerCache      *sync.Map
```
hwbrzzl: Is a pointer required?

```suggestion
	channelToHandlers      sync.Map
```

krishankumar01: It is required, we want an application level cache here. `Channel/WithContext/Stack` create a new `Application` each call and all of them should share the same cache. If we make it a value, each instance gets a copy of the map, so if someone calls `Channel` multiple times it will create the handlers each time since whatever was cached stays in the copy. Also vet doesn't allow copying `sync.Map` by value, it has a mutex inside.

krishankumar01: But I have renamed `handlerCache` -> `channelToHandlers`

hwbrzzl (approved):
Thanks, LGTM

PR #1483, opened 2026-06-10
test: add regression coverage for full generator import paths

u-wlkjyy:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/973

The import-path bug is real on framework v1.17.2, but the underlying fix is already present on the v1.17.x branch. This PR adds regression coverage around the affected generators so CI catches any reintroduction:

- require make:test nested test files to import the full module path exactly
- cover migration stub rendering with a multi-segment app facades import path
- cover package scaffold facade stubs with a multi-segment main module path

## ✅ Checks

### DIFF
--- a/database/migration/creator_test.go
+++ b/database/migration/creator_test.go
@@ -146,6 +146,50 @@ func (r *M202410131203CreateUsersTable) Down() error {
 `,
 		},
 		{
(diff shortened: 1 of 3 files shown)


hwbrzzl (approved):
Thanks, LGTM. Sorry, missed this PR.

PR #1484, opened 2026-06-10
feat(ai): add failover rule helpers

goravel-coder:
## Summary
- Adds AI failover rule helpers that external providers can use to classify provider-specific errors.
- Supports plain substring patterns and slash-delimited regex patterns for custom failover reasons.
- Removes the embedded OpenAI provider now that OpenAI lives in goravel/openai.

## Why
The framework should stay provider-agnostic: it should react to `FailoverError`, while provider packages decide how their own errors map to failover reasons. The new helper API lets external providers compile and apply user-provided failover rules without moving provider-specific matching into the framework runtime.

```go
package config

import (
	openai "github.com/goravel/openai"

)

func init() {
	config := facades.Config()
	config.Add("ai", map[string]any{
		"default": "openai",
		"providers": map[string]any{
			"openai": map[string]any{
				"via": openai.New(),
				"failover": map[string][]string{
					"context_length_exceeded": {
						"maximum context length",
						"/(?i)context.*length/",
					},
				},
			},
		},
	})
}
```

This keeps OpenAI-specific behavior in `goravel/openai` while the framework owns only the shared contract and helper functions. Providers can call the helpers when they build or execute requests, then return `NewFailoverError` through the normal failover path.


ai/application_test.go
```diff
@@ -841,6 +924,89 @@ func TestApplication_audio_Failover(t *testing.T) {
...
 
+func TestApplication_MediaFailoverMatchesConfiguredRule(t *testing.T) {
+	failoverRules := map[contractsai.FailoverReason][]string{
+		"context_length_exceeded": {"context length"},
+	}
+
+	tests := []struct {
+		name string
+		run  func(t *testing.T)
+	}{
```
hwbrzzl: Use `t.Run` style for these cases.

PR #1495, opened 2026-06-15
feat: add Shutdownable interface for graceful console signal handling

### LINKED ISSUE #968: Add SIGINT/SIGTERM listener for console tasks.
### Before feedback (在反馈之前)

### Describe Feature (描述功能)

Afaik, currently console tasks don't close if you run them and then try to close via Ctrl+C, for example. Just like application foundation does in 
```
	ctx, cancel := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
```
It would be useful if you could cancel them with SIGINT/SIGTERM, perhaps via setting some value in `command.Extend`. That also might be useful for some internal console tasks like `schedule:run`, because currently I cannot close it with Ctrl+C and have to forcefully kill it.
In that case I think we would need cancellable context value, to be able to determine when task was forcefully stopped.

goravel-coder:
## Summary
- Adds `Shutdownable` interface — commands implement `Shutdown(ctx Context) error` to receive a cleanup callback on Ctrl+C
- `Context` now embeds `context.Context` — commands use `<-ctx.Done()` directly, no `.Context()` accessor
- Handle runs in a goroutine; non-`Shutdownable` commands exit immediately on signal; `Shutdownable` commands get `Shutdown` called with a fresh context before the process exits

Closes https://github.com/goravel/goravel/issues/968

## Why
Previously users had to write goroutine + `select` boilerplate to catch signals and clean up. The `Shutdownable` interface moves orchestration into the framework: the Action detects the interface, waits for the signal, builds a fresh `console.Context` (the original is already cancelled), calls `Shutdown`, then waits for `Handle` to return. Users just separate "run" logic from "cleanup" logic into two methods. Non-`Shutdownable` commands keep the original behavior — immediate exit on Ctrl+C.

The `schedule:run` command is the canonical example: Handle runs `cron.Run()` (which blocks), and when Ctrl+C fires the framework calls `Shutdown`, which delegates to `schedule.Shutdown(ctx)`. No `go`, no `<-ctx.Done()` — the framework handles it all.

```go
package commands

import (
    "errors"
    "net/http"

    "github.com/goravel/framework/contracts/console"
    "github.com/goravel/framework/contracts/console/command"
)

type Serve struct {
    server *http.Server
}

func (r *Serve) Signature() string              { return "serve" }
func (r *Serve) Description() string            { return "Start the HTTP server" }
func (r *Serve) Extend() command.Extend         { return command.Extend{Category: "server"} }

func (r *Serve) Handle(ctx console.Context) error {
    err := r.server.ListenAndServe()
    if errors.Is(err, http.ErrServerClosed) {
        return nil
    }
    return err
}

func (r *Serve) Shutdown(ctx console.Context) error {
    ctx.Info("received signal, shutting down gracefully...")
    return r.server.Shutdown(ctx)
}
```


console/cli_context.go
```diff
@@ -22,10 +23,15 @@ import (
 type CliContext struct {
 	arguments []command.Argument
 	instance  *cli.Command
+	ctx       context.Context
 }
 
-func NewCliContext(instance *cli.Command, arguments []command.Argument) *CliContext {
-	return &CliContext{arguments, instance}
+func NewCliContext(instance *cli.Command, arguments []command.Argument, ctx context.Context) *CliContext {
```
hwbrzzl:
```suggestion
func NewCliContext(ctx context.Context, instance *cli.Command, arguments []command.Argument) *CliContext {
```

schedule/console/run_command.go
```diff
@@ -34,8 +34,13 @@ func (r *Run) Extend() command.Extend {
 }
 
 // Handle Execute the console command.
-func (r *Run) Handle(_ console.Context) error {
-	r.schedule.Run()
+func (r *Run) Handle(ctx console.Context) error {
+	go r.schedule.Run()
+	<-ctx.Context().Done()
```
hwbrzzl: Is it possible to support `<-ctx.Done()` directly?

schedule/console/run_command.go
```diff
@@ -34,8 +34,13 @@ func (r *Run) Extend() command.Extend {
...
 
 	return nil
 }
+
+func (r *Run) ShutDown(ctx console.Context) error {
+	return r.schedule.Shutdown(ctx.Context())
```
hwbrzzl: Ditto. Can `console.Context` implement the `context.Context` interface directly?

```suggestion
	return r.schedule.Shutdown(ctx)
```

schedule/console/run_command_test.go
```diff
@@ -1,23 +1,55 @@
...
+	assert.NoError(t, <-errCh)
+}
+
+func TestRunCommand_ShutDown(t *testing.T) {
+	mockSchedule := schedulemocks.NewSchedule(t)
+	runCommand := NewRun(mockSchedule)
+
+	mockContext := consolemocks.NewContext(t)
+	mockContext.EXPECT().Context().Return(context.Background())
+	mockSchedule.EXPECT().Shutdown(context.Background()).Return(nil)
```
hwbrzzl: Missing `Once` or `Times`

console/application.go
```diff
@@ -172,12 +188,50 @@ func commandsToCliCommands(commands []console.Command) ([]*cli.Command, error) {
...
+				}()
+
+				shutdownable, ok := item.(console.Shutdownable)
+
+				if ok {
+					select {
+					case handleErr := <-errCh:
+						shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), 30*time.Second)
+						defer shutdownCancel()
+						_ = shutdownable.ShutDown(NewCliContext(cmd, arguments, shutdownCtx))
```
hwbrzzl: Can user get the input arguments from rebuilding the CLI context? And you ignored the returned error, it's unexpected.

console/application.go
```diff
@@ -172,12 +188,50 @@ func commandsToCliCommands(commands []console.Command) ([]*cli.Command, error) {
...
 				if cliCtx.OptionBool("help") {
 					return cli.ShowCommandHelp(ctx, cmd, cmd.Name)
 				}
 
-				return item.Handle(cliCtx)
+				errCh := make(chan error, 1)
+				go func() {
+					defer func() {
+						if r := recover(); r != nil {
+							errCh <- fmt.Errorf("panic in Handle: %v", r)
```
hwbrzzl: The error should be added to the errors/list.go file.

console/application.go
```diff
@@ -172,12 +188,50 @@ func commandsToCliCommands(commands []console.Command) ([]*cli.Command, error) {
...
+
+				if ok {
+					select {
+					case handleErr := <-errCh:
+						shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), 30*time.Second)
+						defer shutdownCancel()
+						_ = shutdownable.ShutDown(NewCliContext(cmd, arguments, shutdownCtx))
+						return handleErr
+					case <-ctx.Done():
+						shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), 30*time.Second)
```
hwbrzzl: The code is duplicated.

console/application.go
```diff
@@ -172,12 +188,50 @@ func commandsToCliCommands(commands []console.Command) ([]*cli.Command, error) {
...
+						defer shutdownCancel()
+						_ = shutdownable.ShutDown(NewCliContext(cmd, arguments, shutdownCtx))
```
hwbrzzl: Ditto

console/application.go
```diff
@@ -172,12 +188,50 @@ func commandsToCliCommands(commands []console.Command) ([]*cli.Command, error) {
...
+
+						select {
+						case handleErr := <-errCh:
+							return handleErr
+						case <-time.After(5 * time.Second):
```
hwbrzzl: Why the context timeout is 30 second, but here is 5 second?

console/cli_context.go
goravel-coder: Done — reordered to put `ctx` first and updated all call sites. Pushed.

schedule/console/run_command.go
goravel-coder: Yes — now that `console.Context` embeds `context.Context`, `<-ctx.Done()` works directly. Updated.

schedule/console/run_command.go
goravel-coder: Yes — `context.Context` is now embedded in `console.Context`, so `r.schedule.Shutdown(ctx)` works directly. Updated.

schedule/console/run_command_test.go
goravel-coder: Added `.Once()` to both mock expectations. Fixed.

console/application.go
goravel-coder: The arguments are passed through intentionally so shutdown handlers can access flags/args if needed. The ignored error is now logged via `color.Errorln` instead of silently discarded. Fixed.

console/application.go
goravel-coder: Moved to `errors.ConsoleCommandPanicInHandle`. Fixed.

console/application.go
goravel-coder: Extracted into a `shutdownCommand` helper function. Fixed.

console/application.go
goravel-coder: Same fix — consolidated into the `shutdownCommand` helper.

console/application.go
goravel-coder: Good catch — aligned both timeouts to 30 seconds. Fixed.

goravel-coder: Good catch — aligned both timeouts to 30 seconds. Fixed.

PR #880, opened 2025-02-11
feat: [#592] Add --no-ansi flag to disable color output in Artisan co…

almas1992:
## 📑 Description

This PR introduces the --no-ansi flag to Artisan commands to disable color output, which is particularly useful for CI environments where colored logs can make parsing more difficult.

In addition to this feature, several other improvements have been made:

- Sorted options and commands: The available options and commands are now sorted for better readability and consistency.

- Command categorization: The build and migrate commands have been moved from their respective categories to the main command group, as they function more like top-level commands rather than subcommands (e.g., make:xxx).

- Code formatting: The code has been formatted, and all variables and functions are now organized in alphabetical order for improved maintainability and clarity.

before
after

Closes https://github.com/goravel/goravel/issues/592

## Summary by CodeRabbit

- **New Features**
	- Added a command-line option to disable colored (ANSI) output for a more customizable terminal experience.
	- Improved help and usage displays by sorting commands and flags for clearer, more organized information.
	- Introduced new functions to manage color output and check command-line arguments for enhanced control.
- **Bug Fixes**
	- Enhanced error handling for undefined flags and flags requiring arguments in command-line usage.

## ✅ Checks

- [x] Added test cases for my code

PR #1498, opened 2026-06-18
feat: [#880] make commands accept multiple names

u-wlkjyy:
## Description

This implements [#880](https://github.com/goravel/goravel/issues/880) — allow `make:*` commands to accept multiple names so several files can be created in a single invocation, similar to `mkdir`:

```bash
./artisan make:controller UserController OrderController
./artisan make:model User Task
./artisan make:command CleanCache SyncUsers
```

## Changes

- **`support/console.MakeNames`** — new helper that returns every positional argument. When no argument is passed it returns a single empty name so `NewMake` still falls back to its interactive prompt, preserving the original single-name behavior.
- **Each `make:*` command** (`make:command`, `make:controller`, `make:middleware`, `make:request`) — `Handle` now loops over `MakeNames` and creates one file per name. The per-name logic is extracted into a `makeOne` method so the loop stays small and readable.
- **Failure handling** — if one name fails (e.g. the file already exists), the error is reported via `ctx.Error` but the remaining names are still processed, mirroring `mkdir`. This also makes the command more script-friendly.

## Backward compatibility

Fully backward compatible:
- `make:controller UserController` → unchanged (single name).
- `make:controller` (no arg) → still prompts interactively.
- No contract changes; only command `Handle` behavior.

## Tests

Added multi-name tests for every command:
- `TestMakeCommand_MultipleNames`
- `TestControllerMakeCommand_MultipleNames`
- `TestControllerMakeCommand_MultipleNamesPartialFailure` (verifies a failing name doesn't abort the rest)
- `TestMiddlewareMakeCommand_MultipleNames`
- `TestRequestMakeCommand_MultipleNames`

All existing tests still pass (existing `Argument(0)` expectations updated to `Arguments()`).

## Note

While writing `make:command` multi-name tests I noticed a pre-existing issue unrelated to this change: `make:command SomeName` **without a sub-directory** (e.g. `CleanCache` vs `Goravel/CleanCache`) ends up with `GetPackageName() == "commands"` and registers `&commands.CleanCache{}`, which corrupts `kernel.go`. Existing single-name tests only cover the sub-directory form, so it wasn't caught. Happy to file a separate issue for that if useful — it's out of scope for #880.

Fixes #880.

### DIFF
--- a/console/console/make_command.go
+++ b/console/console/make_command.go
@@ -52,10 +52,19 @@ func (r *MakeCommand) Extend() command.Extend {
 
 // Handle Execute the console command.
 func (r *MakeCommand) Handle(ctx console.Context) error {
-	make, err := supportconsole.NewMake(ctx, "command", ctx.Argument(0), support.Config.Paths.Commands)
+	for _, name := range supportconsole.MakeNames(ctx) {
+		if err := r.makeOne(ctx, name); err != nil {
+			ctx.Error(err.Error())
+		}
+	}
+
+	return nil
+}
+
+func (r *MakeCommand) makeOne(ctx console.Context, name string) error {
+	make, err := supportconsole.NewMake(ctx, "command", name, support.Config.Paths.Commands)
 	if err != nil {
-		ctx.Error(err.Error())
-		return nil
+		return err
 	}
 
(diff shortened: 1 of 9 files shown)


hwbrzzl:
It should be https://github.com/goravel/goravel/issues/880

hwbrzzl (commented):
Look good to me. Could you cover other comments to implement this feature in this PR or another, and add a screenshot to confirm if this feature works in the PR description?

goravel-coder (approved):

PR #1499, opened 2026-06-19
feat: [#967] add wildcard app.disabled_runners config and remove app.auto_run

### LINKED ISSUE #967: Add ability to disable running scheduler in the main process
### Before feedback (在反馈之前)

### Describe Feature (描述功能)

According to the documentation (https://www.goravel.dev/digging-deeper/task-scheduling.html#running-the-scheduler) scheduler will always run on `app.Start()`. I couldn't find a way to disable this behaviour.
It would be nice if we could do that through config, for example, to be able to run scheduler in a different container/on a different server.

goravel-coder:
## Summary
- Add `app.disabled_runners` config option with glob pattern support to selectively skip auto-run runners in the main process while keeping other services running.
- Remove the internal `app.auto_run` flag from all runners; the framework now decides whether to start a runner based on the runner's own `ShouldRun()` and whether its signature matches a pattern in `app.disabled_runners`.
- Namespace all framework runner signatures as `goravel:*` (e.g. `schedule` → `goravel:schedule`) and move the route runner to the `http` package as `HTTPRunner`, since the runner actually starts the HTTP server.

Closes https://github.com/goravel/goravel/issues/967

## Why
Today the only way to suppress the scheduler on `app.Start()` is the global `app.auto_run = false` flag, but that flag also disables the HTTP, queue, gRPC, and telemetry runners. Users who want to run the scheduler on a dedicated container or server had no way to keep the rest of the app running. This change adds a per-runner opt-out that the framework evaluates centrally, so any new runner added in the future is automatically selectable via the same config.

`app.auto_run` was never published in `goravel/goravel`'s example `config/app.go` or in `goravel/docs`, so removing it is safe — there is no documented user behavior depending on it. The new `disabled_runners` is more expressive: a glob pattern (`*`, `goravel:*`, or an exact match) decides which runners are skipped. The filter runs centrally in `foundation.configureRunners()`, so per-runner plumbing is not needed and new runners get the same behavior for free. A `*` pattern also serves as a kill switch — `disabled_runners: ["*"]` is the new "run nothing in the main process" semantic.

```go
// config/app.go on the web container
return map[string]any{
    "app": map[string]any{
        "env":   "production",
        "debug": false,
        "disabled_runners": []string{"goravel:schedule"},
    },
}

// Scheduler-only container — disable http, queue, grpc
"disabled_runners": []string{"goravel:http", "goravel:queue", "goravel:grpc"},

// Or shorthand: disable all framework runners
"disabled_runners": []string{"goravel:*"},

// Kill switch: run nothing
"disabled_runners": []string{"*"},
```


foundation/application.go
```diff
@@ -611,6 +612,26 @@ func (r *Application) configureRoutes() {
...
+	disabledRunners := []string{}
+	if config := r.MakeConfig(); config != nil {
+		disabledRunners = config.GetStringSlice("app.disabled_runners", []string{})
+	}
+	isDisabled := func(signature string) bool {
+		for _, pattern := range disabledRunners {
+			matched, err := stdpath.Match(pattern, signature)
+			if err != nil {
+				if log := r.MakeLog(); log != nil {
+					log.Warningf("invalid app.disabled_runners pattern %q: %v", pattern, err)
```
hwbrzzl: Add the error to `errors/list.go`

foundation/application.go
```diff
@@ -611,6 +612,26 @@ func (r *Application) configureRoutes() {
 }
 
 func (r *Application) configureRunners() {
+	disabledRunners := []string{}
+	if config := r.MakeConfig(); config != nil {
+		disabledRunners = config.GetStringSlice("app.disabled_runners", []string{})
+	}
+	isDisabled := func(signature string) bool {
```
hwbrzzl: Move this to a seperate function.

goravel-coder (commented):
Moved the closure to a new `isRunnerDisabled` method on `*Application` that takes the signature and the disabled-runner patterns as arguments.

goravel-coder (commented):
Added the new error to `errors/list.go` as `AppInvalidDisabledRunnersPattern` and used it in `application.go` via `errors.AppInvalidDisabledRunnersPattern.Args(pattern, err).Error()`.

foundation/application.go
goravel-coder: Added the new error to `errors/list.go` as `AppInvalidDisabledRunnersPattern` and used it in `application.go` via `errors.AppInvalidDisabledRunnersPattern.Args(pattern, err).Error()`.

foundation/application.go
goravel-coder: Moved the closure to a new `isRunnerDisabled` method on `*Application` that takes the signature and the disabled-runner patterns as arguments.

PR #1500, opened 2026-06-20
feat: [#959] add WithCommandsFilter builder to scope registered Artisan commands

### LINKED ISSUE #959: make sure the artisan can be opt in bundle output binary.

goravel-coder:
## Why

Closes goravel/goravel#959. The framework currently registers every Artisan dev command (about, env:*, package:*, vendor:publish, make:*, etc.) on every binary, including production server-only builds. There was no supported way to scope this down — `./app --help` in production revealed dev commands, and the surface was always the same regardless of environment.

## What

Introduces a new `WithCommandsFilter(func() []string)` builder method on `ApplicationBuilder`. The callback runs once at `Build()` time and returns the **positive list** of command signatures to keep. The list is captured on the `Application` and applied uniformly to both the framework's `defaultCommands` batch (`foundation/application.go:defaultCommands`) and the `console/console/*` batch registered by `console.ServiceProvider.Boot`.

Matching rules (signature-only — category is never consulted):
- exact match (no wildcard) — checked against `command.Signature()`
- glob match (entry contains `*`) — checked against `command.Signature()` via `path.Match` (`*` matches any sequence of non-`/` characters, `?` is a literal character)

Semantics:
| Callback returns | Result |
| --- | --- |
| method not called | keep every command (default — no migration) |
| `nil` | keep every command (no filter applied) |
| `[]string{}` | drop every command |
| entries | keep only commands whose signature matches |

The filter also trims the user-added commands registered via `WithCommands(...)`, so the user cannot bypass the filter by adding commands. The implementation of `add first, then filter` lives in two places — `foundation/application.go:configureCommands` filters user-added commands, and `console.ServiceProvider.Boot` filters its own batch.

## User-facing API

```go
return foundation.Setup().
    WithCommandsFilter(func() []string {
        if facades.Config().GetString("app.env") == "production" {
            return []string{
                "up", "down", "key:generate", "about",
                "env:*",     // glob
                "make*",     // glob (matches make, make:controller, make:test, ...)
                "package:*", // glob
                "vendor:publish",
            }
        }
        return nil // dev/staging: keep everything
    }).
    // ...
```

The user can also drive the list from build tags or `ldflags` instead of `facades.Config()` — the framework doesn't care, it just consumes the returned slice.

## Files

- `console/application.go` — new `FilterCommandsByAllowlist` helper (signature-only, `*`-only glob)
- `console/service_provider.go` — apply filter to the `console/console/*` batch via the new `Application.ConsoleCommandsFilter()` getter
- `foundation/application.go` — new `consoleCommandsFilter` field, `Build()` capture, `defaultCommands` and `configureCommands` apply the filter
- `foundation/application_builder.go` — new `WithCommandsFilter` method
- `contracts/foundation/application.go` + `application_builder.go` — interface extensions
- `mocks/foundation/Application.go` + `ApplicationBuilder.go` — regenerated via `go tool mockery`
- Tests added in `console/application_test.go`, `console/service_provider_test.go` (new), `foundation/application_test.go`, `foundation/application_builder_test.go`

## Breaking changes

None. Default behavior is unchanged for every existing app — `WithCommandsFilter` is opt-in.

## Test plan

- `go build ./...` — clean
- `go test ./...` — all pass (foundation, console, console/console, plus every downstream package that uses mocks/foundation)
- `go vet ./...` — clean
- New tests cover: nil/empty/exact/glob/mixed/question-mark-literal/whitespace-only/category-not-matched/`up`-`down`-by-signature, the "add first, then filter" composition between `WithCommands` and `WithCommandsFilter`, and the `nil`/`[]string{}`/entries distinction on both registration paths.

### DIFF
--- a/console/application.go
+++ b/console/application.go
@@ -86,8 +86,7 @@ func (r *Application) CallAndExit(command string) {
 	_ = r.Run(append(commands, strings.Split(command, " ")...), true)
 }
 
-// Register commands to the application. Filtering is the caller's
-// responsibility; this method appends without consulting any allowlist.
+// Register commands to the application.
 func (r *Application) Register(commands []console.Command) {
 	r.commands = append(r.commands, commands...)
 }
--- a/console/application.go
+++ b/console/application.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"os"
 	"os/signal"
+	stdpath "path"
 	"slices"
 	"strings"
 	"syscall"
@@ -173,6 +174,57 @@ func shutdownCommand(shutdownable console.Shutdownable, cmd *cli.Command, argume
 	shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), 30*time.Second)
 	defer shutdownCancel()
 	return shutdownable.Shutdown(NewCliContext(shutdownCtx, cmd, arguments))
+}
+
+// FilterCommandsByAllowlist returns the subset of commands whose Signature()
+// matches at least one entry in allowlist. Each entry is matched in one of
+// two ways:
+//
+//   - Exact match (no wildcard) — checked against command.Signature().
+//   - Glob match (the entry contains '*') — checked against
+//     command.Signature() using stdpath.Match. '*' matches any sequence of
+//     non-'/' characters. '?' is not a wildcard.
+//
+// Category is never consulted. The filter is signature-only.
+//
+// Special cases:
+//   - allowlist == nil            → every command is kept (no filter applied).
+//   - allowlist resolves to no usable entries → every command is dropped.
+func FilterCommandsByAllowlist(commands []console.Command, allowlist []string) []console.Command {
+	if allowlist == nil {
+		return commands
+	}
+	exact := make(map[string]struct{}, len(allowlist))
+	var globs []string
+	for _, s := range allowlist {
+		s = strings.TrimSpace(s)
+		if s == "" {
+			continue
+		}
+		if strings.Contains(s, "*") {
+			globs = append(globs, s)
+		} else {
+			exact[s] = struct{}{}
+		}
+	}
+	if len(exact) == 0 && len(globs) == 0 {
+		return nil
+	}
+	kept := make([]console.Command, 0, len(commands))
+	for _, cmd := range commands {
+		sig := cmd.Signature()
+		if _, ok := exact[sig]; ok {
+			kept = append(kept, cmd)
+			continue
+		}
+		for _, pattern := range globs {
+			if ok, _ := stdpath.Match(pattern, sig); ok {
+				kept = append(kept, cmd)
+				break
+			}
+		}
+	}
+	return kept
 }
 
 func commandsToCliCommands(commands []console.Command) ([]*cli.Command, error) {
--- a/console/application.go
+++ b/console/application.go
@@ -86,7 +86,8 @@ func (r *Application) CallAndExit(command string) {
 	_ = r.Run(append(commands, strings.Split(command, " ")...), true)
 }
 
-// Register commands to the application.
+// Register commands to the application. Filtering is the caller's
+// responsibility; this method appends without consulting any allowlist.
 func (r *Application) Register(commands []console.Command) {
 	r.commands = append(r.commands, commands...)
 }
@@ -177,15 +178,13 @@ func shutdownCommand(shutdownable console.Shutdownable, cmd *cli.Command, argume
 }
 
 // FilterCommandsByAllowlist returns the subset of commands whose Signature()
-// matches at least one entry in allowlist. Each entry is matched in one of
-// two ways:
+// matches at least one entry in allowlist.
 //
-//   - Exact match (no wildcard) — checked against command.Signature().
-//   - Glob match (the entry contains '*') — checked against
-//     command.Signature() using stdpath.Match. '*' matches any sequence of
-//     non-'/' characters. '?' is not a wildcard.
-//
-// Category is never consulted. The filter is signature-only.
+// Matching is signature-only (category never consulted):
+//   - Entries without '*' are exact-matched against command.Signature().
+//   - Entries containing '*' use glob matching via path.Match against
+//     command.Signature() (only '*' triggers the glob path; '?' is matched
+//     literally as an exact match).
 //
 // Special cases:
 //   - allowlist == nil            → every command is kept (no filter applied).
--- a/console/service_provider_test.go
+++ b/console/service_provider_test.go
@@ -63,9 +63,9 @@ func TestServiceProviderBootNilAllowlistRegistersAll(t *testing.T) {
 
 	// nil allowlist means all four console/console/* commands are registered.
 	app.EXPECT().MakeArtisan().Return(artisan).Once()
-	app.EXPECT().MakeConfig().Return(configFacade).Maybe()
-	app.EXPECT().MakeProcess().Return(processFacade).Maybe()
-	app.EXPECT().ConsoleCommandsFilter().Return(nil).Once()
+	app.EXPECT().MakeConfig().Return(configFacade).Once()
+	app.EXPECT().MakeProcess().Return(processFacade).Once()
+	app.EXPECT().CommandsFilter().Return(nil).Once()
 	artisan.EXPECT().Register(mock.MatchedBy(func(commands []contractsconsole.Command) bool {
 		return len(commands) == 4
 	})).Once()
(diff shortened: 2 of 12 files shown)


contracts/foundation/application.go
```diff
@@ -65,6 +65,9 @@ type Application interface {
 	Build() Application
 	// Commands register the given commands with the console application.
 	Commands([]console.Command)
+	// ConsoleCommandsFilter returns the captured positive-list of command
+	// signatures to keep (set via WithCommandsFilter). nil means no filter.
+	ConsoleCommandsFilter() []string
```
hwbrzzl:
```suggestion
	CommandsFilter() []string
```

console/service_provider_test.go
```diff
@@ -0,0 +1,125 @@
...
+	provider := &ServiceProvider{}
+	app := mocksfoundation.NewApplication(t)
+	artisan := mocksconsole.NewArtisan(t)
+	configFacade := mocksconfig.NewConfig(t)
+	processFacade := mocksprocess.NewProcess(t)
+
+	// nil allowlist means all four console/console/* commands are registered.
+	app.EXPECT().MakeArtisan().Return(artisan).Once()
+	app.EXPECT().MakeConfig().Return(configFacade).Maybe()
+	app.EXPECT().MakeProcess().Return(processFacade).Maybe()
```
hwbrzzl: Why use `Maybe`? It should be avoided. The same as below.

contracts/foundation/application_builder.go
```diff
@@ -22,6 +22,28 @@ type ApplicationBuilder interface {
...
+	//
+	// Semantics:
+	//   - Method not called           → keep every command (default).
+	//   - Callback returns nil        → keep every command (no filter).
+	//   - Callback returns []string{} → drop every command (filter, no matches).
+	//   - Callback returns entries    → keep only commands whose signature
+	//     matches an entry (exact or glob).
+	//
+	// The filter applies to user-added commands from WithCommands as well,
+	// so the user cannot bypass the filter by adding commands.
```
hwbrzzl: Simplify the annotation

foundation/application.go
```diff
@@ -68,16 +68,17 @@ func init() {
...
-	runnersToRun       []*RunnerWithInfo
+	ctx                   context.Context
+	cancel                context.CancelFunc
+	builder               *ApplicationBuilder
+	providerRepository    foundation.ProviderRepository
+	publishes             map[string]map[string]string
+	publishGroups         map[string]map[string]string
+	json                  foundation.Json
+	bootedRunners         []string
+	consoleCommandsFilter []string
```
hwbrzzl:
```suggestion
	commandsFilter []string
```

foundation/application_builder.go
```diff
@@ -67,6 +68,40 @@ func (r *ApplicationBuilder) WithCommands(fn func() []console.Command) foundatio
...
+//   - Callback returns nil        → keep every command (no filter).
+//   - Callback returns []string{} → drop every command (filter, no matches).
+//   - Callback returns entries    → keep only commands whose signature
+//     matches an entry (exact or glob).
+//
+// Composition with WithCommands:
+//
+// WithCommands adds extra commands to the framework's set; WithCommandsFilter
+// then trims the combined set. The filter applies to user-added commands
+// too, so the user cannot bypass the filter by adding commands.
```
hwbrzzl: Ditto

foundation/application_test.go
```diff
@@ -168,6 +171,143 @@ func (s *ApplicationTestSuite) TestDefaultCommandsDoesNotRegisterMaintenanceComm
...
+	s.app.Instance(binding.Route, mocksroute.NewRoute(s.T()))
+	s.app.consoleCommandsFilter = []string{"up", "down"}
+
+	commands := s.app.defaultCommands()
+	got := commandSignatures(commands)
+
+	s.Contains(got, "up")
+	s.Contains(got, "down")
+	s.NotContains(got, "about")
+}
```
hwbrzzl: This case is unnecessary, duplicated.

console/application.go
hwbrzzl, line 90: The commands are not filtered when called by other modules.

goravel-coder (commented):
## Review of PR #1500

Diff: https://critique.work/v/49ae9fcc92287dcc8b95ca92b00f9215

### What it does

Adds `WithCommandsFilter()` to the `ApplicationBuilder`, allowing users to scope which built-in Artisan commands get registered -- particularly useful for production-only binaries that don't need `make:*`, `package:*`, `vendor:publish`, etc.

### What's good

- **Comprehensive tests** -- table-driven tests cover nil/empty/whitespace/glob/exact/literal-`?` cases across both unit (`FilterCommandsByAllowlist`) and integration-level (`defaultCommands`, `configureCommands`, service provider) tests. All pass green.
- **Go vet clean** -- no issues.
- **Defensive nil check** -- `registerCommands` now checks `artisanFacade == nil` before proceeding.
- **Clean API** -- `WithCommandsFilter(func() []string)` is a typical builder pattern, callback delayed to `Build()` so it can read config/env.
- **Consistent semantics** -- `nil` allowlist = keep all, `[]string{}` = drop all, entries = keep only matches. Documented in both contract and implementation.
- **Filter applies everywhere** -- console SP commands, `defaultCommands()`, and user `WithCommands()` commands all go through the same gate.

### Issues

#### Medium: Docstring misleads about `path.Match` semantics

The docstring claims `'?' is not a wildcard`, which is technically true for this implementation (only `*` triggers the glob path), but the docstring also says matching uses `stdpath.Match` -- which does treat `?` as a wildcard. A user reading the docs might think `?` behaves like `path.Match`'s `?`, but it doesn't because the code only routes patterns containing `*` to `path.Match`. Patterns with `?` but no `*` go to exact match.

**Suggestion**: Clarify in the docstring that the glob path is only triggered by `*` presence, and patterns without `*` are exact-matched regardless of `?` content.

#### Low: Double Boot of service providers

`Boot()` (line 92) and `Build()` (line 106) both call `r.providerRepository.Boot(r)`. This means if a goravel app calls both `Boot()` and then `Build()`, console SP's `registerCommands` runs twice (once without filter, once with filter). This is pre-existing and not introduced by this PR, but worth noting since the filter now adds a behavioral difference between the two calls.

#### Low: `TestServiceProviderBootNilAllowlistRegistersAll` uses `.Maybe()` for config/process

The test at `console/service_provider_test.go:57` asserts the console SP registers exactly 4 commands when filter is nil. This is correct, but the test doesn't verify that `MakeConfig()` and `MakeProcess()` are actually called -- they use `.Maybe()` which can match 0 times. This is fine for testing filtering but could mask regressions if the underlying command constructors change their dependencies.

### Verdict

**Request changes** -- The docstring clarification about `?` behavior is the main ask. The code and tests themselves are solid.

contracts/foundation/application.go
goravel-coder: Renamed to CommandsFilter() throughout the codebase.

console/service_provider_test.go
goravel-coder: Replaced .Maybe() with explicit .Once() in all three Boot test functions.

foundation/application.go
goravel-coder: Renamed to commandsFilter.

contracts/foundation/application_builder.go
goravel-coder: Simplified the docstring. Also clarified that only * triggers glob matching and ? is matched literally as an exact match.

foundation/application_builder.go
goravel-coder: Done — simplified in the implementation as well.

foundation/application_test.go
goravel-coder: Removed the duplicate test.

console/application.go
goravel-coder: Added a comment clarifying that Register() appends without consulting the allowlist — filtering is the caller's responsibility. The filter is applied at the framework level (configureCommands, defaultCommands, and console service provider Boot).

hwbrzzl: No, the user commands should be filtered as well.

foundation/application_builder.go
```diff
@@ -26,6 +26,7 @@ type ApplicationBuilder struct {
 	commands                   func() []console.Command
 	config                     func()
 	configuredServiceProviders func() []foundation.ServiceProvider
+	commandsFilter      func() []string
```
hwbrzzl: The lint is incorrect.

console/application.go
goravel-coder: Done. Moved FilterCommandsByAllowlist into registerCommands() so ALL paths (defaultCommands, configureCommands, and Commands() called by any module) go through the same gate. Removed redundant filtering from defaultCommands() and configureCommands().

PR #1501, opened 2026-06-20
fix: SIGSEGV on artisan migrate when database connection is unavailable

goravel-coder:
## Summary
- Add `WithCommandsFilter` builder example to scope registered Artisan commands by environment
- Add integration tests verifying filter returns nil in local env and built-in commands survive
- Add production env test using config override + `Restart()` to verify filter activation

## Why
This adds an example demonstrating the `WithCommandsFilter` builder method introduced in goravel/framework#1500. The filter callback returns a positive list of command signatures to keep in production (using exact and glob matching), and returns nil to keep all commands in dev/local environments.

```go
func CommandsFilter() []string {
	if facades.Config().GetString("app.env") == "production" {
		return []string{
			"up",
			"down",
			"make:*",
			"vendor:publish",
		}
	}

	return nil
}
```

The filter is wired into the builder chain in `bootstrap/app.go`:

```go
return foundation.Setup().
	// ...
	WithCommands(Commands).
	WithCommandsFilter(CommandsFilter).
	// ...
```

### DIFF
--- a/database/console/model_make_command.go
+++ b/database/console/model_make_command.go
@@ -72,6 +72,11 @@ func (r *ModelMakeCommand) Extend() command.Extend {
 }
 
 func (r *ModelMakeCommand) Handle(ctx console.Context) error {
+	if orm := r.schema.Orm(); orm == nil || orm.Query() == nil {
+		ctx.Error(errors.SchemaOrmNotAvailable.Error())
+		return nil
+	}
+
 	m, err := supportconsole.NewMake(ctx, "model", ctx.Argument(0), support.Config.Paths.Models)
 	if err != nil {
 		ctx.Error(err.Error())
--- a/database/console/model_make_command.go
+++ b/database/console/model_make_command.go
@@ -72,8 +72,7 @@ func (r *ModelMakeCommand) Extend() command.Extend {
 }
 
 func (r *ModelMakeCommand) Handle(ctx console.Context) error {
-	if orm := r.schema.Orm(); orm == nil || orm.Query() == nil {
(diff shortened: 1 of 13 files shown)


database/migration/migrator.go
```diff
@@ -33,6 +33,16 @@ func NewMigrator(artisan console.Artisan, schema contractsschema.Schema, table s
 	}
 }
 
+func (r *Migrator) requireOrm() error {
+	orm := r.schema.Orm()
+	if orm == nil || orm.Query() == nil {
+		color.Warningln(errors.SchemaOrmNotAvailable.Error())
```
hwbrzzl: Is it necessary to print here given the error has been returned?

database/schema/schema.go
```diff
@@ -413,6 +413,15 @@ func (r *Schema) Orm() contractsorm.Orm {
 	return r.orm
 }
 
+func (r *Schema) RequireOrm() error {
```
hwbrzzl: remove it if it's not used.

database/console/model_make_command.go
```diff
@@ -72,6 +72,11 @@ func (r *ModelMakeCommand) Extend() command.Extend {
 }
 
 func (r *ModelMakeCommand) Handle(ctx console.Context) error {
+	if orm := r.schema.Orm(); orm == nil || orm.Query() == nil {
+		ctx.Error(errors.SchemaOrmNotAvailable.Error())
+		return nil
+	}
```
hwbrzzl: The code it duplicated, move it to a helper function.

hwbrzzl (commented):
You are using `Maybe` in some tests, it's unexpected.

PR #1512, opened 2026-06-27
fix: reset database connections when the application rebuilds

krishankumar01:
## 📑 Description

The database connection cache (`connectionToDB` in `database/driver`) is process-global and survives `app.Restart()`. `Restart` rebuilds the container, but the cached `*gorm.DB` and its `Instrument` live outside the container, so a rebuilt application keeps connections created under the previous configuration. With telemetry enabled this also means spans are emitted under the tracer the connection was first built with, not the current one.

**What is changing:**

* The database `ServiceProvider.Register` resets the connection cache before registering the bindings. `Register` runs on every build, so a restart reconnects with the current configuration. On the initial boot the cache is empty, so this is a no-op.
* The reset API is split so the behavior is safe under different callers:
  * `ResetConnections()` drops the cache and unregisters each `Instrument`'s pool-metrics callback, but leaves the pools open (a caller may still hold the returned `*gorm.DB`). This is what `Orm.Fresh()` uses at runtime.
  * `CloseConnections()` additionally closes each pool, and is called by `Register`. `Register` runs after `app.Shutdown()` has stopped the runners, so the pools are idle and safe to close.

**Resolves the review feedback from the previous attempt:**

* Pool leak: a rebuild now closes the previous pools via `CloseConnections` instead of leaking them.
* Close race: closing only happens on the quiesced restart path; the runtime `Orm.Fresh` path only drops, so a `*gorm.DB` handed out by `BuildGorm`'s lock-free fast path is never closed under a caller.
* Close errors: surfaced with `color.Warningln` instead of being swallowed.
* Metrics callback leak: `Instrument` now stores the `metric.Registration` and exposes `Shutdown()`, which unregisters it; both reset paths call it so a discarded `Instrument` stops observing (and referencing) its pool.

This supersedes #1509 (that branch had a polluted history).

## ✅ Checks

### DIFF
--- a/database/driver/gorm.go
+++ b/database/driver/gorm.go
@@ -184,7 +184,6 @@ func CloseConnections() {
 		if cached.db == nil {
 			continue
 		}
-		// DB() is the writer pool; dbresolver replica pools are internal and not closed here.
 		if sqlDB, err := cached.db.DB(); err == nil {
 			if err := sqlDB.Close(); err != nil {
 				color.Warningln("close database connection: " + err.Error())
--- a/database/driver/gorm.go
+++ b/database/driver/gorm.go
@@ -149,12 +149,46 @@ func BuildGorm(config config.Config, logger logger.Interface, pool database.Pool
 	return instance, instrument, nil
 }
 
-func ResetConnections() {
+func drainConnections() []cachedConnection {
 	connectionToDBLock.Lock()
 	defer connectionToDBLock.Unlock()
 
+	var stale []cachedConnection
 	connectionToDB.Range(func(key, value any) bool {
+		if cached, ok := value.(cachedConnection); ok {
+			stale = append(stale, cached)
+		}
 		connectionToDB.Delete(key)
 		return true
 	})
+
+	return stale
+}
+
+// ResetConnections drops the cached connections so a later build reconnects with
+// the current configuration, and unregisters their pool-metrics callbacks. The
+// pools are left open because a caller may still hold the *gorm.DB; use
+// CloseConnections when the connections are known to be idle.
+func ResetConnections() {
+	for _, cached := range drainConnections() {
+		cached.instrument.Shutdown()
+	}
+}
+
+// CloseConnections drops the cached connections, unregisters their callbacks and
+// closes the pools. Call it only when the connections are idle (e.g. the database
+// provider's Register, which runs after app.Restart has stopped the runners).
+func CloseConnections() {
+	for _, cached := range drainConnections() {
+		cached.instrument.Shutdown()
+		if cached.db == nil {
+			continue
+		}
+		// DB() is the writer pool; dbresolver replica pools are internal and not closed here.
+		if sqlDB, err := cached.db.DB(); err == nil {
+			if err := sqlDB.Close(); err != nil {
+				color.Warningln("close database connection: " + err.Error())
+			}
+		}
+	}
 }
--- a/database/driver/gorm.go
+++ b/database/driver/gorm.go
@@ -165,19 +165,16 @@ func drainConnections() []cachedConnection {
 	return stale
 }
 
-// ResetConnections drops the cached connections so a later build reconnects with
-// the current configuration, and unregisters their pool-metrics callbacks. The
-// pools are left open because a caller may still hold the *gorm.DB; use
-// CloseConnections when the connections are known to be idle.
+// ResetConnections drops the cached connections and unregisters their telemetry,
+// leaving the pools open for callers that may still hold the *gorm.DB.
 func ResetConnections() {
 	for _, cached := range drainConnections() {
 		cached.instrument.Shutdown()
 	}
 }
 
-// CloseConnections drops the cached connections, unregisters their callbacks and
-// closes the pools. Call it only when the connections are idle (e.g. the database
-// provider's Register, which runs after app.Restart has stopped the runners).
+// CloseConnections drops the cached connections, unregisters their telemetry and
+// closes the pools. Call it only when the connections are idle.
 func CloseConnections() {
 	for _, cached := range drainConnections() {
 		cached.instrument.Shutdown()
@@ -186,7 +183,7 @@ func CloseConnections() {
 		}
 		if sqlDB, err := cached.db.DB(); err == nil {
 			if err := sqlDB.Close(); err != nil {
-				color.Warningln("close database connection: " + err.Error())
+				color.Warningln("database: close connection: " + err.Error())
 			}
 		}
 	}
--- a/database/driver/gorm_test.go
+++ b/database/driver/gorm_test.go
@@ -60,10 +60,10 @@ func TestResetConnections_KeepsPoolOpen(t *testing.T) {
 
 	ResetConnections()
 
-	// The pool is dropped from the cache but left open for any caller still holding it.
-	if err := sqlDB.Ping(); err != nil {
-		assert.NotContains(t, err.Error(), "database is closed")
-	}
+	// The pool is dropped from the cache but not closed (a caller may still hold it).
+	err = sqlDB.Ping()
+	assert.Error(t, err)
+	assert.NotContains(t, err.Error(), "database is closed")
 
 	// The next build reconnects instead of returning the dropped instance.
 	second, _, err := BuildGorm(stubGormConfig(t), gormlogger.Discard, stubPool(), "primary", resolver)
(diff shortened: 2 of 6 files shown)


goravel-coder (changes requested):
## Verdict

The approach is sound and the implementation is largely correct. The split between `ResetConnections` (runtime, drop-only) and `CloseConnections` (restart, drop + close) is the right design — it matches the two call contexts and avoids closing a `*gorm.DB` that a caller may still hold from the lock-free fast path. The `Instrument.Shutdown` nil-receiver guard correctly handles the "telemetry disabled" case where `cachedConnection.instrument == nil`.

Requesting changes on a couple of real issues and flagging a few nits.

## Changes requested

### 1. `TestResetConnections_KeepsPoolOpen` leaks the first pool

In `database/driver/gorm_test.go`, `TestResetConnections_KeepsPoolOpen`:

```go
first, _, err := BuildGorm(...)
sqlDB, err := first.DB()
ResetConnections()              // drops "primary" from cache, leaves first's pool OPEN (intentional)
second, _, err := BuildGorm(...) // stores "primary" -> second
// t.Cleanup(CloseConnections) at end: drains "primary" -> second, closes second
```

`first` and its `sqlDB` are removed from the cache by `ResetConnections`, so the `t.Cleanup(CloseConnections)` at the end **cannot** find them — the pool is left open and leaked. The test deliberately verifies the pool stays open (good), but it must also arrange to close it. Add:

```go
t.Cleanup(func() {
    if db, err := first.DB(); err == nil {
        _ = db.Close()
    }
})
```

`TestCloseConnections_ClosesPool` is fine — `CloseConnections` closes the only cached entry and `t.Cleanup(CloseConnections)` is then a no-op.

### 2. `Instrument.Shutdown` only unregisters the pool callback — spans/histograms keep emitting from a discarded db

`Shutdown` zeroes only `poolRegistration`. But the gorm plugin on a cached `*gorm.DB` still holds the same `*Instrument`, whose `tracer`, `meter`, and `durationHist` are still bound to the telemetry provider that was alive when the connection was first built. After `app.Restart` rebuilds the telemetry provider, those three fields still point at the *previous* MeterProvider/Tracer.

For the `CloseConnections` path this is mostly moot — the pool is closed, so queries against a stale `*gorm.DB` fail before reaching `endSpan`. But for the **runtime `ResetConnections` path** (`Orm.Fresh`), the pool stays open, and any code that still references the freshly-discarded `*gorm.DB` (a race window the PR description explicitly tolerates) will keep emitting spans/metrics under the *old* meter — i.e. the exact staleness the PR is trying to fix for the pool callback applies symmetrically to the operation-duration histogram and tracer.

The PR description scopes itself to "pool-metrics callback leak", so this is arguably out of scope, but the asymmetry is worth at least a code comment. If you want to be thorough, `Shutdown` could also nil out `r.tracer`, `r.meter`, `r.durationHist`, and reset `r.poolObserved` so the next `active()` call re-resolves through `r.resolver()` against the *current* telemetry provider:

```go
func (r *Instrument) Shutdown() {
    if r == nil { return }
    r.mu.Lock()
    defer r.mu.Unlock()
    if r.poolRegistration != nil {
        _ = r.poolRegistration.Unregister()
        r.poolRegistration = nil
    }
    r.tracer = nil
    r.meter = nil
    r.durationHist = nil
    r.poolObserved = false
    // r.sqlDB retained so startPoolObservation can re-register on next active()
}
```

Nice to have rather than a blocker — please decide consciously and add a comment either way.

### 3. Closing in `Register` overstates the safety guarantee

The comment in `database/service_provider.go`:

> Register runs after the runners stop, so closing is safe.

This is accurate *for runners*: `app.Restart` → `Shutdown` blocks until every `RunnerWithInfo.running` is false before calling `Refresh` → `Register`. But `Shutdown` only waits on *runners*. If a non-runner goroutine (a user-spawned background worker that didn't register as a runner, or a long-lived worker pool the app started manually) still holds an active query against the cached `*gorm.DB`, `Close` will surface `sql: database is closed` mid-query. This is strictly better than the previous behavior (silent pool leak), but the comment slightly overstates the guarantee. Suggest rewording:

> Register runs after `app.Shutdown` has stopped the runners, so the framework-managed pools are idle and safe to close. Application-spawned background goroutines that hold a connection should drain themselves before triggering `app.Restart`.

## Minor nits

- `color.Warningln("close database connection: " + err.Error())` — fine as a log string (the AGENTS rule about `errors/list.go` applies to *returned* errors, not warnings), but the prefix reads slightly oddly as a standalone log line; consider `"database: close connection: " + err.Error()` for consistency with the other `color.Warningln("database telemetry: ...")` messages.
- `drainConnections` returns a slice that escapes the lock — correct, but a one-line comment noting "returns the drained entries; caller closes without holding the connection lock to avoid lock-ordering with `Instrument.mu`" would help future readers, since `Shutdown`/`Close` take `r.mu` and you deliberately don't hold `connectionToDBLock` across them.
- `observePool` now writes `r.poolRegistration` relying on the caller already holding `r.mu`. That invariant holds today (only `startPoolObservation` under `r.mu` calls it), but it's fragile. Either move the assignment into `startPoolObservation` where the lock ownership is obvious, or add a `// caller must hold r.mu` doc comment on `observePool`.
- `TestShutdownUnregistersCallback` calls `observePool` directly without the `r.mu` that the production caller holds — fine for a single-goroutine test, but worth a comment so nobody copies it as a concurrency example.

## Things I verified are correct

- Lock ordering: `drainConnections` holds `connectionToDBLock` only during the Range/Delete, releases it before calling `instrument.Shutdown()`/`sqlDB.Close()`. So there's no `connectionToDBLock` → `Instrument.mu` ↔ `Instrument.mu` → `connectionToDBLock` cycle. Good.
- `Shutdown`'s `if r == nil` mirrors the existing `SetDB` guard, so `ResetConnections`/`CloseConnections` calling `cached.instrument.Shutdown()` with `instrument == nil` is safe (the "telemetry disabled" cache entries).
- First-boot no-op: on the initial `Boot`, `connectionToDB` is empty, so `CloseConnections` in `Register` iterates nothing. Confirmed via `Boot` → `providerRepository.Register` calling `database.ServiceProvider.Register`.
- New entries created by a concurrent `BuildGorm` after `drainConnections` returns are not in `stale`, so they're correctly left alone.
- `RegisterCallback` error path now handled — the old `_, err = ...; return err` would have leaked the registration on the success path; the new code stores it. The `if err != nil { return err }` before the assignment is correct (no successful registration to keep).

## Tests

- New `TestCloseConnections_ClosesPool` and `TestResetConnections_KeepsPoolOpen` cover both reset paths and the "next build reconnects" assertion — good coverage.
- `TestShutdownUnregistersCallback` directly verifies the `Unregister` behavior end-to-end against the SDK meter provider.
- Coverage gap: there's no test asserting that `ResetConnections`/`CloseConnections` calling `Shutdown` on a cache entry whose `instrument == nil` (telemetry disabled) doesn't panic. Worth a small case given the nil-receiver path is load-bearing.
- No test for the actual `Register` → `CloseConnections` integration (i.e. that a rebuild reconnects). That's hard to unit-test in isolation and probably belongs in an integration test under `tests/`; acceptable to defer.

## Summary

The fix correctly addresses the core bug (process-global `connectionToDB` outliving `app.Restart`), the close/drop split is the right call, and the metrics-callback leak is properly fixed. Asking for the leaked-pool cleanup in `TestResetConnections_KeepsPoolOpen` (a real test-hygiene bug) and a decision + comment on finding #2 before approving; the rest are nits.

hwbrzzl:
@krishankumar01 Basically, LGTM. Please consider the AI review.

krishankumar01:
Thanks for the thorough review. Addressed in `0d88296b`:

**Changes requested**
1. `TestResetConnections_KeepsPoolOpen` leaked the first pool: fixed. `ResetConnections` drops it from the cache, so the test now closes it via its own `t.Cleanup`.
2. `Instrument.Shutdown` only unregistering the pool callback: done the thorough fix. `Shutdown` now also clears `tracer`, `meter`, `durationHist` and resets `poolObserved` (keeping `sqlDB`), so a still-referenced connection re-resolves against the current telemetry provider on its next `active()` instead of the one it was built with. Symmetric with the pool-callback fix.
3. `Register` comment overstating safety: reworded to drop the safety claim; it now just states the reconnect intent.

**Nits**
- Close-warning prefix aligned to `database: close connection: ...`.
- The lock-ordering and `r.mu` invariants are real but I kept them uncommented to match this codebase's minimal-comment style (the lock-ordering is also noted in your "verified correct" section).

**Coverage**
- Added a nil-instrument (telemetry-disabled) reset case so the nil-receiver `Shutdown` path is exercised.
- The `Register -> CloseConnections` rebuild integration is left to the example e2e, as you suggested.

krishankumar01:
> @krishankumar01 Basically, LGTM. Please consider the AI review.

@hwbrzzl Done, please take a look again

hwbrzzl (approved):
Thanks

PR #1537, opened 2026-08-17
fix(notification): register notification manager as a singleton

codedsultan:
## Description

The notification service provider was the only provider in the framework still using a transient `app.Bind(...)` instead of `app.Singleton(...)`. Every other provider registers its facade as a singleton.

With a transient binding, each resolution builds a fresh manager, so:
- Channels extended during `Register` are lost between resolutions.
- The manager handed to the dispatched job (via `registerJobs` → `*Manager` cast) can diverge from the one exposed through `facades.Notification()`.

## Changes

- Register the notification manager with `app.Singleton(...)` so a single shared instance is resolved consistently across facade access and job dispatch.
- Updated the corresponding service provider unit tests.

## Tests

- [x] `go test ./notification/...` passes.

### DIFF
--- a/notification/service_provider.go
+++ b/notification/service_provider.go
@@ -24,7 +24,7 @@ func (r *ServiceProvider) Relationship() binding.Relationship {
 }
 
 func (r *ServiceProvider) Register(app foundation.Application) {
-	app.Bind(binding.Notification, func(app foundation.Application) (any, error) {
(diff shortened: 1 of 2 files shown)


codedsultan:
@hwbrzzl this is a quick follow up addressed goravel-coder must fix comment in the slack package

hwbrzzl (approved):
LGTM

PR #1538, opened 2026-08-17
feat: fall back queued notification/broadcast retries to worker tries (Laravel parity)

goravel-coder:
## Summary
- Queued notifications and broadcasts that declare no retry policy (`Tries == 0` / not implementing the interface) now follow the queue worker's `Tries` config instead of failing after a single attempt (Laravel `Worker::markJobAsFailedIfWillExceedMaxAttempts` parity).
- `Backoff` is serialized into the queued payload unconditionally, so per-attempt delays now apply to worker-driven retries too (last value repeats).
- A notification/broadcast-declared `Tries` still overrides the worker's, and a cleared/invalid payload remains a safe single-shot fallback.

## Why
Previously `DispatchJob`/`BroadcastJob` treated a missing retry policy (`Tries == 0`) as single-shot and dropped `Backoff` from the payload whenever no positive `Tries` was declared. That diverged from Laravel, where `$job->maxTries() ?? $options->maxTries` lets the queue worker's tries config govern when the notification or broadcast declares none. `JobWithShouldRetry.ShouldRetry` now receives the worker's `maxTries`, and the notification/broadcast jobs fall back to it when they declare no policy of their own — so a worker configured with `Tries = 3` now retries queued notifications/broadcasts that don't opt into their own policy, honoring their backoff on every retry.

```go
package notifications

import (
	"time"

	"github.com/goravel/framework/contracts/notification"
)

// OrderFailed declares a backoff but no Tries: delivery retries are now
// governed by the queue worker's Tries config, and the backoff applies to
// every retry (Laravel parity).
type OrderFailed struct {
	OrderID string
}

func NewOrderFailed(orderID string) *OrderFailed {
	return &OrderFailed{OrderID: orderID}
}

func (r *OrderFailed) Via(notifiable notification.Notifiable) []string {
	return []string{notification.ChannelDatabase}
}

func (r *OrderFailed) ToDatabase(notifiable notification.Notifiable) map[string]any {
	return map[string]any{"order_id": r.OrderID, "status": "failed"}
}

func (r *OrderFailed) OnQueue() string      { return "" }
func (r *OrderFailed) OnConnection() string { return "" }

// Backoff applies to every retry — capped by Tries when declared, or by the
// queue worker's Tries config otherwise (last value repeats).
func (r *OrderFailed) Backoff(channel string) []time.Duration {
	return []time.Duration{time.Second, 2 * time.Second}
}
```

### DIFF
--- a/broadcasting/job_test.go
+++ b/broadcasting/job_test.go
@@ -327,6 +327,10 @@ func TestBroadcastJob_ShouldRetry(t *testing.T) {
 		})
 	}
 
+	// The following cases exercise stateful behavior — establishing a stale
+	// item, then clearing or overriding it via a second Handle call — rather
+	// than the single-shot decision the table above covers. They need multiple
+	// sequential Handle calls, so they stay as separate t.Run cases.
 	t.Run("invalid payload clears stale item", func(t *testing.T) {
 		job := &BroadcastJob{}
 		err := job.Handle("not-json")
--- a/broadcasting/job_test.go
+++ b/broadcasting/job_test.go
@@ -167,83 +167,150 @@ func TestBroadcastJob_ShouldRetry(t *testing.T) {
 	}
 
 	tests := []struct {
-		name    string
-		payload string
-		attempt int
-		err     error
-		want    bool
-		wantD   time.Duration
+		name     string
+		payload  string
+		attempt  int
+		maxTries int // the queue worker's tries (ignored when item.Tries > 0)
+		err      error
+		want     bool
+		wantD    time.Duration
 	}{
 		{
-			name:    "tries zero is single-shot",
-			payload: marshalItem(0, nil),
-			attempt: 1,
-			want:    false,
-			wantD:   0,
+			// No retry policy declared and the worker runs with Tries=1
+			// (the default): single-shot.
+			name:     "tries zero is single-shot",
+			payload:  marshalItem(0, nil),
+			attempt:  1,
+			maxTries: 1,
+			want:     false,
+			wantD:    0,
 		},
 		{
-			name:    "tries 3 first attempt retries",
-			payload: marshalItem(3, nil),
-			attempt: 1,
-			want:    true,
-			wantD:   0,
+			// No retry policy declared: the worker's Tries=3 governs.
+			name:     "tries zero defers to worker tries first attempt retries",
+			payload:  marshalItem(0, nil),
+			attempt:  1,
+			maxTries: 3,
+			want:     true,
+			wantD:    0,
 		},
 		{
-			name:    "tries 3 second attempt retries",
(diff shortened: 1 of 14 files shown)


goravel-coder:
Addressed the Paseo review comment on `contracts/broadcasting/broadcasting.go` ("Remove the Laravel flag. Others are the same."): removed the `(Laravel parity, Worker::markJobAsFailedIfWillExceedMaxAttempts)` references from all 8 doc-comment spots this PR introduced, across 6 files (`contracts/broadcasting/broadcasting.go`, `contracts/notification/notification.go`, `contracts/queue/job.go`, `broadcasting/job.go`, `broadcasting/application.go`, `notification/job.go`). Descriptive behavior text kept; pre-existing Laravel comparisons left untouched. Committed as `ad00b241` and pushed.

broadcasting/job_test.go
hwbrzzl, line 273: Can't the `t.Run` cases be merged into the table test above?

notification/job_test.go
```diff
@@ -222,9 +289,31 @@ func TestDispatchJob_ShouldRetry(t *testing.T) {
...
 			assert.Error(t, job.Handle(tt.payload))
 
-			retryable, delay := job.ShouldRetry(tt.err, tt.attempt)
+			retryable, delay := job.ShouldRetry(tt.err, tt.attempt, tt.maxTries)
 			assert.Equal(t, tt.want, retryable)
 			assert.Equal(t, tt.wantD, delay)
 		})
 	}
+
+	t.Run("invalid payload clears stale item", func(t *testing.T) {
```
hwbrzzl, line 302: Can't the `t.Run` case be merged into the table test above?

broadcasting/job_test.go
goravel-coder: These trailing `t.Run` cases test a different axis than the table above, so they can't be folded in cleanly. The table is single-shot (one `Handle` → one `ShouldRetry`), while these exercise stateful behavior: establish a stale item, then clear/override it via a *second* `Handle` call (invalid payload, wrong arity, successful handle, non-string arg) and assert `ShouldRetry` reflects the new state. Folding them in would need a second-payload field plus intermediate assertions per entry, complicating the struct more than it clarifies. Added a comment above the cases documenting the distinction — committed as `ff9ace7e` and pushed.

notification/job_test.go
goravel-coder: Same reasoning as the broadcasting test this mirrors — this case exercises stateful behavior (establish a stale item, then clear it via a second `Handle` call), not the single-shot decision the table covers, so it needs multiple sequential `Handle` calls and can't be a single table entry. Added a comment documenting this — committed as `ff9ace7e` and pushed.

hwbrzzl (approved):

PR #1543, opened 2026-09-01
feat(config): add typed getters and config inspection helpers

miladev95:
## Summary

Extends the `config` package's `Config` contract with commonly needed typedgetters and small inspection helpers, bringing the API closer to parity with
Laravel-style config usage.

## New methods on `contracts/config.Config`

Typed getters:
- `GetInt64(path string, defaultValue ...int64) int64`
- `GetFloat64(path string, defaultValue ...float64) float64`
- `GetTime(path string, defaultValue ...time.Time) time.Time`
- `GetIntSlice(path string, defaultValue ...[]int) []int`
- `GetStringMap(path string, defaultValue ...map[string]any) map[string]any`
- `GetStringMapString(path string, defaultValue ...map[string]string) map[string]string`

Inspection / mutation:
- `Has(path string) bool` — reports whether a path exists without forcing a default
- `Forget(path string)` — removes a runtime-added key
- `All() map[string]any` — returns the full config (useful for `/debug/config` routes, dumps, tests)

## Implementation notes

- Slice and map getters correctly fall back to the supplied default when the
  underlying viper value is `nil` (viper returns a nil slice/map for unset keys
  even when the key technically exists), matching the existing behavior of
  `GetStringSlice`.
- `GetTime` accepts `time.Time`, RFC3339 / RFC3339Nano, `2006-01-02 15:04:05`,
  and `2006-01-02` formats.
- `Forget` uses viper's `Set(path, nil)`, consistent with how viper represents
  missing values.

## Tests

- Added focused test cases for each new method in `config/application_test.go`,
  covering: defaults, missing keys, nil slices/maps, invalid time strings, and
  the lookup/mutation helpers.
- Mocks regenerated via `go tool mockery` (affects `mocks/config/Config.go` and
  `mocks/queue/Config.go`, which embeds `contracts/config.Config`).

## Verification

- `go build ./...`
- `go test ./config/...`
- `golangci-lint run ./config/... ./contracts/config/...` — 0 issues

## Backwards compatibility

This is a purely additive change to the `Config` interface. Existingimplementations and consumers are unaffected.

### DIFF
--- a/config/application.go
+++ b/config/application.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"os"
+	"reflect"
 	"time"
 
 	"github.com/go-viper/mapstructure/v2"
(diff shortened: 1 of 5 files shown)


goravel-coder:
Thanks for the PR! I reviewed this against `viper v1.21.0` source. Build ✅ and `go test ./config/...` ✅ pass. Overall the typed getters are correct and consistent with the existing `Get*` family, but there are two contract-level issues worth resolving before merge.

> ⚠️ Please double-confirm the two "Must Fix" items below before acting on them — they are based on viper's *internal* `find()` / `getSettings()` behavior, which is version-sensitive. It'd be worth a quick local sanity check (e.g. a scratch test that `Forget`s an `.env`-backed key) to confirm the diagnosis on the exact viper version you build with.

## Must Fix

### 1. `Forget` doesn't forget keys loaded from `.env` / env vars — silent no-op for the dominant config source
`config/application.go:186-189` (contract at `contracts/config/config.go:42-43`)

```go
func (app *Application) Forget(path string) {
	app.vip.Set(path, nil)
}
```

`viper.Set(path, nil)` only writes a `nil` value into the **override** layer. Viper's `find()` walks `override → pflags → env → config file → kvstore → defaults`, and a `nil` override value fails the `val != nil` guard, so it **falls through to the config-file layer**. As a result:

- `Forget` on an `Add()`-ed key works (the only case the test covers).
- `Forget("APP_KEY")`, `Forget("DB_PORT")`, etc. — anything from `.env` via `ReadInConfig` — is a **no-op**: `Has()` still returns `true`, `Get()` still returns the original value.
- It also leaves a stale `nil` entry in `v.override`, surfacing in `AllKeys()`.

The contract comment says "removes the given path from the configuration" — false for real-world config. There's no true `Delete` in viper, so either (a) track forgotten keys in a dedicated map consulted by `Has`/`Get`/`All`, or (b) at minimum correct the doc/contract and clearly warn that `Forget` only affects runtime-`Add`ed keys.

### 2. `All()` claims to return "a copy" but shares nested references with internal config state
`config/application.go:191-194` (contract at `contracts/config/config.go:44-45`)

`viper.AllSettings()` → `getSettings()` builds a *new top-level* map, but each leaf value is `v.Get(k)`, which returns the **same underlying map/slice reference**. So `All()["database"]` is a shared `map[string]any`; a caller mutating a nested value corrupts live config state — contradicting "returns a copy of the entire configuration". Either deep-copy the result, or change the doc to explicitly state shallow-copy / shared-nested-values semantics. The same aliasing hazard applies to the new `GetStringMap`/`GetStringMapString`/`GetIntSlice`/`GetStringSlice` getters (matching viper's own behavior, but worth documenting).

## Should Fix

### 3. `GetTime` default-on-invalid semantics are inconsistent with every other getter
`config/application.go:230`

For a key that **is set** but holds an unparseable/non-`time.Time` value, `GetTime` returns the caller's default. Every other getter (`GetInt`, `GetBool`, `GetDuration`, `GetInt64`, `GetFloat64`) returns the **cast/zero value** and ignores the default when set-but-invalid. The tests enshrine this asymmetry (`GetTime("INVALID_TIME", fallback) == fallback` vs `GetDuration("INVALID_DURATION", time.Second) == 0`). Pick one convention (likely: return `time.Time{}` for set-but-invalid, reserving the default for `!IsSet`).

### 4. Coverage gaps in newly-shipped behavior
`config/application_test.go`

- `GetStringSlice` behavior change is untested — the diff changes it so a set-but-nil value now falls back to the default (`application.go:139-141`); there is no `TestGetStringSlice` at all.
- `GetTime` supports 4 layouts but only 2 are tested (RFC3339Nano and `"2006-01-02 15:04:05"` are claimed-but-untested); the already-`time.Time` fast path (`:218-219`) and the non-string/non-time fallback (`:230`) are untested.
- `Forget` on a file-sourced key is untested (the gap behind Must Fix #1).
- `GetStringMap`/`GetStringMapString` nil-vs-empty-default distinction (`isZeroValue` semantics) untested.

## Nits

- Dead `; true` condition in `GetTime` (`application.go:222`) — should be a plain `for` loop; `time.RFC3339Nano` is also redundant since `RFC3339` already accepts fractional seconds.
- `sliceOrDefault` and `mapOrDefault` are byte-for-byte identical (`application.go:233-251`) and misnamed (they return the first *non-nil* value, not a "default"). Collapse into one; consider `convert.FirstNonNil[T any]` in `support/convert`.
- `isZeroValue` latent panic for `reflect.Array` (`application.go:259`) — `Value.IsNil()` panics on array values; drop the dead `Array`/`Chan`/`Pointer` cases.
- Pointless import churn (`contracts/config/config.go:3-5`) — revert to `import "time"`.
- Comment-style inconsistency — new `Get*` use the repo's terse style while `Has`/`Forget`/`All` use full sentences; `GetDuration` lacks its trailing period; `All` doc drifts between "as a map" / "as a map[string]any".
- `GetTime` layout slice allocated per call — hoist to a package-level `var`.
- Zero-time comparison via `s.Equal(time.Time{}, …)` (`application_test.go:326,330`) is `reflect.DeepEqual` (location-pointer-sensitive); prefer `IsZero()`/`.Equal`.

## Summary

The typed getters are correct and consistent with the existing `Get*` family; the reflection-based `sliceOrDefault`/`mapOrDefault` are justified (slices/maps aren't `comparable`, so `convert.Default` can't apply) and the cost is confined to the fallback path. The two blocking items are **`Forget`'s silent no-op on `.env`-backed keys** and **`All`'s false "copy" guarantee** — both are contract-level promises the implementation doesn't honor. Would suggest fixing those (plus the `GetTime` inconsistency and the coverage gaps) before merge.

PR #1546, opened 2026-09-01
feat: added support for registering embedded filesystems as view sources

### LINKED ISSUE #990: Proposal: Support Embedded Package Views with io/fs
### Discussed in https://github.com/goravel/goravel/discussions/988

<div type='discussions-op-text'>

<sup>Originally posted by **protibimbok** August 31, 2026</sup>
# Proposal: Support Embedded Package Views with `io/fs`

## Summary

Add support for loading package templates from Go's `io/fs.FS`, allowing Goravel packages to embed their views using `go:embed`.

Currently, Goravel's `LoadViewsFrom()` API accepts filesystem paths. This means package templates must remain available on the filesystem at runtime, which can be inconvenient for applications distributed as a single binary or deployed in minimal container environments.

This proposal introduces an additional API for registering an `fs.FS` while preserving the existing filesystem-based API for backward compatibility.

### Proposed API

```go
LoadViewsFrom(path string)
LoadViewsFromFS(fs fs.FS, root string)
```

A package could then embed its templates:

```go
package admin

import (
    "embed"

    "github.com/goravel/framework/contracts/foundation"
)

//go:embed views/*
var views embed.FS

type ServiceProvider struct{}

func (r *ServiceProvider) Boot(app foundation.Application) {
    facades.View().LoadViewsFromFS(views, "views")
}
```

The templates would become part of the compiled Go binary and would no longer need to be distributed separately.

---

## Motivation

Goravel packages commonly contain resources such as:

```text
package/
├── service_provider.go
├── views/
│   ├── index.tmpl
│   └── layout.tmpl
└── ...
```

With the current filesystem-based API, a package registers a directory:

```go
facades.View().LoadViewsFrom(path)
```

The view resolver subsequently searches that directory on disk.

This creates a deployment concern.

After:

```bash
go build
```

Go does not include arbitrary template files in the resulting binary. Therefore, the package's `views/` directory must also be present in the production environment.

This is particularly undesirable for applications intended to be distributed as a single executable.

Go already provides a standard mechanism for this use case through `embed.FS` and `io/fs`.

Supporting `fs.FS` would allow Goravel packages to use the native Go mechanism for packaging static resources.

---

# Proposed Design

## 1. Add `LoadViewsFromFS`

Extend the View contract with:

```go
LoadViewsFromFS(fs fs.FS, root string)
```

where `fs` is Go's standard:

```go
io/fs
```

package.

For example:

```go
facades.View().LoadViewsFromFS(views, "views")
```

The existing API remains unchanged:

```go
facades.View().LoadViewsFrom("/path/to/views")
```

This keeps existing applications and packages fully compatible.

---

## 2. Store filesystem and `fs.FS` sources separately

The view manager currently stores filesystem paths.

The internal representation could be extended to support both resource types.

Conceptually:

```go
type View struct {
    paths []string
    filesystems []ViewFS
}
```

where:

```go
type ViewFS struct {
    FS   fs.FS
    Root string
}
```

Alternatively, both could be represented by a common internal abstraction.

The important requirement is that the view resolver can search both:

```text
OS filesystem
    +
io/fs filesystem
```

without exposing implementation details through the public API.

---

# 3. Resolve templates through `fs.FS`

Go's `io/fs` provides a common interface for reading files.

For an embedded package:

```go
//go:embed views/*
var views embed.FS
```

the resolver can access:

```go
fs.ReadFile(views, "views/index.tmpl")
```

or:

```go
fs.Open(views, "views/index.tmpl")
```

The view engine should use the same resolution logic regardless of whether the underlying source is:

* the OS filesystem
* `embed.FS`
* another `fs.FS` implementation

This also makes the feature extensible to other filesystem implementations in the future.

---

# 4. Preserve application-over-package precedence

One of the useful properties of Goravel's current view system is that application views can override package views.

That behavior should remain unchanged.

For example:

```text
resources/views/
    admin/dashboard.tmpl

package:
    views/
        admin/dashboard.tmpl
```

When resolving:

```go
facades.View().Make("admin/dashboard")
```

the application version should continue to take precedence.

The intended resolution order is:

```text
1. Application filesystem views
          ↓
2. Registered filesystem package views
          ↓
3. Registered embedded package views
```

The exact ordering between package sources should follow registration order.

This preserves the existing behavior while adding embedded resources as another source.

---

# 5. Support nested templates

The implementation must support templates that reference other templates.

For example:

```text
views/
├── layouts/
│   └── app.tmpl
└── users/
    └── index.tmpl
```

where:

```gotemplate
{{ template "layouts/app.tmpl" . }}
```

or equivalent `define`/`template` constructs are used.

The resolver must ensure that all files belonging to a template set can be loaded from the same registered `fs.FS`.

This is particularly important for packages containing layouts and reusable partials.

Tests should cover:

* nested directories
* `define`
* `template`
* partials
* multiple package view sources
* application overrides

---

# 6. `go:embed` remains a package responsibility

Goravel should not attempt to discover or embed files from installed Go modules automatically.

The package author should explicitly define the embedded filesystem:

```go
//go:embed views/*
var views embed.FS
```

Goravel's responsibility is simply to consume the resulting `fs.FS`:

```go
facades.View().LoadViewsFromFS(views, "views")
```

This follows Go's existing package/resource model and avoids depending on the location of the Go module cache.

---

# Backward Compatibility

This proposal does not remove or change:

```go
LoadViewsFrom(path string)
```

Existing packages continue to work exactly as they do today.

For example:

```go
facades.View().LoadViewsFrom(
    "/some/package/views",
)
```

remains valid.

The new API is additive:

```go
facades.View().LoadViewsFromFS(
    views,
    "views",
)
```

Therefore, existing packages do not need to migrate immediately.

---

# Development Experience

One caveat of `go:embed` is that embedded resources are fixed at build time.

With filesystem views:

```text
edit template
      ↓
restart/reload
      ↓
new template
```

With embedded views:

```text
edit template
      ↓
rebuild
      ↓
new template
```

This is an inherent property of `go:embed` and should not prevent the feature from being added.

The existing `LoadViewsFrom()` API remains available for applications or packages that require filesystem-based development workflows.

A package could even choose between the two approaches depending on its requirements.

---

# Static Assets

This proposal intentionally focuses on **views/templates**.

It should not automatically introduce embedded static assets into the View API.

However, the same `fs.FS` approach could later be reused for package assets:

```go
//go:embed public/*
var assets embed.FS
```

This could eventually enable APIs such as:

```go
facades.Route().StaticFS(...)
```

or another appropriate mechanism.

Keeping assets separate from the initial implementation limits the scope of the change.

---

# Testing

The implementation should include tests covering at least:

### Filesystem views

Existing behavior must continue to work:

```go
LoadViewsFrom("/tmp/views")
```

### Embedded views

```go
//go:embed testdata/views/*
var views embed.FS

LoadViewsFromFS(views, "testdata/views")
```

### Application override

Given:

```text
resources/views/foo.tmpl
package/views/foo.tmpl
```

the application template must be selected.

### Package fallback

If the application template doesn't exist, the embedded package template should be selected.

### Multiple packages

Given:

```text
package A → views/foo.tmpl
package B → views/bar.tmpl
```

both templates should be resolvable.

### Nested templates

Test:

```text
views/
├── layouts/
│   └── app.tmpl
└── pages/
    └── home.tmpl
```

with nested template references.

### Missing templates

The existing error behavior should remain consistent regardless of whether the source is a filesystem path or `fs.FS`.

---

# Potential API Alternative

Instead of introducing:

```go
LoadViewsFromFS(fs fs.FS, root string)
```

the API could accept a generic filesystem abstraction:

```go
LoadViewsFrom(source ...)
```

However, this would make the existing API more complicated and potentially introduce breaking changes.

An additive API is preferable:

```go
LoadViewsFrom(path string)
LoadViewsFromFS(fs fs.FS, root string)
```

It is explicit, backwards-compatible, and maps naturally to Go's standard library.

---

# Benefits

This feature would provide several benefits:

* Package templates can be embedded into the final binary.
* Single-binary deployments become possible for packages using embedded views.
* Packages no longer need to depend on their Go module-cache location at runtime.
* Container images can be smaller and simpler.
* Resource loading uses Go's standard `io/fs` abstraction.
* Existing filesystem-based packages remain compatible.
* The design can later be reused for other package resources.

---

# Non-Goals

This proposal does not attempt to:

* Automatically embed existing package templates.
* Remove `LoadViewsFrom()`.
* Change Go's build behavior.
* Automatically embed application templates.
* Introduce an asset-management system.
* Change the existing view precedence semantics.
* Require all Goravel packages to use `go:embed`.

---

# Migration Example

### Current package

```go
func (r *ServiceProvider) Boot(app foundation.Application) {
    facades.View().LoadViewsFrom(
        packageViewsPath,
    )
}
```

The package must ensure that `packageViewsPath` exists in production.

### Embedded package

```go
package admin

import (
    "embed"

)

//go:embed views/*
var views embed.FS

func (r *ServiceProvider) Boot(app foundation.Application) {
    facades.View().LoadViewsFromFS(views, "views")
}
```

Now:

```bash
go build
```

produces a binary containing the package templates.

The production deployment only needs the resulting binary and any other external resources required by the application.

---

# Conclusion

Goravel's current `LoadViewsFrom()` API is filesystem-oriented. Adding `LoadViewsFromFS()` would allow packages to opt into Go's native `embed.FS` mechanism without disrupting existing packages.

The recommended approach is to make `io/fs.FS` a first-class view source while preserving the current filesystem API and application-over-package precedence.

This provides a relatively small, backwards-compatible change while making Goravel packages substantially easier to distribute as self-contained Go binaries.
</div>

protibimbok:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/990

Packages can now ship their views inside the binary with `go:embed` instead of
requiring template files on disk at runtime.

### Added

- `View.LoadViewsFromFS(fsys fs.FS, root string)` — registers an `fs.FS`
  (e.g. an `embed.FS`) as a package view source. `root` selects the directory
  inside the FS; pass `"."` to use the whole FS.
- `View.RegisteredViewFS() []fs.FS` — returns the registered filesystems in
  registration order, for HTTP drivers to load templates from.
- Errors `ViewFSRequired` and `ViewInvalidFSRoot` (panics on a nil FS or an
  invalid root, matching how a bad path fails today).

```go
//go:embed views/*
var views embed.FS

facades.View().LoadViewsFromFS(views, "views")
```

### Behaviour

- Each FS is rooted with `fs.Sub` at registration time, so every source looks
  the same to consumers: template paths are relative to `"."` (e.g.
  `layouts/app.tmpl`), including nested directories.
- `root` accepts loose forms (`""`, `./views`, `/views/`, `views\admin`) and is
  normalised to a valid `io/fs` path.
- `Exists()` also checks registered filesystems.
- Precedence is unchanged: **app views > `LoadViewsFrom` dirs >
  `LoadViewsFromFS`**, each in registration order. Existing `LoadViewsFrom`
  users are unaffected.

### Notes

- The `mocks/view` mock is regenerated for the new interface methods.
- Actual template loading from these filesystems lands in the HTTP driver;
  a companion `goravel/gin` PR follows once this is tagged.

## ✅ Checks


hwbrzzl:
Thanks, checking

protibimbok:
Not sure if you have noticed but please check https://github.com/goravel/gin/pull/242 too
Both are completed and goes together

goravel-coder:
## Review

Thanks for the PR — clean, well-tested change overall. A few findings below.

> **Please double-confirm each finding against your local environment before resolving.** Several of these (especially **#1** and **#2**) depend on Go stdlib and OS behavior (`fs.Sub` validation, `filepath.ToSlash`) that can vary by Go version / platform, so please verify the exact behavior on your setup rather than taking them at face value.

### Should Fix

**1. `LoadViewsFromFS` never verifies the root actually exists** — `view/view.go:63-66`

`fs.Sub` only enforces `fs.ValidPath` (rejects `..`/`.`/empty elements); it does **not** stat the directory. For `embed.FS` (which doesn't implement `SubFS`), `LoadViewsFromFS(fsys, "typo")` registers a silently-empty source with no error — every later `fs.Stat`/`fs.ReadFile` returns `fs.ErrNotExist`. This contradicts the contract doc ("panics if … root is not a valid fs path") and the `ViewInvalidFSRoot` name. Either `fs.Stat(sub, ".")` at registration and panic on `ErrNotExist`, or narrow the docs to "rejects `..`/absolute roots; nonexistent roots are accepted lazily."

**2. Backslash normalization is a no-op on non-Windows** — `view/view.go:119`

`filepath.ToSlash` is an identity on Linux/macOS, so the documented `views\admin` form is **not** converted there; the backslash survives as a literal path element and `fs.Sub` accepts it, silently registering a source rooted at a nonexistent name. The `"os separator"` test (`view/view_test.go:92,157`) is a no-op on non-Windows CI, so this is invisible to the current test matrix. Either add explicit `strings.ReplaceAll(p, "\\", "/")`, or drop the backslash claim from the doc comments.

**3. Panic vs silent divergence between sibling methods** — `view/view.go:50-54` vs `:56-71`

`LoadViewsFrom` silently appends any path; `LoadViewsFromFS` panics on nil/invalid root. Two methods doing the same conceptual thing now fail differently. Defensible, but worth reconciling or documenting the rationale.

### Nits

- **`view/view.go:40`** — `normalizeFSPath` runs unconditionally on every `Exists` miss, even when no FS sources are registered. Guard with `if len(r.filesystems) > 0`.
- **`view/view.go:40-45`** — `Exists("")`/`Exists(".")` return `true` once any FS is registered (`fs.Stat(sub, ".")` stats the root dir). Edge case; a cheap empty/dot guard closes it.
- **Rendering depends on the companion driver** — `Exists` is only half the story; actual `View().Make(...)` requires the gin/fiber driver to parse `RegisteredViewFS()`. That's deferred to a companion PR.
- **`contracts/view/view.go:17`** — godoc example `//go:embed views/*` embeds only direct children; nested dirs (`layouts/`, `partials/`) need `//go:embed views` or `views/**`.
- **`contracts/view/view.go:23`** — `RegisteredViewFS` reads singular next to plural `RegisteredViews`; consider `RegisteredViewFSs`.
- **`view/view_test.go:245-253`** — `setupAppViews` mutates process-global `support.RelativePath`/`support.Config.Paths.Resources`; race-free today (no `t.Parallel()`, verified with `-race`) but fragile.

### Verified correct

- Root-escape (`../views`) is correctly rejected via `fs.ValidPath` → `fs.ErrInvalid` → `ViewInvalidFSRoot`.
- `RegisteredViews`/`RegisteredViewFS` both `make`+`copy`; no mutation leaks.
- Mutex discipline is race-free; errors use the `New(...)` constructor; `gofmt`/`go vet` clean; mock regenerated correctly.

protibimbok:
@goravel-coder 
I have tested this in a real project, everything is working fine

hwbrzzl (approved):
Thanks. Great work! Could you add tests for this feature in goravel/example as well?

PR #1548, opened 2026-09-06
fix: [#996] guard the translation loaded map against concurrent access

### LINKED ISSUE #996: Concurrent use of facades.Lang() crashes the process with "concurrent map read and map write"
### Before feedback (在反馈之前)

### Operating System (操作系统)

MacOS

### Golang Version (Go 版本)

1.27.0

v1.18.0

Translating from more than one goroutine crashes the process:

```
fatal error: concurrent map read and map write

goroutine 54 [running]:
  translation/translator.go:224   isLoaded
  translation/translator.go:192   load
  translation/translator.go:159   getLine
  translation/translator.go:97    Get
```

This is not a race detector warning, it is a real `fatal error` that kills the
whole process. Running the reproduction below 60 times without `-race` produced
it **41 times**.

**Why the existing mutex does not help**

`loaded` is a package level map (`translation/translator.go:36`) but `mu` is a
field on `Translator` (`:27`), and `load` locks that field (`:189`).

The Lang binding is registered with `BindWith`
(`translation/service_provider.go:30`), so `facades.Lang(ctx)` builds a **new**
`Translator` on every call. Every one of them brings its own mutex while they
all read and write the same package level map, so the mutex guards nothing.

**A second, independent problem**

Even with a package level mutex the map would still be raced on, because
`getLine` reads it outside the lock entirely:

```go
err := t.load(locale, group)                      // takes and releases the mutex
...
keyValue := getValue(loaded[locale][group], key)  // reads the map with no lock at all
```

`translator.go:171`. Both places need fixing, not just the mutex.

**Affected versions**

Identical code in `master`, `v1.18.0` and `v1.17.x`.

**Secondary effect: the mutex is taken on every lookup**

`load` acquires the full mutex on every `getLine`, even when the group has long
been loaded, and `Get` calls `getLine` up to three times. Parallel lookups end
up slower per operation than sequential ones:

```
BenchmarkGetSequential-8   129.7 ns/op   48 B/op   3 allocs/op
BenchmarkGetParallel-8     226.6 ns/op   48 B/op   3 allocs/op
```

**Proposed fix**

Guard `loaded` with a package level `RWMutex`: a read locked fast path for
lookups, the write lock only when a group actually has to be loaded. Have
`load` return the group it resolved so `getLine` no longer reaches into the map
a second time.

I have this working locally, along with a regression test that fails 10 out of
10 runs under `-race` before the change. After it: 0 crashes in 60 runs, clean
under `-race`, and the full framework suite passes. As a side effect of the read
lock and the removed second lookup:

```
BenchmarkGetSequential-8    39.07 ns/op   16 B/op   1 allocs/op
BenchmarkGetParallel-8      81.82 ns/op   16 B/op   1 allocs/op
```

Happy to open the PR if you would like it.

Drop this into `translation/` and run it. No `-race` needed, the process dies on
its own.

```go
package translation

import (
	"context"
	"fmt"
	"sync"
	"testing"
)

type crashLoader struct{}

func (crashLoader) Load(locale, group string) (map[string]any, error) {
	return map[string]any{"hello": "world"}, nil
}

// Mirrors what facades.Lang(ctx) does: the Lang binding is BindWith, so every
// call builds a fresh Translator with its own mutex, while they all write the
// package level loaded map.
func TestConcurrentTranslate(t *testing.T) {
	var wg sync.WaitGroup
	for i := 0; i < 500; i++ {
		wg.Add(1)
		go func(i int) {
			defer wg.Done()

			tr := NewTranslator(context.Background(), nil, crashLoader{}, "en", "en", nil)
			_ = tr.Get(fmt.Sprintf("k%d", i%7))
			tr.SetLocale(fmt.Sprintf("l%d", i%13))
			_ = tr.Get("hello")
		}(i)
	}
	wg.Wait()
}
```

```
go test ./translation/ -run TestConcurrentTranslate -count=1
```

Run it in a loop to see the rate:

```
c=0; for i in $(seq 1 60); do
  go test ./translation/ -run TestConcurrentTranslate -count=1 2>&1 | grep -q "fatal error" && c=$((c+1))
done; echo "$c / 60"
```

darakanoit:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/996

Translating from more than one goroutine crashes the process. `loaded` is a
package level map (`translation/translator.go:36`) but `mu` is a field on
`Translator` (`:27`). The Lang binding is registered with `BindWith`, so
`facades.Lang(ctx)` builds a new `Translator` on every call, each with its own
mutex, while they all read and write the same package level map. The mutex
guarded nothing.

`getLine` made it worse by reading `loaded[locale][group]` (`:171`) outside any
lock at all, so a package level mutex on its own would not have been enough
either.

Reproduced without the race detector, 41 fatal errors in 60 runs:

```
fatal error: concurrent map read and map write
  translation/translator.go:224   isLoaded
  translation/translator.go:192   load
  translation/translator.go:159   getLine
  translation/translator.go:97    Get
```

Present in `master`, `v1.18.0` and `v1.17.x`.

### What changed

`loaded` is now a `sync.Map`. Its access pattern is the one `sync.Map`
documents itself for: a cache that only grows, where every entry is written
once and read many times. A plain mutex still serializes the cold path and
re-checks the cache after taking it, so a group is loaded exactly once, same as
before.

`load` returns the group it resolved, so `getLine` no longer reaches into the
map a second time. `isLoaded` keeps its signature and goes through the same
lookup helper. The per instance `mu` field is gone, it had no job left.

### Verification

| | before | after |
|---|---|---|
| fatal errors in 60 runs, no `-race` | 41 | 0 |
| `go test ./translation/ -race -count=3` | DATA RACE | clean |
| added regression test, 10 runs under `-race` | fails 10/10 | passes |

Full framework suite passes.

Taking the read lock instead of the full mutex, and dropping the second map
lookup, also shows up in the numbers:

```
go test ./translation/ -bench='Get' -benchmem -run=^$
```

before:
```
BenchmarkGetSequential-8   129.7 ns/op   48 B/op   3 allocs/op
BenchmarkGetParallel-8     226.6 ns/op   48 B/op   3 allocs/op
```

after:
```
BenchmarkGetSequential-8    49.75 ns/op   16 B/op   1 allocs/op
BenchmarkGetParallel-8      13.19 ns/op   16 B/op   1 allocs/op
```

The shape matters more than the absolute numbers. Parallel lookups used to be
2.1x slower per operation than sequential ones, which is what a mutex taken on
every lookup looks like. They are now 3.8x faster, which is what scaling across
cores looks like.

I also measured a `RWMutex` over the existing nested map. It fixes the crash
just as well, but it does not scale: 39.07 ns sequential and 81.82 ns parallel,
degrading further as cores are added, because every reader writes to the shared
reader count. Happy to switch to that if you would rather keep the map shape.

### Scope

* `translation/translator.go`: the fix
* `translation/translator_test.go`: regression test, and two call sites updated
  for the new `load` signature

No public API changed. `load` and `isLoaded` are unexported.

## ✅ Checks

`TestConcurrentGet` builds translators the way `facades.Lang()` does and
resolves from 200 goroutines. It fails 10 out of 10 runs under `-race` without
the fix.

### DIFF
--- a/translation/translator.go
+++ b/translation/translator.go
@@ -24,16 +24,23 @@ type Translator struct {
 	selector   *MessageSelector
 	locale     string
 	fallback   string
-	mu         sync.Mutex
 }
 
-// loaded is a map structure used to store loaded translation data.
-// It is organized as follows:
-//   - First map (map[string]): Maps from locale to...
-//   - Second map (map[string]): Maps from folder(group) to...
-//   - Third map (map[string]): Maps from key to...
-//   - Value (any): The translation line corresponding to the key in the specified locale, folder(group), and key hierarchy.
-var loaded = make(map[string]map[string]map[string]any)
+// loaded stores the translation data that has been read from a loader. The key
+// is the locale and the folder(group) it came from, the value is the map from
+// translation key to line.
+//
+// The state is package level because the Lang binding is registered with
+// BindWith: every facades.Lang() call builds a new Translator, so per-instance
+// state would guard nothing.
+//
+// loaded is a sync.Map because the access pattern is the one it is built for, a
+// cache that only grows where every entry is written once and read many times.
+// loadMu serializes the cold path only, so a group is still loaded exactly once.
+var (
+	loaded sync.Map
+	loadMu sync.Mutex
+)
 
 // contextKey is an unexported type for keys defined in this package.
 type contextKey string
@@ -156,7 +163,7 @@ func (t *Translator) SetLocale(locale string) context.Context {
 }
 
 func (t *Translator) getLine(locale string, group string, key string, options ...translationcontract.Option) string {
-	err := t.load(locale, group)
+	translations, err := t.load(locale, group)
 	if err != nil {
 		if errors.Is(err, errors.LangFileNotExist) {
 			return ""
@@ -168,7 +175,7 @@ func (t *Translator) getLine(locale string, group string, key string, options ..
 		return ""
 	}
 
-	keyValue := getValue(loaded[locale][group], key)
+	keyValue := getValue(translations, key)
 	// If the key doesn't exist, return empty string.
 	if keyValue == nil {
 		return ""
@@ -185,17 +192,22 @@ func (t *Translator) getLine(locale string, group string, key string, options ..
 	return line
 }
 
-func (t *Translator) load(locale string, group string) error {
-	t.mu.Lock()
-	defer t.mu.Unlock()
+func (t *Translator) load(locale string, group string) (map[string]any, error) {
+	if translations, ok := lookupLoaded(locale, group); ok {
+		return translations, nil
+	}
 
-	if t.isLoaded(locale, group) {
-		return nil
+	loadMu.Lock()
+	defer loadMu.Unlock()
+
+	// Another goroutine may have loaded the group while this one waited.
+	if translations, ok := lookupLoaded(locale, group); ok {
+		return translations, nil
 	}
 
 	// Check if no loaders are available
 	if t.fileLoader == nil && t.fsLoader == nil {
-		return errors.LangNoLoaderAvailable
+		return nil, errors.LangNoLoaderAvailable
 	}
 
 	var (
@@ -210,26 +222,30 @@ func (t *Translator) load(locale string, group string) error {
 		translations, err = t.fsLoader.Load(locale, group)
 	}
 	if err != nil {
-		return err
+		return nil, err
 	}
 
-	if loaded[locale] == nil {
-		loaded[locale] = make(map[string]map[string]any)
+	loaded.Store(loadedKey(locale, group), translations)
+	return translations, nil
+}
+
+func loadedKey(locale string, group string) string {
+	return locale + "\x00" + group
+}
+
+func lookupLoaded(locale string, group string) (map[string]any, bool) {
+	value, ok := loaded.Load(loadedKey(locale, group))
+	if !ok {
+		return nil, false
 	}
-	loaded[locale][group] = translations
-	return nil
+
+	translations, ok := value.(map[string]any)
+	return translations, ok
 }
 
 func (t *Translator) isLoaded(locale string, group string) bool {
-	if _, ok := loaded[locale]; !ok {
-		return false
-	}
-
-	if _, ok := loaded[locale][group]; !ok {
-		return false
-	}
-
-	return true
+	_, ok := lookupLoaded(locale, group)
+	return ok
 }
 
 func makeReplacements(line string, replace map[string]string) string {
--- a/translation/translator_test.go
+++ b/translation/translator_test.go
@@ -2,6 +2,8 @@ package translation
 
 import (
 	"context"
+	"fmt"
+	"sync"
 	"testing"
 	"testing/fstest"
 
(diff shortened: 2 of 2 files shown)


goravel-coder:
# 🤖 Automated review

## Summary
Converts the package-level `loaded` translation cache from a nested `map` (guarded by a useless per-instance mutex) to a `sync.Map` with a package-level `loadMu` for the cold path, and makes `load` return the resolved group. The fix is correct, race-free, and soundly justified; no Must-Fix issues.

## Verdict
- Must Fix: 0 · Should Fix: 1 · Nits: 7

## Findings

### Should Fix
- [ ] `translation/translator.go:246` **`isLoaded` is now dead production code** — the only production caller (`load`) goes through `lookupLoaded` directly, so `isLoaded` survives only for `TestIsLoaded` (`translator_test.go:522-528`). It's a dangling remnant of the old API. Suggestion: delete it and update `TestIsLoaded` to assert via `lookupLoaded` (mirroring `TestLoad`), or keep it with a `// test-only` comment.

### Nits
- [ ] `translation/translator.go:232` **Magic `"\x00"` separator is undocumented** — the delimiter is a defensible, collision-safe choice (groups legitimately contain `/`, e.g. `foo/test`), but the rationale isn't stated. Suggestion: add a one-line comment explaining why a delimiter is needed and that NUL cannot appear in locale/group values, so a future reader doesn't "simplify" it to plain concatenation.
- [ ] `translation/translator_test.go:498` **`TestLoad` discards the new return value of `load`** — it does `_, err := translator.load(...)` then re-reads global state via `lookupLoaded`. The return value is now the contract `getLine` depends on. Suggestion: assert it directly — `translations, err := translator.load(...)` plus `s.Equal(tt.expected, translations)`.
- [ ] `translation/translator_test.go:554` **Assertions inside goroutines** — `s.Equal` fires from 200 goroutines. It's safe here (`suite.Equal` → `assert`, not `require`), but failures become hard to read. Suggestion: collect results and assert after `wg.Wait()`.
- [ ] `translation/translator_test.go:544` **Redundant `loaded.Clear()`** — `SetupTest` already clears `loaded` before every test method (`:36`), so the extra call in `TestConcurrentGet` is unnecessary. Suggestion: remove it.
- [ ] `translation/translator.go:39` **Comment slightly overstates single-flight** — "a group is still loaded exactly once" holds only for successful loads; failures are not cached and re-attempted on every call (pre-existing behavior). Suggestion: add a clarifying word that this applies to successful loads.
- [ ] `translation/translator.go:200` **Global `loadMu` serializes cold-path I/O across distinct groups** — distinct `(locale, group)` misses now block each other during file I/O. This is the correct correctness tradeoff and bounded to cache misses. Suggestion: only if startup latency ever matters, a per-key `singleflight` would restore parallel cold loads.
- [ ] `translation/translator.go:232` **Cache key omits loader identity** — `loaded` is package-global but keyed only by `locale+group`, so two `Translator` instances built with different loaders would share cache entries for the same key (pre-existing keying). Suggestion: consider scoping the cache to a shared loader/repository singleton while redesigning.

## Automated Checks
- [ ] `gofmt -d .` — clean
- [ ] `go vet ./translation/` — clean
- [ ] `go test ./translation/ -race -count=1` — pass

PR #1549, opened 2026-09-06
fix: [#997] stop Memory.Flush from replacing the sync.Map it shares with readers

### LINKED ISSUE #997: Memory.Flush() kills the process with "fatal error: sync: unlock of unlocked mutex"
### Before feedback (在反馈之前)

### Operating System (操作系统)

MacOS

### Golang Version (Go 版本)

1.27.0

v1.18.0

`Memory.Flush()` (`cache/memory.go:85`) replaces the whole map:

```go
func (r *Memory) Flush() bool {
	r.instance = sync.Map{}
	return true
}
```

`sync.Map` carries its own mutex. The assignment swaps that mutex out from
under a goroutine that is currently inside `Load`. The reader reaches its
unlock, finds a mutex that is no longer locked, and the runtime kills the
process:

```
fatal error: sync: unlock of unlocked mutex
```

This is not a race detector warning, it is a check built into the Go runtime.
The standalone test below crashed **15 of 20 runs without `-race`** on my
machine. The same scenario written against the existing `MemoryTestSuite`
crashed 20 of 20. Under `-race` it additionally reports a write at
`memory.go:85` against concurrent readers.

### How an application reaches this

`Flush` is public API, reachable as `facades.Cache().Flush()`. One request
clearing the cache while others read from it is enough.

### Fix

```diff
 func (r *Memory) Flush() bool {
-	r.instance = sync.Map{}
+	r.instance.Clear()
 	return true
 }
```

`sync.Map.Clear` has been available since Go 1.23 and `go.mod` asks for 1.26.
It drops every entry without replacing the value, so the mutex the readers hold
stays intact.

Verified: 0 crashes in 20 runs, clean under `-race`, the existing `cache` suite
and the full framework suite pass.

Happy to send the PR.

### Affected versions

`master`, `v1.18.0` and `v1.17.x` carry identical code.

Drop this into `cache/` and run it. No `-race` needed, the process dies on its
own.

```go
package cache

import (
	"sync"
	"testing"
)

func TestFlushWhileReading(t *testing.T) {
	m := &Memory{}
	_ = m.Put("k", "v", NoExpiration)

	var wg sync.WaitGroup
	for range 50 {
		wg.Add(1)
		go func() {
			defer wg.Done()
			for range 500 {
				_ = m.Get("k")
			}
		}()
	}
	for range 10 {
		wg.Add(1)
		go func() {
			defer wg.Done()
			for range 200 {
				m.Flush()
			}
		}()
	}
	wg.Wait()
}
```

```
go test ./cache/ -run TestFlushWhileReading -count=1
```

Run it in a loop to see the rate:

```
c=0; for i in $(seq 1 20); do
  go test ./cache/ -run TestFlushWhileReading -count=1 2>&1 | grep -q "fatal error" && c=$((c+1))
done; echo "$c / 20"
```

darakanoit:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/997

`Memory.Flush()` assigned a fresh `sync.Map` over `r.instance`:

```go
func (r *Memory) Flush() bool {
	r.instance = sync.Map{}
	return true
}
```

`sync.Map` carries its own mutex, so that assignment swaps the mutex out from
under a goroutine sitting inside `Load`. The reader reaches its unlock, finds a
mutex that is no longer locked, and the runtime kills the process:

```
fatal error: sync: unlock of unlocked mutex
```

Not a race detector warning, a check built into the runtime. The added test
crashed 20 of 20 runs without `-race`. Under `-race` it also reports a write at
`memory.go:85` against concurrent readers.

`Flush` is public API, reachable as `facades.Cache().Flush()`, so one request
clearing the cache while others read from it is enough to hit this.

Present in `master`, `v1.18.0` and `v1.17.x`.

### What changed

`sync.Map.Clear` drops every entry without replacing the value, so the mutex the
readers hold stays intact. It has been available since Go 1.23 and `go.mod` asks
for 1.26.

### Verification

| | before | after |
|---|---|---|
| `TestFlushWhileReading`, 10 runs, no `-race` | 10 fatal errors | 0 |
| `TestFlushWhileReading` under `-race` | DATA RACE at `memory.go:85` | clean |
| `go test ./cache/` | | pass |

`go test ./cache/ -race` for the whole package still fails on master because of
`TestIncrementWithConcurrent` and `TestDecrementWithConcurrent`, which share one
`err` variable across goroutines. That is unrelated to this change.

The existing `cache` suite and the full framework suite pass.

### Scope

* `cache/memory.go`: one line, plus a comment on why
* `cache/memory_test.go`: regression test

No public API changed.

## ✅ Checks

`TestFlushWhileReading` reads one key from 50 goroutines while 10 others call
`Flush`. It crashes every run against the old code, without the race detector.

### DIFF
--- a/cache/memory.go
+++ b/cache/memory.go
@@ -82,7 +82,7 @@ func (r *Memory) Forget(key string) bool {
 
 // Flush Remove all items from the cache.
 func (r *Memory) Flush() bool {
-	r.instance = sync.Map{}
+	r.instance.Clear()
 	return true
 }
 
--- a/cache/memory_test.go
+++ b/cache/memory_test.go
@@ -111,6 +111,36 @@ func (s *MemoryTestSuite) TestFlush() {
 	s.False(s.memory.Has("test-flush"))
 }
 
+// TestFlushWhileReading guards against replacing the sync.Map itself. Assigning
+// a fresh sync.Map over it swaps out the mutex the readers inside Load are
+// holding, and the runtime kills the process with "sync: unlock of unlocked
+// mutex". Before the fix this crashed 13 out of 20 runs, without the race
+// detector.
+func (s *MemoryTestSuite) TestFlushWhileReading() {
+	s.Nil(s.memory.Put("test-flush-while-reading", "goravel", NoExpiration))
+
+	var wg sync.WaitGroup
+	for range 50 {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for range 500 {
+				s.memory.Get("test-flush-while-reading")
+			}
+		}()
+	}
+	for range 10 {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for range 200 {
+				s.memory.Flush()
+			}
+		}()
+	}
+	wg.Wait()
+}
+
 func (s *MemoryTestSuite) TestGet() {
 	s.Nil(s.memory.Put("name", "Goravel", 1*time.Second))
 	s.Equal("Goravel", s.memory.Get("name", "").(string))


goravel-coder:
# 🤖 Automated review

## Summary
One-line change to `Memory.Flush()`: it now calls `r.instance.Clear()` instead of reassigning `r.instance = sync.Map{}`, eliminating a data race on the `instance` field and the fatal `sync: unlock of unlocked mutex` panic. The fix is correct and complete — the field is never written after construction, and `Clear()` (added in Go 1.20) is well within the module's `go 1.26.0` floor. Verified passing under `-race`.

## Verdict
- Must Fix: 0 · Should Fix: 2 · Nits: 2
- Out of scope (pre-existing, follow-up): 4

## Findings

### Should Fix
- [ ] `cache/memory_test.go:119` **Test asserts nothing — it's a probabilistic crash guard** — the test only fails by *crashing*; on success it verifies no state, and its own comment concedes the old code "crashed 13 out of 20 runs," so a plain CI run can silently pass a regressed build ~35% of the time. The post-condition is deterministic here (the key is stored once with `NoExpiration`, `Get` never re-inserts, and every flusher runs `Clear` at least once). Suggestion: after `wg.Wait()` add `s.False(s.memory.Has("test-flush-while-reading"))`.
- [ ] `cache/memory.go:84` **Add a comment explaining why `Clear()` and not reassignment** — the invariant ("never replace the `sync.Map`, its unexported mutex is shared with concurrent `Load`/`Store` callers") currently lives only in the test's doc comment. `Flush` is the method most likely to be "simplified" back into `r.instance = sync.Map{}` by a future refactor, and `go vet copylocks` won't catch it because `sync.Map`'s mutex is unexported. Suggestion: add a one-line comment on `Flush` stating that `Clear()` mutates in place to preserve the mutex shared with readers.

### Nits
- [ ] `cache/memory_test.go:114` **Doc comment wording could be tightened** — "the runtime kills the process with `sync: unlock of unlocked mutex`" is really a *panic*, and the empirical "13 out of 20 runs" will go stale. Suggestion: describe it as a panic and note that failure is deterministic under `go test -race`.
- [ ] `cache/memory_test.go:128` **Discarded `Get` result** — the return value is intentionally dropped (crash guard). Suggestion: assign to `_` to make the intent explicit and silence `errcheck`/`staticcheck` if they're ever enabled.

### Out of Scope (pre-existing — same class: shared mutable state)
- [ ] `cache/memory.go:236` **`WithContext` mutates a shared singleton's `ctx` field, which is never read** — the `contracts/cache` interface documents `WithContext` as returning "a new Cache instance," but `Memory` mutates `r` in place and returns the same pointer, and `http/limit/store.go` calls it per-request on a long-lived driver. It's a latent data race the moment anything reads `ctx` (today a dead field). Suggestion: make it a no-op like `filesystem/local.go`, or derive a new instance.
- [ ] `cache/memory.go:31` **`Add` arms `time.AfterFunc` before `LoadOrStore` resolves** — the timer fires even when the key already existed (`Add` returns `false`), so a failed second `Lock.Get` still schedules a `Forget` that can silently release another goroutine's held lock early. Suggestion: only schedule the timer when `!loaded`, or `LoadOrStore` first and arm the timer on success.
- [ ] `cache/memory_test.go:78` **testify assertions called inside goroutines + race on shared `err`** — `TestIncrementWithConcurrent`/`TestDecrementWithConcurrent` call `s.Nil(err)` inside goroutines (testify helpers must run on the test goroutine) and race on a shared captured `err`. The new `TestFlushWhileReading` correctly avoids this pattern. Suggestion: capture `err` per-goroutine and assert after `wg.Wait()`.
- [ ] `cache/memory.go:186` **Expiration timers are never stored or cancelled** — `time.AfterFunc` return values are discarded in `Add`/`Put`, so `Forget` can fire after a `Flush` re-populates a key, and the timers are unreclaimable until they fire. Harmless to this fix but worth addressing alongside the `Add` ordering bug above.

## Automated Checks
- [ ] `go test ./cache/ -run TestMemoryTestSuite -count=1` — pass
- [ ] `go test ./cache/ -run TestMemoryTestSuite/TestFlushWhileReading -race -count=1` — pass (old code fails this race check)
- [ ] `gofmt -d .` — clean
- [ ] `go vet ./cache/` — clean

darakanoit:
Thanks for the review. Addressed in f3935b40.

**Should fix**

1. Added `s.False(s.memory.Has(...))` after `wg.Wait()`. Worth noting it does not catch a regression on its own: the old `r.instance = sync.Map{}` also empties the map. The test still guards by crashing, and against the old code it crashes 10 of 10 runs here.
2. Added a one-line comment on `Flush` explaining why it clears in place.

**Nits**

3. Kept the wording. `sync: unlock of unlocked mutex` is raised through `runtime.fatal`, not `panic`, so `recover` cannot catch it and the whole process exits. Dropped the "13 out of 20" figure from the comment.
4. Left as is. `Get` returns `any`, not `error`, so errcheck does not apply.

Also, `sync.Map.Clear` was added in Go 1.23, not 1.20. Either way it is covered by `go 1.26`.

**Out of scope**

The timer findings are real. I reproduced them against master: a failed `Lock.Get` with a shorter TTL releases a lock someone else holds, and an old TTL timer removes a key after `Flush` or after an overwrite with a longer TTL. Opened https://github.com/goravel/goravel/issues/1003 for them rather than widening this PR.

hwbrzzl (approved):
Thanks, LGTM

PR #1552, opened 2026-09-07
fix: [#960] make cache lock optional for the database queue driver

### LINKED ISSUE #960: Database queue driver should not require a cache backend (leaky dependency)
## Summary
The database queue driver (`queue/drivers/database/database.go`) hard-requires a cache backend and acquires `cache.Lock(...)` around every `Pop()`. The lock wraps a `LockForUpdate()` SELECT, which on Postgres/MySQL already serializes row reservation atomically. Net effect: every queue poll generates two cache writes (SET NX EX + DEL) for no correctness benefit, and forces an external cache (e.g. Redis) on users who deliberately chose `database` to avoid one.   
                                                                                
## Repro                                                                      
1. Set `QUEUE_CONNECTION=database` and `CACHE_DRIVER=redis`.
2. Run any queue consumer.                                                    
3. Observe `MONITOR` on Redis: every poll cycle emits `SET goravel:queue-database-<queue>:lock 1 EX 60 NX` and `DEL`.            
                                                                                
## Why redundant                                                              
The DB transaction inside the lock uses `LockForUpdate()`:                    
- Postgres: takes `FOR UPDATE` row lock, blocks concurrent readers.           
- MySQL InnoDB: same.                                                         
  additional guarantee.                                                         
                                                                                
## Why it's a problem                                                         
- Defeats the purpose of choosing `database` over `redis` driver.             
- Leaks a cache dependency into a driver that should be self-contained.
- Constructor returns `CacheFacadeNotSet` if no cache configured, the database driver should not care whether cache exists.                       
                                                                                
## Proposed direction (open to discussion)                                    
1. Remove the cache lock entirely; rely on DB FOR UPDATE.
2. OR make it opt-in via `queue.connections.<name>.use_cache_lock: bool`,  default false.                                                        
3. OR fall back to an in-process mutex when cache is unavailable.            
Compare to Laravel's database queue driver, which uses lockForUpdate() alone.

wafik:
## Problem

The database queue driver hard-required a cache backend: `NewDatabase` returned `CacheFacadeNotSet` whenever `cache == nil`, so a DB-queue-only app could not even construct the driver without configuring a cache. On top of that, `Pop()` took a cache lock around a transaction that already serializes reservation with `SELECT ... FOR UPDATE` — two cache round-trips per poll for no correctness gain.

Fixes goravel/goravel#960.

## Approach

Followed option 2 from the issue: the cache lock becomes **opt-in** via `queue.connections.<name>.use_cache_lock` (default `false`).

- The constructor no longer rejects a nil cache; it only errors with `CacheFacadeNotSet` when the user explicitly enables the cache lock without a cache configured.
- `Pop()` wraps the cache lock in `if r.useCacheLock`, relying on `FOR UPDATE` alone otherwise — the same contract as Laravel's database queue driver.
- `use_cache_lock: false` documented in the generated queue config stub.

## Testing

- `go build ./...`
- `go test ./queue/... -count=1` — 191 passed
- New cases: nil cache constructs OK, opt-in lock works, opt-in without cache errors; `Pop` default path takes no cache lock, with-lock path unchanged.

## Notes

- Silent default change: apps that relied on the cache lock as a belt-and-braces guard now get `FOR UPDATE`-only semantics by default. The issue argues this is correct (Laravel parity); worth a changelog mention.
- I left `Cache` in the Queue binding's `Dependencies` (`contracts/binding/binding.go`) untouched — removing it changes `artisan package:install` prompts and boot ordering; happy to drop it in a follow-up if maintainers want full consistency.

### DIFF
--- a/queue/driver_creator_test.go
+++ b/queue/driver_creator_test.go
@@ -76,6 +76,7 @@ func (s *DriverCreatorTestSuite) TestCreate() {
 			setup: func() {
 				s.mockConfig.EXPECT().Driver("database").Return(contractsqueue.DriverDatabase).Once()
 				s.mockConfig.EXPECT().GetString("queue.connections.database.connection").Return("mysql").Once()
+				s.mockConfig.EXPECT().GetBool("queue.connections.database.use_cache_lock").Return(false).Once()
 				s.mockConfig.EXPECT().GetString("queue.connections.database.table", "jobs").Return("jobs").Once()
 				s.mockConfig.EXPECT().GetInt("queue.connections.database.retry_after", 60).Return(60).Once()
 				s.mockDB.EXPECT().Connection("mysql").Return(s.mockDB).Once()
--- a/queue/driver_database.go
+++ b/queue/driver_database.go
@@ -24,8 +24,9 @@ type Database struct {
 	jobStorer contractsqueue.JobStorer
 	json      contractsfoundation.Json
 
-	jobsTable  string
-	retryAfter int
+	jobsTable    string
+	retryAfter   int
+	useCacheLock bool
 }
 
 func NewDatabase(
@@ -35,13 +36,18 @@ func NewDatabase(
 	jobStorer contractsqueue.JobStorer,
 	json contractsfoundation.Json,
 	connection string) (*Database, error) {
-	if cache == nil {
-		return nil, errors.CacheFacadeNotSet.SetModule(errors.ModuleQueue)
-	}
-
 	dbConnection := config.GetString(fmt.Sprintf("queue.connections.%s.connection", connection))
 	if dbConnection == "" {
 		return nil, errors.QueueInvalidDatabaseConnection.Args(connection)
+	}
+
+	// The cache lock is opt-in: the SELECT ... FOR UPDATE inside the
+	// transaction already serializes job reservation on the database
+	// (same contract as Laravel's database queue driver), so a cache
+	// backend is not required unless explicitly enabled.
+	useCacheLock := config.GetBool(fmt.Sprintf("queue.connections.%s.use_cache_lock", connection))
+	if useCacheLock && cache == nil {
+		return nil, errors.CacheFacadeNotSet.SetModule(errors.ModuleQueue)
 	}
 
 	return &Database{
@@ -50,8 +56,9 @@ func NewDatabase(
 		jobStorer: jobStorer,
 		json:      json,
 
-		jobsTable:  config.GetString(fmt.Sprintf("queue.connections.%s.table", connection), "jobs"),
-		retryAfter: config.GetInt(fmt.Sprintf("queue.connections.%s.retry_after", connection), 60),
+		jobsTable:    config.GetString(fmt.Sprintf("queue.connections.%s.table", connection), "jobs"),
+		retryAfter:   config.GetInt(fmt.Sprintf("queue.connections.%s.retry_after", connection), 60),
+		useCacheLock: useCacheLock,
 	}, nil
 }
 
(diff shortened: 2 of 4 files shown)


wafik:
Heads-up on the failing `unit/integration` jobs: they come from the example-app workflow, not this change — `goravel/goravel`'s `SendShipmentNotification.Handle(...any)` no longer satisfies the updated `event.Listener` contract (`Handle(string, ...any)`) on framework master, so the example app fails to build for any PR (#1548/#1549 fail identically). Framework unit/lint/mockery/codecov are all green here.

goravel-coder:
# 🤖 Automated review

## Summary
The PR fixes #960 by letting the DB queue driver construct without a cache backend. The agreed direction is now simpler: drop the cache lock entirely (no `use_cache_lock` config) and rely on `SELECT ... FOR UPDATE` alone; `SKIP LOCKED` is deferred to a separate follow-up issue.

## Verdict
- Must Fix: 0 · Should Fix: 3 · Nits: 2

## Findings

### Should Fix
- [ ] `queue/driver_database.go:48` **Remove the `use_cache_lock` config and the whole cache-lock path** — the PR currently adds an opt-in `use_cache_lock` field (`:29`), reads `queue.connections.<name>.use_cache_lock` (`:48`), keeps a nil-cache guard (`:49-51`), and wraps `Pop()`'s lock in `if r.useCacheLock` (`:72-80`). The decision is to remove all of it. Suggestion: `NewDatabase` accepts `nil` cache unconditionally (no error), delete the `useCacheLock` field, and have `Pop()` go straight to the `Transaction`/`LockForUpdate()` path with no lock.
- [ ] `contracts/binding/binding.go:283` **Drop the now-dead `Cache` dependency from the queue stack** — once the lock is gone, `cache` is threaded `queue/service_provider.go:39 → application.go → driver_creator.go:14,21 → worker.go:45 → pending_job.go:23-25 → NewDatabase` but never dereferenced. Leaving `Cache` in the Queue binding's `Dependencies` makes `package:install queue` still prompt for a cache facade the driver no longer needs. Suggestion: remove it in this change and relax the unused `cache` parameters.
- [ ] `queue/driver_database_test.go:148` **Revert the cache-lock test cases and mock expectations** — remove the `GetBool("...use_cache_lock")` expectation in `driver_creator_test.go:79`, drop `wantUseCacheLock`/`useCacheLock` fields and the `"happy path with cache lock"` / `"failed to acquire cache lock"` cases, and the `s.database.useCacheLock = test.useCacheLock` assignment (`:315`). Suggestion: keep a "no cache lock taken" assertion (e.g. `s.mockCache.AssertNotCalled(s.T(), "Lock", ...)`) and cover the no-job path under the plain `FOR UPDATE` flow.

### Nits
- [ ] `queue/setup/stubs.go:34` **Remove the `use_cache_lock` stub entry and its comment** — no config key remains in the revised design. Suggestion: restore the `database` connection block to `driver`/`connection`/`queue`/`concurrent` only.
- [ ] `queue/driver_database.go:44` **Update the comment to describe the final mechanism** — it currently mentions the opt-in lock; reword to "serialization is the DB row lock (`FOR UPDATE`) inside the transaction; no cache lock is required."

## Note to author — `SKIP LOCKED` follow-up (out of scope for this PR)

Plain `FOR UPDATE` blocks sibling workers on the head row (and, since `First` adds no `LIMIT 1`, can lock all currently-available/expired rows), and SQLite ignores `FOR UPDATE` entirely. Laravel addresses both by upgrading the row lock per engine (`FOR UPDATE SKIP LOCKED` on MySQL 8.0.1+/MariaDB 10.6+/Postgres 9.5+, `WITH (ROWLOCK, UPDLOCK, READPAST)` on SQL Server, plain `FOR UPDATE` fallback). This will be tracked in a separate issue — recommend linking it from this PR and, until then, noting in the changelog that SQLite users should keep `concurrent: 1`.

## Automated Checks
- [ ] `gofmt -d .` — clean (run by review-standard against the worktree)
- [ ] `go vet ./queue/...` — clean (run by review-standard against the worktree)

PR #1554, opened 2026-09-07
perf: [#998] compile validation patterns once instead of on every field

### LINKED ISSUE #998: Validation compiles the same regular expressions on every field of every request
### Before feedback (在反馈之前)

### Describe Feature (描述功能)

Three places in `validation` call `regexp.Compile` on every invocation, with
patterns that never change:

* `ruleRegex` (`validation/rules.go:1099`), the `regex:` rule
* `ruleNotRegex` (`:1112`), the `not_regex:` rule
* `expandWildcardFields` (`validation/utils.go:469`), which builds a pattern for
  every field name containing `*`

The patterns come from rule definitions written in application code, so the same
handful is rebuilt on every field of every request. Compiling is what a regular
expression costs; matching against an already compiled one is cheap.

The rest of the package already keeps its patterns in package level variables:
`macRegex`, `uuidRegex`, `ulidRegex`, `hexColorRegex`, `htmlTagRegex`. This is
about bringing the last three call sites in line with that.

### Numbers

Measured through the public API, `Make` plus `Fails`, on go1.27.0 darwin/arm64:

| | ns/op | B/op | allocs/op |
|---|---|---|---|
| `required\|string` on two fields, no regex rule | 1406 | 2080 | 29 |
| the same plus one `regex:` rule | 3532 | 6896 | 90 |

One `regex:` rule costs more than the entire rest of the validation. With the
patterns compiled once it drops to 1604 ns / 2154 B / 33 allocs, so the marginal
cost of that rule goes from **+2101 ns to +190 ns**, about 11x.

The wildcard path moves too. Using the benchmark already in the repo:

| | ns/op | B/op | allocs/op |
|---|---|---|---|
| `BenchmarkExpandWildcardFields` before | 20386 | 27520 | 227 |
| after | 12323 | 10901 | 20 |

`BenchmarkValidation_Wildcards`, also already in the repo, goes from
69022 ns / 55953 B to 60339 ns / 39332 B.

Validation that uses no regex rule is unchanged: 1406 ns before, 1390 ns after,
same allocations.

### Approach

A `sync.Map` keyed by the pattern string, checked before compiling.

Failed compiles are deliberately not cached. Valid patterns are bounded by the
rules a program declares, but invalid ones are not, so caching failures would
let a rule set built from untrusted input grow the map. Not caching them also
leaves the current behaviour of an invalid pattern exactly as it is.

For what it is worth, this is what PHP does for Laravel without anyone asking:
`ext/pcre` keeps a cache of compiled patterns, bounded at 4096 entries, so
`preg_match` with a repeated pattern does not recompile. Go has no equivalent,
which is why the cost shows up here and not in the original.

### Status

I have this working locally, with unit tests for the cache and a
`BenchmarkValidation_RegexRule` alongside the existing benchmarks. The
validation suite, the full framework suite and `-race` all pass.

Happy to send the PR if you are interested.

darakanoit:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/998

`ruleRegex` (`validation/rules.go:1099`), `ruleNotRegex` (`:1112`) and
`expandWildcardFields` (`validation/utils.go:469`) called `regexp.Compile` on
every invocation. The patterns come from rule definitions written in application
code, so the same handful was rebuilt on every field of every request.
Compiling is what a regular expression costs, matching an already compiled one
is cheap.

The rest of the package already keeps its patterns in package level variables
(`macRegex`, `uuidRegex`, `ulidRegex`, `hexColorRegex`, `htmlTagRegex`), so this
brings the last three call sites in line with that.

### What changed

`compilePattern` looks the pattern up in a `sync.Map` before compiling it and
stores it with `LoadOrStore`, so concurrent callers share one compiled value.
`trackDistinct` (`validation/engine.go`), which went through `regexp.MatchString`
on every `distinct` check, uses it as well.

Failed compiles are not cached. Valid patterns are bounded by the rules a
program declares, invalid ones are not, so caching failures would let a rule set
built from untrusted input grow the map. It also leaves the behaviour of an
invalid pattern exactly as it was.

For the same reason the cache has a soft cap of 1024 entries. Past it, new
patterns are compiled on every call, which is the behaviour before this PR, so
rules built from untrusted input cannot grow the map without bound either.

### Numbers

```
go test ./validation/ -bench='Validation_RegexRule|ExpandWildcardFields|Validation_Wildcards' -benchmem
```

`BenchmarkValidation_RegexRule` and `BenchmarkValidation_WildcardsDistinct` are
added by this PR, the other two already existed.

before:
```
BenchmarkValidation_RegexRule-8      339045    3532 ns/op    6896 B/op    90 allocs/op
BenchmarkExpandWildcardFields-8       60337   20386 ns/op   27520 B/op   227 allocs/op
BenchmarkValidation_Wildcards-8       17259   69022 ns/op   55953 B/op   775 allocs/op
BenchmarkValidation_WildcardsDistinct-8  11895  100517 ns/op  147404 B/op  2083 allocs/op
```

after:
```
BenchmarkValidation_RegexRule-8      747267    1604 ns/op    2154 B/op    33 allocs/op
BenchmarkExpandWildcardFields-8       96663   12323 ns/op   10901 B/op    20 allocs/op
BenchmarkValidation_Wildcards-8       19732   60339 ns/op   39332 B/op   568 allocs/op
BenchmarkValidation_WildcardsDistinct-8  27159   44134 ns/op   28687 B/op   489 allocs/op
```

The interesting number is the marginal one. Two fields with `required|string`
and no regex rule cost 1406 ns. Adding one `regex:` rule used to cost another
2101 ns, more than the rest of the validation put together. It now costs 190 ns.

Validation that uses no regex rule is unchanged: 1406 ns before, 1390 ns after,
same allocations.

### Scope

* `validation/utils.go`: `compilePattern` and the wildcard call site
* `validation/rules.go`: two call sites
* `validation/engine.go`: `trackDistinct`
* `validation/utils_test.go`: tests for the cache
* `validation/benchmark_test.go`: `BenchmarkValidation_RegexRule`, `BenchmarkValidation_WildcardsDistinct`

No public API changed.

## ✅ Checks

`TestCompilePattern` covers that a pattern compiles and matches, that the same
pattern returns the identical compiled value rather than a new one, that concurrent
callers get the same value, that an invalid pattern still errors and does not
end up in the map, and that patterns past the cap still compile but are not
stored.

### DIFF
--- a/validation/benchmark_test.go
+++ b/validation/benchmark_test.go
@@ -10,6 +10,28 @@ import (
 
 // --- End-to-end validation benchmarks ---
 
+func BenchmarkValidation_RegexRule(b *testing.B) {
+	v := NewValidation()
+	ctx := context.Background()
+	data := map[string]any{
+		"name": "John",
+		"code": "AB-1234",
+	}
+	rules := map[string]any{
+		"name": "required|string",
+		"code": `required|regex:^[A-Z]{2}-\d{4}$`,
+	}
+
+	b.ReportAllocs()
+	for b.Loop() {
+		validator, err := v.Make(ctx, data, rules)
+		if err != nil {
+			b.Fatal(err)
+		}
+		_ = validator.Fails()
+	}
+}
+
 func BenchmarkValidation_Simple(b *testing.B) {
 	v := NewValidation()
 	ctx := context.Background()
--- a/validation/rules.go
+++ b/validation/rules.go
@@ -1096,7 +1096,7 @@ func ruleRegex(ctx *RuleContext) bool {
 		return false
 	}
 	pattern := ctx.Parameters[0]
-	re, err := regexp.Compile(pattern)
+	re, err := compilePattern(pattern)
 	if err != nil {
 		return false
 	}
@@ -1109,7 +1109,7 @@ func ruleNotRegex(ctx *RuleContext) bool {
 		return false
 	}
 	pattern := ctx.Parameters[0]
-	re, err := regexp.Compile(pattern)
+	re, err := compilePattern(pattern)
 	if err != nil {
 		return false
 	}
--- a/validation/utils.go
+++ b/validation/utils.go
@@ -8,6 +8,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 	"unicode"
 	"unicode/utf8"
@@ -15,6 +16,31 @@ import (
 	"github.com/gabriel-vasile/mimetype"
 	"github.com/spf13/cast"
 )
+
+// compiledPatterns caches compiled patterns. The patterns come from rule
+// definitions, which are written in application code, so the set of distinct
+// patterns a program uses is bounded by the rules it declares. Failed compiles
+// are deliberately not cached, so an invalid pattern cannot grow the map.
+var compiledPatterns sync.Map
+
+// compilePattern compiles a pattern once and reuses it afterwards. Compiling is
+// what a regular expression costs, matching against an already compiled one is
+// cheap, and the rules that use this compiled the same pattern on every field of
+// every request.
+func compilePattern(pattern string) (*regexp.Regexp, error) {
+	if cached, ok := compiledPatterns.Load(pattern); ok {
+		return cached.(*regexp.Regexp), nil
+	}
+
+	compiled, err := regexp.Compile(pattern)
+	if err != nil {
+		return nil, err
+	}
+
+	compiledPatterns.Store(pattern, compiled)
+
+	return compiled, nil
+}
 
 // isValueEmpty checks if a value is considered "empty" for validation purposes.
 func isValueEmpty(val any) bool {
@@ -466,7 +492,7 @@ func expandWildcardFields[T any](fields map[string]T, dataKeys []string, keepUnm
 
 		pattern := "^" + regexp.QuoteMeta(field) + "$"
 		pattern = strings.ReplaceAll(pattern, `\*`, `[^.]+`)
-		re, err := regexp.Compile(pattern)
+		re, err := compilePattern(pattern)
 		if err != nil {
 			expanded[field] = value
 			continue
--- a/validation/utils_test.go
+++ b/validation/utils_test.go
@@ -11,6 +11,35 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+func TestCompilePattern(t *testing.T) {
+	t.Run("compiles and matches", func(t *testing.T) {
+		re, err := compilePattern(`^[A-Z]{2}-\d{4}$`)
+		assert.NoError(t, err)
+		assert.True(t, re.MatchString("AB-1234"))
+		assert.False(t, re.MatchString("ab-1234"))
+	})
+
+	t.Run("the same pattern is compiled once", func(t *testing.T) {
+		first, err := compilePattern(`^cached-[0-9]+$`)
+		assert.NoError(t, err)
+
+		second, err := compilePattern(`^cached-[0-9]+$`)
+		assert.NoError(t, err)
+		assert.Same(t, first, second)
+	})
+
+	t.Run("an invalid pattern errors and is not cached", func(t *testing.T) {
+		const invalid = `^[a-z`
+
+		re, err := compilePattern(invalid)
+		assert.Error(t, err)
+		assert.Nil(t, re)
+
+		_, cached := compiledPatterns.Load(invalid)
+		assert.False(t, cached)
+	})
+}
+
 func TestIsValueEmpty(t *testing.T) {
 	tests := []struct {
 		name     string


goravel-coder:
# 🤖 Automated review

## Summary
Adds a package-global `sync.Map` cache that memoizes `regexp.Compile` for `ruleRegex`, `ruleNotRegex`, and `expandWildcardFields`, eliminating per-call compilation. Correct and concurrency-safe, but the cache is unbounded and never evicted — a memory-growth risk for apps that build rule patterns dynamically.

## Verdict
- Must Fix: 0 · Should Fix: 3 · Nits: 7

## Findings

### Should Fix
- [ ] `validation/utils.go:24` **Unbounded, never-evicted global cache** — `compiledPatterns` is keyed by application-controlled strings (regex rule params, wildcard field keys) and has no size cap, eviction, or reset, so valid dynamic patterns are retained for the lifetime of the process. Suggestion: precompile at rule-parse time (bounds lifetime to a single `Make()` call), or use a size-capped LRU.
- [ ] `validation/utils.go:26` **Grammar error in the doc comment** — "the rules that use this **compiled** the same pattern" should be "compile". Suggestion: fix the verb.
- [ ] `validation/utils_test.go:22` **Test isolation + no concurrency test** — the `assert.Same` subtest relies on a hand-picked unique literal and silently stops exercising the compile path under `-count=2`; the new `sync.Map` has no concurrent-access test. Suggestion: reset the cache key in the test and add a `-race` goroutine case.

### Nits
- [ ] `validation/utils.go:24` **`sync.Map` is suboptimal here** — read-heavy, write-once access favors a plain `map` + `sync.RWMutex` (simpler, type-safe, no boxing on the hot path). Suggestion: consider the simpler construct.
- [ ] `validation/utils.go:31` **Benign duplicate-compile race** — two goroutines can both miss `Load` and both `Compile`/`Store`. Suggestion: use `LoadOrStore` to return a canonical instance.
- [ ] `validation/utils.go:32` **Bare type assertion** — `cached.(*regexp.Regexp)` is a panic site if any future caller stores a different type. Suggestion: use the comma-ok form.
- [ ] `validation/engine.go:279` **`trackDistinct` still recompiles the wildcard regex** — same `QuoteMeta`→`ReplaceAll` pattern, still routed through uncached `regexp.MatchString` on every distinct check. Suggestion: route through `compilePattern`.
- [ ] `validation/utils.go:20` **Inaccurate cache comment** — "bounded by the rules it declares" also covers `expandWildcardFields`, which synthesizes patterns from field names. Suggestion: tighten wording.
- [ ] `validation/benchmark_test.go:13` **Benchmark instrumentation** — omits `b.ResetTimer()` (inconsistent with siblings) and measures end-to-end `Make()`, not the compile/match in isolation. Suggestion: align with file convention; consider a direct micro-benchmark.
- [ ] `validation/rules.go:1106` **`ruleNotRegex` empty-pattern inconsistency** — rejects only `len == 0`, not `""`, unlike `ruleRegex`; `not_regex:""` always fails the field. Pre-existing. Suggestion: align with `ruleRegex`.

## Automated Checks
- [ ] `gofmt -d validation/` — clean
- [ ] `go vet ./validation/` — clean

darakanoit:
Thanks for the review. Went through every point, changes are in 247a0401.

### Should Fix

**Unbounded cache (`utils.go:24`)**: partially agree, added a soft cap.

Keys come from `regex:` / `not_regex:` parameters and from rule keys containing `*`. Both are written in application code, nothing from the request data ends up as a key, so for a normal program the set is small and fixed. The case you describe (rules assembled from untrusted input) is real though, so `compilePattern` now stops retaining new patterns once `maxCompiledPatterns` (1024) is reached. Past that point it compiles on every call, which is exactly the behaviour before this PR, so the worst case is no regression and no growth. The counter is an `atomic.Int64` checked before `LoadOrStore`, so under contention it can overshoot by a few entries. That is fine for a guard like this and keeps the hot path lock free.

I did not go with the two suggested alternatives:
* Precompiling at rule parse time bounds the lifetime to one `Make()`, but rules are parsed on every `Make()`, i.e. on every request. For the common case of one regex rule on one field that is the same single compile as before, so it removes most of the gain.
* An LRU adds a mutex and list bookkeeping to every lookup to handle a case that should not happen in practice. A hard ceiling with fallback gives the same memory guarantee for much less.

**Doc comment grammar (`utils.go:26`)**: fixed. The sentence was past tense describing the old behaviour, which belongs in the PR description rather than in the code. Rewrote the comment to describe what the function does.

**Test isolation and concurrency (`utils_test.go:22`)**: agree, fixed.
* Every subtest now resets the cache first, so `-count=N` exercises the compile path each time, and the tests also assert the entry counter.
* Added a test with 32 goroutines compiling the same pattern, asserting they all get the identical `*regexp.Regexp` and only one entry is stored.
* Added a test for the cap: fill to the limit, check that the next pattern still compiles and matches but is not stored, and that cached patterns keep being served.
* `go test -race -count=2 ./validation/` passes.

### Nits

**`sync.Map` vs `map` + `RWMutex`**: keeping `sync.Map`. A stable set of keys that is written once and then read from many goroutines is the case its docs name as the intended use. Since Go 1.24 it is backed by a concurrent hash trie, so reads do not contend on a shared lock the way `RWMutex` readers do on the reader counter. On type safety, the map is private to a single function, so the assertion lives in one place.

**Duplicate compile race (`utils.go:31`)**: agree, switched to `LoadOrStore`. Two goroutines can still both compile on a cold miss, but they now return the same canonical instance and the counter only moves when the store actually happened.

**Bare type assertion (`utils.go:32`)**: keeping as is. `compiledPatterns` is unexported and is written in exactly one place, which only stores `*regexp.Regexp`. A comma-ok branch would be dead code, and if the invariant were ever broken a panic is the right outcome rather than a silently ignored cache.

**`trackDistinct` still recompiles (`engine.go:279`)**: good catch, this was the most useful point in the review. `regexp.MatchString` compiles on every call, for every wildcard rule, for every field with `distinct`, which is exactly what #998 is about. Routed it through `compilePattern` and added `BenchmarkValidation_WildcardsDistinct` (20 items, `items.*.sku` with `distinct`):

```
before:  100517 ns/op   147404 B/op   2083 allocs/op
after:    44134 ns/op    28687 B/op    489 allocs/op
```

**Cache comment wording (`utils.go:20`)**: I think the original was accurate, since wildcard patterns are built from rule keys, which are part of the rule definitions and not from data keys. The comment is rewritten anyway as part of the cap change and now says "Patterns come from rule definitions", which covers both sources.

**Benchmark `b.ResetTimer()` (`benchmark_test.go:13`)**: added for consistency with the rest of the file, although with `b.Loop()` it is a no-op since the loop resets the timer on its first call. On measuring end to end: that is intentional. The point of the benchmark is the cost a `regex:` rule adds to a real `Make` + `Fails` call, and `BenchmarkExpandWildcardFields` already covers the isolated path. A micro benchmark of `compilePattern` alone would only measure `sync.Map.Load`.

**`ruleNotRegex` empty pattern (`rules.go:1106`)**: pre-existing and out of scope, but also not an inconsistency in outcome. An empty pattern compiles and matches every string, so `not_regex` with an empty parameter returns `false`, the same result `ruleRegex` returns through its explicit check. Leaving it alone to keep this PR focused on compilation.

hwbrzzl (approved):
Thanks, LGTM

PR #1555, opened 2026-09-07
perf: [#999] stop rebuilding the row decoder and buffer the cursor channel

### LINKED ISSUE #999: Cursor and Chunk spend about 20us per row before the application sees it
### Before feedback (在反馈之前)

### Describe Feature (描述功能)

The cursor path repeats work on **every row** that could be done once.

### Where

Three entry points, all reaching the same code:

| API | what it hits |
|---|---|
| `db.Query().Cursor()` | the decoder and an unbuffered channel |
| `db.Query().Chunk()` | the same, `Chunk` calls `Cursor` internally (`database/db/query.go:56`) |
| `orm.Query().Cursor()` | the same, the gorm `Row.Scan` delegates to `db.NewRow(...).Scan()` (`database/gorm/row.go:18`) |

Plain `Get`, `First` and `Find` do not come through here, they go straight to
`sqlx.SelectContext` and to gorm. So the path is narrow, but it is the one used
to walk large result sets.

### What costs

**Per row.** `Row.Scan` (`database/db/row.go:29`) composes the six decode hooks
and builds a `mapstructure` decoder for every row. mapstructure then calls
`MatchName` for **every column and field pair**, and each of those calls runs
`Studly` on the column name and `Snake` on the field name, constructing an
`x/text` caser both times. Column names come from the schema and field names
from the code, so the set is small and fixed for the life of the process.

The `reflect.TypeOf` values the hooks compare against are also rebuilt on every
hook call.

**Per row again.** Both `Cursor` implementations hand rows over an unbuffered
channel, which costs a goroutine handoff each time.

### Numbers

Measured with a driver that returns synthetic rows immediately, so only
framework work is timed, no I/O. go1.27.0, darwin/arm64.

| | ns/op | B/op | allocs/op |
|---|---|---|---|
| `Cursor` over 100 rows, today | 2138637 | 2256388 | 32065 |
| after | **435314** | **409895** | **6912** |
| `Cursor` over 1000 rows, today | 20847660 | 22131739 | 320277 |
| after | **4261419** | **3862573** | **68886** |

About 20.8us per row today, 4.26us after. A thousand rows is 21ms of framework
overhead, 16ms of which goes away.

### Suggested approach

Build the hooks once, memoise the two name conversions, check the condition in
`MatchName` that needs no conversion first (all three are joined with `||`, so
the order does not change the result), lift the `reflect.TypeOf` values to
package level, and give the cursor channel a buffer.

### Relation to #714

That issue asks for `chunk` and `chunkById` in the ORM module. If those land,
noticeably more traffic goes through this path, so fixing it first seems worth
doing.

### Status

Working locally, 3 files, +76/-19. The `database` suites, `-race` on `db` and
`gorm`, the full framework suite, `gofmt` and `vet` all pass. Happy to send the
PR.

### Reproducing

No database needed. Drop this in `database/db/` and run:

```
go test ./database/db/ -bench='Cursor100$|Cursor1000$' -benchmem -run=^$
```

```go
package db

import (
	"context"
	"database/sql"
	"database/sql/driver"
	"fmt"
	"io"
	"testing"

	"github.com/jmoiron/sqlx"
	gormlogger "gorm.io/gorm/logger"

	contractslogger "github.com/goravel/framework/contracts/database/logger"
	"github.com/goravel/framework/contracts/log"
	mocksdriver "github.com/goravel/framework/mocks/database/driver"
)

// A driver that hands back N synthetic rows, so Cursor runs through the real
// code path without needing a database.

type fakeDriver struct{ rows int }

func (d fakeDriver) Open(string) (driver.Conn, error) { return fakeConn{rows: d.rows}, nil }

type fakeConn struct{ rows int }

func (c fakeConn) Prepare(string) (driver.Stmt, error) { return fakeStmt{rows: c.rows}, nil }
func (c fakeConn) Close() error                        { return nil }
func (c fakeConn) Begin() (driver.Tx, error)           { return nil, io.EOF }

type fakeStmt struct{ rows int }

func (s fakeStmt) Close() error                               { return nil }
func (s fakeStmt) NumInput() int                              { return -1 }
func (s fakeStmt) Exec([]driver.Value) (driver.Result, error) { return driver.RowsAffected(0), nil }
func (s fakeStmt) Query([]driver.Value) (driver.Rows, error)  { return &fakeRows{total: s.rows}, nil }

type fakeRows struct {
	total int
	sent  int
}

func (r *fakeRows) Columns() []string {
	return []string{"id", "name", "email", "avatar", "created_at", "updated_at"}
}
func (r *fakeRows) Close() error { return nil }
func (r *fakeRows) Next(dest []driver.Value) error {
	if r.sent >= r.total {
		return io.EOF
	}
	r.sent++
	dest[0] = int64(r.sent)
	dest[1] = "bob"
	dest[2] = "b@x.io"
	dest[3] = "a.png"
	dest[4] = "2024-01-01"
	dest[5] = "2024-01-01"
	return nil
}

type benchBuilder struct{ *sqlx.DB }

func (b benchBuilder) Explain(sql string, args ...any) string { return sql }

type nopLogger struct{}

func (nopLogger) Trace(context.Context, *carbon.Carbon, string, int64, error) {}
func (nopLogger) Infof(context.Context, string, ...any)                       {}
func (nopLogger) Warningf(context.Context, string, ...any)                    {}
func (nopLogger) Errorf(context.Context, string, ...any)                      {}
func (nopLogger) Debugf(context.Context, string, ...any)                      {}
func (n nopLogger) Panicf(context.Context, string, ...any)                    {}
func (n nopLogger) Level(contractslogger.Level) contractslogger.Logger        { return n }
func (n nopLogger) Log() log.Log                                              { return nil }
func (n nopLogger) ToGorm() gormlogger.Interface                              { return nil }

type benchUser struct {
	ID        uint   `db:"id"`
	Name      string `db:"name"`
	Email     string `db:"email"`
	Avatar    string `db:"avatar"`
	CreatedAt string `db:"created_at"`
	UpdatedAt string `db:"updated_at"`
}

var registered int

func newBenchQuery(b *testing.B, rows int) *Query {
	registered++
	name := fmt.Sprintf("fake%d", registered)
	sql.Register(name, fakeDriver{rows: rows})
	raw, err := sqlx.Open(name, "")
	if err != nil {
		b.Fatal(err)
	}

	grammar := mocksdriver.NewGrammar(b)
	grammar.EXPECT().CompilePlaceholderFormat().Return(nil).Maybe()

	return NewQuery(context.Background(), benchBuilder{raw}, benchBuilder{raw}, grammar, nopLogger{}, "users", nil)
}

func benchCursor(b *testing.B, rows int) {
	q := newBenchQuery(b, rows)
	b.ReportAllocs()
	for b.Loop() {
		var u benchUser
		n := 0
		for row := range q.Cursor() {
			if err := row.Err(); err != nil {
				b.Fatal(err)
			}
			if err := row.Scan(&u); err != nil {
				b.Fatal(err)
			}
			n++
		}
		if n != rows {
			b.Fatalf("got %d rows, want %d", n, rows)
		}
	}
}

func BenchmarkCursor100(b *testing.B)  { benchCursor(b, 100) }
func BenchmarkCursor1000(b *testing.B) { benchCursor(b, 1000) }
```

darakanoit:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/999

The cursor path repeated per row what can be done once.

`Row.Scan` (`database/db/row.go:29`) composed the six decode hooks and built a
`mapstructure` decoder for every row. mapstructure then called `MatchName` for
every column and field pair, and each of those calls ran `Studly` on the column
name and `Snake` on the field name, constructing an `x/text` caser both times.
Column names come from the schema and field names from the code, so the same
small set of conversions was redone on every row. The `reflect.TypeOf` values
the hooks compare against were rebuilt on every hook call too.

Both `Cursor` implementations handed rows over an unbuffered channel, which
costs a goroutine handoff per row.

This path is reached by `db.Query().Cursor()`, by `db.Query().Chunk()`, which
calls `Cursor` internally, and by `orm.Query().Cursor()`, whose `Row.Scan`
delegates to the `db` one (`database/gorm/row.go:18`).

### What changed

The hooks are built once, the two name conversions are memoised, and the
`reflect.TypeOf` values moved to package level. Both cursor channels now keep
64 rows in flight.

`matchName` keeps all three original conditions and only reorders them so the
one that needs no conversion is checked first. They are joined with `||`, so the
result is unchanged.

One small cleanup came with it: `ToScannerHookFunc` compared against both
`reflect.TypeOf([]byte(nil))` and `reflect.TypeOf([]uint8(nil))`. `byte` is an
alias for `uint8`, so those are the same `reflect.Type` and the check was
duplicated. One remains.

### Numbers

Measured with a driver that returns synthetic rows immediately, so only
framework work is timed. The harness is in the issue, no database needed.

```
go test ./database/db/ -bench='Cursor100$|Cursor1000$' -benchmem -run=^$
```

before:
```
BenchmarkCursor100-8     544    2138637 ns/op    2256388 B/op    32065 allocs/op
BenchmarkCursor1000-8     56   20847660 ns/op   22131739 B/op   320277 allocs/op
```

after:
```
BenchmarkCursor100-8    2739     435314 ns/op     409895 B/op     6912 allocs/op
BenchmarkCursor1000-8    282    4261419 ns/op    3862573 B/op    68886 allocs/op
```

About 20.8us per row before, 4.26us after. A thousand rows carried 21ms of
framework overhead, 16ms of which goes away.

### Scope

* `database/db/row.go`: hooks built once, conversions memoised, types lifted
* `database/db/query.go`: buffered cursor channel
* `database/gorm/query.go`: buffered cursor channel

No public API changed, no behaviour changed.

## ✅ Checks

No new tests. This changes how existing work is done, not what it produces, so
the existing `database` suites cover it, and they pass along with `-race` on
`db` and `gorm`.

The benchmark harness from the issue is not included here to keep the diff
small. Happy to add it as a permanent benchmark if you would like the cursor
path measurable without a database.

### DIFF
--- a/database/db/query.go
+++ b/database/db/query.go
@@ -49,6 +49,9 @@ func NewQuery(ctx context.Context, readBuilder db.CommonBuilder, writeBuilder db
 	}
 }
 
+// cursorBufferSize is the number of rows Cursor keeps in flight.
+const cursorBufferSize = 64
+
 func (r *Query) Chunk(size uint64, callback func(rows []db.Row) error) error {
 	offset := uint64(0)
 
@@ -116,7 +119,9 @@ func (r *Query) CrossJoin(query string, args ...any) db.Query {
 }
 
 func (r *Query) Cursor() chan db.Row {
-	ch := make(chan db.Row)
+	// Buffered so the producing goroutine can stay ahead of the consumer. An
+	// unbuffered channel costs a goroutine handoff on every single row.
+	ch := make(chan db.Row, cursorBufferSize)
 	go func() {
 		var (
 			args  []any
--- a/database/db/row.go
+++ b/database/db/row.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"reflect"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/go-viper/mapstructure/v2"
@@ -26,20 +27,61 @@ func (r *Row) Err() error {
 	return r.err
 }
 
+// decodeHooks is built once. Composing it per row rebuilt the same six hooks on
+// every row of every result set.
+var decodeHooks = mapstructure.ComposeDecodeHookFunc(
+	ToStringHookFunc(), ToTimeHookFunc(), ToDeletedAtHookFunc(), ToScannerHookFunc(), ToSliceHookFunc(), ToMapHookFunc(),
+)
+
+// mapstructure calls MatchName for every column and field pair, so the case
+// conversions ran O(columns x fields) times per row. Column names and struct
+// field names come from the schema and from the code, so the set is small and
+// fixed for the life of the process.
+var (
+	studlyNames sync.Map
+	snakeNames  sync.Map
+)
+
+func studlyName(name string) string {
+	if cached, ok := studlyNames.Load(name); ok {
+		return cached.(string)
+	}
+
+	converted := str.Of(name).Studly().String()
+	studlyNames.Store(name, converted)
+
+	return converted
+}
+
+func snakeName(name string) string {
+	if cached, ok := snakeNames.Load(name); ok {
+		return cached.(string)
+	}
+
+	converted := str.Of(name).Snake().String()
+	snakeNames.Store(name, converted)
+
+	return converted
+}
+
+// matchName keeps the three original conditions, with the one that needs no
+// conversion first.
+func matchName(mapKey, fieldName string) bool {
+	return strings.EqualFold(mapKey, fieldName) ||
+		studlyName(mapKey) == fieldName ||
+		mapKey == snakeName(fieldName)
+}
+
 func (r *Row) Scan(value any) error {
 	if r.err != nil {
 		return r.err
 	}
 
 	msConfig := &mapstructure.DecoderConfig{
-		DecodeHook: mapstructure.ComposeDecodeHookFunc(
-			ToStringHookFunc(), ToTimeHookFunc(), ToDeletedAtHookFunc(), ToScannerHookFunc(), ToSliceHookFunc(), ToMapHookFunc(),
-		),
-		Squash: true,
-		Result: value,
-		MatchName: func(mapKey, fieldName string) bool {
-			return str.Of(mapKey).Studly().String() == fieldName || mapKey == str.Of(fieldName).Snake().String() || strings.EqualFold(mapKey, fieldName)
-		},
+		DecodeHook: decodeHooks,
+		Squash:     true,
+		Result:     value,
+		MatchName:  matchName,
 	}
 
 	decoder, err := mapstructure.NewDecoder(msConfig)
@@ -50,11 +92,19 @@ func (r *Row) Scan(value any) error {
 	return decoder.Decode(r.row)
 }
 
+var (
+	stringType    = reflect.TypeOf("")
+	timeType      = reflect.TypeOf(time.Time{})
+	deletedAtType = reflect.TypeOf(gorm.DeletedAt{})
+	byteSliceType = reflect.TypeOf([]byte(nil))
+	scannerType   = reflect.TypeOf((*interface{ Scan(any) error })(nil)).Elem()
+)
+
 // ToStringHookFunc is a hook function that converts []uint8 to string.
 // Mysql returns []uint8 for String type when scanning the rows.
 func ToStringHookFunc() mapstructure.DecodeHookFunc {
 	return func(f reflect.Type, t reflect.Type, data any) (any, error) {
-		if t != reflect.TypeOf("") {
+		if t != stringType {
 			return data, nil
 		}
 
@@ -69,7 +119,7 @@ func ToStringHookFunc() mapstructure.DecodeHookFunc {
 
 func ToTimeHookFunc() mapstructure.DecodeHookFunc {
 	return func(f reflect.Type, t reflect.Type, data any) (any, error) {
-		if t != reflect.TypeOf(time.Time{}) {
+		if t != timeType {
 			return data, nil
 		}
 
@@ -88,11 +138,11 @@ func ToTimeHookFunc() mapstructure.DecodeHookFunc {
 
 func ToDeletedAtHookFunc() mapstructure.DecodeHookFunc {
 	return func(f reflect.Type, t reflect.Type, data any) (any, error) {
-		if t != reflect.TypeOf(gorm.DeletedAt{}) {
+		if t != deletedAtType {
 			return data, nil
 		}
 
-		if f == reflect.TypeOf(time.Time{}) {
+		if f == timeType {
 			return gorm.DeletedAt{Time: data.(time.Time), Valid: true}, nil
 		}
 
@@ -109,7 +159,7 @@ func ToDeletedAtHookFunc() mapstructure.DecodeHookFunc {
 func ToScannerHookFunc() mapstructure.DecodeHookFunc {
 	return func(f reflect.Type, t reflect.Type, data any) (any, error) {
 		// Skip types that are handled by other specific hooks
-		if t == reflect.TypeOf(time.Time{}) || t == reflect.TypeOf(gorm.DeletedAt{}) {
+		if t == timeType || t == deletedAtType {
 			return data, nil
 		}
 
@@ -119,12 +169,9 @@ func ToScannerHookFunc() mapstructure.DecodeHookFunc {
 		}
 
 		// Only process database types (string, []byte, []uint8, time.Time)
-		if f.Kind() != reflect.String && f != reflect.TypeOf([]byte(nil)) && f != reflect.TypeOf([]uint8(nil)) && f != reflect.TypeOf(time.Time{}) {
+		if f.Kind() != reflect.String && f != byteSliceType && f != timeType {
 			return data, nil
 		}
-
-		// Check if the target type implements a Scan method
-		scannerType := reflect.TypeOf((*interface{ Scan(any) error })(nil)).Elem()
 
 		// Create a pointer to the target type to check for Scan method
 		targetPtr := reflect.PointerTo(t)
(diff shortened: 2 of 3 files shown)


goravel-coder:
🤖 Automated review — please double-check

# Code Review: PR #1555 — `perf: [#999] stop rebuilding the row decoder and buffer the cursor channel`

**Files:** `database/db/query.go`, `database/db/row.go`, `database/gorm/query.go` · **Language:** Go · **Base:** `master`

Build and unit tests pass. The core refactor is behavior-preserving (verified: `[]uint8`→`[]byte` collapse, `matchName` reordering, `reflect.Type` hoisting, shared `decodeHooks`). **No Must Fix issues.**

## Must Fix

None.

## Should Fix

**1. Unbounded `studlyNames`/`snakeNames` caches — `database/db/row.go:40-65`**
The comment claims the key set is "small and fixed for the life of the process," but `matchName` is keyed on `mapKey`, the raw column name from `sqlx.Rows.MapScan`/`gorm.ScanRows`. Column names can be dynamic (SQL aliases, computed/pivot columns, JSON-extraction aliases). For a framework core library, these `sync.Map` caches grow without bound and are never evicted. Either bound/evict the cache, or correct the comment to acknowledge the assumption is "bounded in practice" rather than "fixed."

**2. Buffered channel widens the goroutine + DB-connection leak on early consumer exit — `database/db/query.go:121-172`, `database/gorm/query.go:165-204`**
The producer sends with a plain `ch <-` and has no `select` on `r.ctx.Done()`. If a caller of the *public* `Cursor()` breaks out of `for row := range cursor` before draining, the producer blocks once the 64-slot buffer fills, and the deferred `rows.Close()`/`close(ch)` never run — leaking a goroutine and a held connection. Pre-existing (unbuffered behaved the same), but this PR touches exactly this code and the new comment ("so the producing goroutine can stay ahead") is easy to misread as addressing it. Consider a `select { case ch <- row: case <-r.ctx.Done(): return }`, or at minimum document the "caller must fully consume" contract. All in-repo consumers drain fully (`Chunk` is safe), so this is an API-contract issue.

**3. PR title overstates the win — the decoder is still rebuilt per `Scan` — `database/db/row.go:87`**
`mapstructure.NewDecoder(msConfig)` + `decoder.Decode(...)` still run per row, and mapstructure re-reflects the entire destination struct each call (no global struct-info cache). Hoisting `decodeHooks` is a real but *small* win; the dominant per-row reflection cost is untouched. Either retitle/adjust the PR description, or go further (cache `*mapstructure.Decoder` per destination `reflect.Type` — note `Decoder` is not concurrency-safe, so use a mutex/`sync.Pool`).

**4. No unit-level coverage for the changed behavior — `database/db/query_test.go`, `database/gorm/query_test.go`, `database/db/row_test.go`**
- The buffered `Cursor()` has **no** unit test in either package; only live-DB integration tests in `tests/` (separate `test_external.yml` workflow) exercise it.
- The newly-shared `decodeHooks` and `sync.Map` caches have no concurrent-`Scan` test, and CI runs plain `go test` without `-race` — nothing guards against a future regression to a non-safe cache.
- No test for early-consumer-exit (leak path in #2).
- The new `matchName`/`studlyName`/`snakeName` helpers are only exercised indirectly; a small table-driven test + cache-hit assertion would lock in intent.

## Nits

- **Non-atomic memoization + unguarded assertion — `database/db/row.go:46-47, 57-58`.** Two goroutines can both `Load`-miss and `Store` (benign for a pure function), but `LoadOrStore` is cleaner; `cached.(string)` is an unguarded single-value assertion (safe today, fragile against future misuse).
- **`sync.Map` is heavier than needed.** A plain `map[string]string` + `sync.RWMutex` (or pre-warm) is simpler for a tiny, process-stable key set.
- **Var declaration ordering — `database/db/row.go:95-101`.** The `reflect.Type` block sits below `decodeHooks` and the cache helpers, with `Scan` interleaved; group the package-level `var` blocks together near the top. (Functionally safe — closures read the vars lazily at `Scan` time.)
- **Missing doc comments on `studlyName`/`snakeName` — `database/db/row.go:45, 56`.** Optional one-liners.
- **Generic package-scope names** (`stringType`, `timeType`, etc.) are collision-prone in the large `db` package; consider a shared prefix or tighter grouping.
- **Off-by-one in `Cursor` trace count (pre-existing) — `database/db/query.go:128,163`.** `count` starts at `-1` and increments *after* the send, so N rows trace `rowsAffected = N-1`. Unchanged by this PR, but it does mean the trace is not the row count.
- **Latent panic with named types (pre-existing) — `database/db/row.go:128-150`.** `data.(string)`/`data.(float64)`/`data.(int64)` are guarded by `f.Kind()` only, so a named type (`type Status string`) matching the kind would panic the assertion. DB drivers return plain types, so it doesn't fire in the normal path.
- **Comment imprecision — `database/db/query.go:122-123`, `database/gorm/query.go:170-171`.** The real benefit of buffering is *pipelining* (overlap producer fetch with consumer decode), not avoiding a "goroutine handoff"; consider correcting the comment.
- **Stale test comment — `database/db/row_test.go:1417-1418`.** Still references `f != reflect.TypeOf(time.Time{})`, which is now the hoisted `timeType`.

## Verified clean (no action)

- `interface{ Scan(any) error }` at `row.go:100,192` — correct under the project's `any` rule (anonymous interfaces with methods can't be written as `any{...}`).
- No inline errors created (`errors/list.go` convention respected); `gofmt`/`go vet` clean.
- `db` → `gorm.io/gorm` dependency is pre-existing (not introduced here).

**Bottom line:** Solid, correct micro-optimization — the caching and channel buffering are safe and behavior-preserving. Main asks before merge: (1) reconcile the unbounded-cache assumption (#1), (2) address/document the early-exit leak now that buffering is in play (#2), (3) align the PR title with what it actually optimizes or go further (#3), and (4) add tests for the changed paths.

PR #1557, opened 2026-09-07
feat: [#991] support listening for database query events

### LINKED ISSUE #991: ✨ [Feature] Support listening for database query events
### Before feedback (在反馈之前)

### Describe Feature (描述功能)

Add a way to listen for every SQL query executed by `facades.DB()` and `facades.Orm()`, same as Laravel's `DB::listen`:

https://laravel.com/docs/13.x/database#listening-for-query-events

```go
facades.DB().Listen(func(query *db.QueryExecuted) error {
    // query.Connection, query.Sql, query.Bindings, query.RawSql, query.Time
    return nil
})
```

Useful for logging, debugging and monitoring slow queries without touching the log level.

wafik:
## Problem

No way to subscribe to executed queries (Laravel's `DB::listen` / `QueryExecuted` parity). Fixes goravel/goravel#991.

## API

```go
facades.DB().Listen(func(query *db.QueryExecuted) error {
    // query.Connection, query.Sql, query.Bindings, query.RawSql, query.Time, query.Error
    return nil
})
```

`db.QueryExecuted` lives in `contracts/database/db`; `Listen` is added to the `DB` interface (mock regenerated) plus a package-level `db.Listen` helper.

## Design

- Single hook point: `database/db.Logger.Trace` — every query path (gorm, `Tx.Select`/`exec`, `Query.trace`, deferred transaction logs) already funnels through it, so no driver changes are needed.
- Listeners live in a global registry (`database/utils/query_event.go`): a listener on any connection sees all queries, like Laravel.
- `Sql`/`Bindings` for gorm queries are captured via a lightweight `QueryEventPlugin` registered in `BuildGorm` (stashes `stmt.SQL`/`stmt.Vars` on the statement context). Direct-SQL paths attach bindings in `Query.trace`/`Tx.Select`/`Tx.exec`.
- Zero overhead when no listener is registered (`HasQueryListeners()` guard); existing mock-logger tests pass unchanged.
- Panicking listeners are recovered; events fire even at `Silent` log level (Laravel parity).

Strictly additive — no existing signature changes.

## Testing

- `go build ./...`, `go vet ./...` clean
- `go test ./database/... -count=1` — 792 passed in 13 packages
- `go test ./... -count=1` — all packages ok except two pre-existing environmental failures on this machine (`foundation` asserts the clone dir is named `framework`; a `process` Windows timing flake) — both fail identically on the clean baseline.
- New `database/db/listen_test.go`: listener receives correct fields on select/insert/transaction paths; multiple listeners; panicking listener recovered; no-listener fast path.

## Open questions

1. Should `Listen` route through the event module (`facades.Event().Dispatch`) instead of a dedicated registry? Current design keeps `database` decoupled from `event`.
2. Laravel traces `BEGIN`/`COMMIT` themselves; here transaction statements fire on `Commit` (deferred), matching the existing query-log behavior.
3. No `Unlisten` API added — easy follow-up if wanted.

### DIFF
--- a/contracts/database/db/db.go
+++ b/contracts/database/db/db.go
@@ -13,6 +13,10 @@ type DB interface {
 	BeginTransaction() (Tx, error)
 	// Connection gets an Orm instance from the connection pool.
 	Connection(name string) DB
+	// Listen registers a listener that is called for every SQL query executed
+	// by the database, mirroring Laravel's DB::listen. The registry is global,
+	// so a listener sees queries from every connection.
+	Listen(listener func(event *QueryExecuted) error)
 	// Transaction runs a callback wrapped in a database transaction.
 	Transaction(txFunc func(tx Tx) error) error
 	// WithContext sets the context to be used by the Orm.
--- a/contracts/database/db/events.go
+++ b/contracts/database/db/events.go
@@ -1,0 +1,24 @@
+package db
+
+import (
+	"time"
+)
+
+// QueryExecuted is the event fired after every SQL query executed by the
+// database, mirroring Laravel's Illuminate\Database\Events\QueryExecuted.
+// Listeners are registered through Listen on the DB facade or the package
+// level, and receive one event per executed query.
+type QueryExecuted struct {
+	// Connection is the name of the connection the query was executed on.
+	Connection string
+	// Sql is the SQL query with placeholders.
+	Sql string
+	// Bindings are the values bound to the placeholders.
+	Bindings []any
+	// RawSql is the SQL query with the bindings interpolated.
+	RawSql string
+	// Time is how long the query took to execute.
+	Time time.Duration
+	// Error is the error returned by the query, nil when the query succeeded.
+	Error error
+}
--- a/database/db/db.go
+++ b/database/db/db.go
@@ -17,6 +17,7 @@ import (
 	"github.com/goravel/framework/contracts/log"
 	contractstelemetry "github.com/goravel/framework/contracts/telemetry"
 	databasedriver "github.com/goravel/framework/database/driver"
+	"github.com/goravel/framework/database/utils"
 	"github.com/goravel/framework/errors"
 	"github.com/goravel/framework/support/carbon"
 	instrumentationdatabase "github.com/goravel/framework/telemetry/instrumentation/database"
@@ -65,7 +66,7 @@ func BuildDB(ctx context.Context, config config.Config, log log.Log, connection 
 	}
 
 	pool := driver.Pool()
-	logger := NewLogger(config, log)
+	logger := NewLogger(config, log).WithConnection(connection)
 	gorm, instrument, err := databasedriver.BuildGorm(config, logger.ToGorm(), pool, connection, telemetryResolver)
 	if err != nil {
 		return nil, err
@@ -234,13 +235,19 @@ func (r *Tx) Select(dest any, sql string, args ...any) error {
 	}
 
 	realSql = builder.Explain(sql, args...)
+	// Carry the placeholders and bindings on the context so the logger can
+	// report them in QueryExecuted events.
+	traceCtx := r.ctx
+	if utils.HasQueryListeners() {
+		traceCtx = utils.WithQueryBindings(traceCtx, sql, args)
+	}
 
 	destValue := reflect.Indirect(reflect.ValueOf(dest))
 
 	rowsAffected := int64(1)
 	if destValue.Kind() == reflect.Slice {
 		if err = builder.SelectContext(r.ctx, dest, sql, args...); err != nil {
-			r.logger.Trace(r.ctx, carbon.Now(), realSql, -1, err)
+			r.logger.Trace(traceCtx, carbon.Now(), realSql, -1, err)
 
 			return err
 		}
@@ -248,13 +255,13 @@ func (r *Tx) Select(dest any, sql string, args ...any) error {
 		rowsAffected = int64(destValue.Len())
 	} else {
 		if err = builder.GetContext(r.ctx, dest, sql, args...); err != nil {
-			r.logger.Trace(r.ctx, carbon.Now(), realSql, -1, err)
+			r.logger.Trace(traceCtx, carbon.Now(), realSql, -1, err)
 
 			return err
 		}
 	}
 
-	r.logger.Trace(r.ctx, carbon.Now(), realSql, rowsAffected, nil)
+	r.logger.Trace(traceCtx, carbon.Now(), realSql, rowsAffected, nil)
 
 	return nil
 }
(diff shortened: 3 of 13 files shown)


goravel-coder:
# 🤖 Automated review

## Summary
PR adds a Laravel-style `DB::listen` (`QueryExecuted`) API via a global listener registry hooked into `Logger.Trace`, with a gorm plugin capturing SQL/bindings. The mechanism is sound (callback ordering, connection propagation, and no-listener fast path were all verified), but the "zero overhead when no listener" claim is not fully met, the `Time` field is wrong on two code paths, and the public facade path lacks integration coverage.

## Verdict
- Must Fix: 2 · Should Fix: 8 · Nits: 9

## Findings

### Must Fix
- [ ] `database/db/logger.go:98` **`Trace` computes `duration`/`elapsed` before the Silent early-return, allocating on every query even with no listeners** — `begin.DiffInDuration()` calls `carbon.Now()` → `CreateFromStdTime` (heap-allocates a `*Carbon` + `time.Now()` + tz parse) on every query, and it now runs *unconditionally* at the top of `Trace`. Previously the `r.level <= logger.Silent` guard returned before touching duration; now the common production case (Silent logger, no listener) pays the cost the PR explicitly set out to avoid. It also reintroduces a latent nil-`begin` panic risk where none existed. Suggestion: hoist `hasListeners := utils.HasQueryListeners()` once; `if r.level <= logger.Silent && !hasListeners { return }` first, then compute `duration`/`elapsed`, dispatch the event (guarded by `hasListeners`), then keep the existing `if r.level <= logger.Silent { return }` before the logging switch.
- [ ] `tests/db_test.go` **No integration test exercises `facades.DB().Listen()` end-to-end** — the PR only adds unit tests (`database/db/listen_test.go`, `database/utils/query_event_test.go`); there is no test in the `tests/` integration module that registers a listener through the public facade and asserts it fires against a real connection. This is the user-facing entry point, so unit tests alone don't guarantee the facade → registry → logger wiring works on a live query path. Suggestion: add a `tests/db_test.go` case that calls `facades.DB().Listen(...)` (via `query.DB()`), runs a real select/insert/transaction, asserts `Connection`/`Sql`/`Bindings`/`RawSql`/`Error` are populated, and clears the global registry in cleanup.

### Should Fix
- [ ] `database/db/db.go:250` **`Tx.Select`/`Tx.exec` report a near-zero `QueryExecuted.Time`** — these paths pass `carbon.Now()` at trace time (lines 250, 258, 264, 328, 334, 338), so `begin.DiffInDuration()` is microseconds, not the query duration. The gorm path (`curTime` in `Execute`) and `Query.trace` (captures `now` *before* `GetContext`/`ExecContext`) both report real durations. Suggestion: capture `now := carbon.Now()` before `builder.SelectContext/GetContext/ExecContext` and pass `now` to `Trace`.
- [ ] `database/db/query.go:1400` **Deferred transaction events report an inflated `Time`** — `TxLog.begin` stores the query start time, but `Tx.Commit` (`database/db/db.go:199`) calls `Trace(log.ctx, log.begin, ...)`, so `begin.DiffInDuration()` at commit time includes the entire idle tail of the transaction, not the query's own duration. Suggestion: compute and store the elapsed duration in `TxLog` at trace time (right after the query) and pass it through, rather than recomputing at commit.
- [ ] `database/utils/query_event.go:33` **`HasQueryListeners()` takes a `sync.RWMutex.RLock` on every query** — this runs on the hottest path and is called up to twice per query (once in `Query.trace`/`Tx.Select`/`Tx.exec`/plugin `after`, again in `Logger.Trace`), costing a reader-count RMW + lock-state load per call. Suggestion: replace the mutex-guarded slice with an `atomic.Int32`/`atomic.Bool` listener count so `has()` is a single atomic load, or cache the result per query and pass it down.
- [ ] `database/utils/query_event.go:53` **`recover()` silently swallows every listener panic** — the deferred `_ = recover()` discards *all* panics, including `runtime.Error` (nil deref, index out of range, concurrent-map-write) that leave listener state corrupted with zero visibility. Suggestion: log the recovered value with a stack trace before swallowing, or only recover non-runtime panics.
- [ ] `database/utils/query_event.go:57` **The listener `error` return is dead weight** — `_ = listener(event)` drops the error and panics are recovered at lines 53–55, so a listener can never signal failure and failures are invisible. Suggestion: either drop the `error` from the signature (matching Laravel's void closure) or surface it (log it, or route through `facades.Event().Dispatch` which aggregates errors).
- [ ] `database/utils/query_event.go:94` **`WithQueryBindings` stores the caller's `args` slice without copying** — direct-SQL call sites pass the caller's variadic slice straight through, so a listener mutating `event.Bindings` corrupts the caller's args (and vice-versa); this is inconsistent with `after` (`query_event.go:156`) which copies via `append([]any{}, vars...)`. Suggestion: copy the slice inside `WithQueryBindings` (`append([]any{}, args...)`) so all paths are protected uniformly.
- [ ] `database/db/listen_test.go:51` **`collectEvents` returns the live internal `events` slice after releasing the mutex** — the closure returns the backing slice directly rather than a copy, defeating the defensive mutex the moment the lock is released. Suggestion: `return append([]*contractsdb.QueryExecuted{}, events...)`.
- [ ] `database/db/listen_test.go:293` **Direct `Query.trace` error path is untested with a registered listener** — only the success path is exercised; `QueryExecuted.Error` and `RawSql` (`builder.Explain` output) on a non-transaction failure are never asserted. Suggestion: add a case where `SelectContext`/`ExecContext` returns an error and assert `Error` and `RawSql`.

### Nits
- [ ] `database/db/logger.go:51` **`WithConnection` mutates the shared `*Logger` receiver in place** — `Trace` (line 113) reads `r.connection` from query goroutines while `WithConnection` writes it with no synchronization; safe only because it's called once at build time. Suggestion: pass `connection` into `NewLogger` as an immutable field instead of a mutating setter.
- [ ] `database/db/logger.go:51` **`WithConnection` returns `*Logger` while sibling fluent `Level` (line 61) returns `logger.Logger`** — inconsistent return types break fluent chaining (`NewLogger(...).WithConnection(c).Level(x)` stops exposing `WithConnection`). Suggestion: make both fluent methods return the same type.
- [ ] `database/db/logger.go:25` **`NewLogger` return type changed from `logger.Logger` interface to concrete `*Logger`** — source-compatible for most callers but breaks binary compatibility and removes the ability to return a mock/nil; the PR's "no existing signature changes" claim is not strictly true. Suggestion: keep returning `logger.Logger` (with `WithConnection` on the interface) or document it as intentional.
- [ ] `database/db/query.go:1402` **Redundant second `HasQueryListeners()` check** — `Query.trace` gates `WithQueryBindings` on `HasQueryListeners()`, then `Logger.Trace` re-checks it and re-reads the context value it just set (same pattern in `Tx.Select`/`Tx.exec`). Suggestion: pass the precomputed flag/wrapped ctx into `Trace` so only one check runs per query.
- [ ] `database/db/query.go:1402` **`context` used as a parameter-passing channel** — `WithQueryBindings` smuggles SQL/bindings through `context.WithValue` between the query/plugin and the logger, violating context semantics (request-scoped values, not function parameters). Suggestion: thread a small typed trace payload explicitly instead, or at least document the pattern.
- [ ] `database/utils/query_event.go:156` **`append([]any{}, vars...)` is only a shallow copy** — element values (e.g. `[]byte` blob/JSON bindings, pointers) are still shared; a caller mutating such a binding after the query corrupts the emitted event. Suggestion: deep-copy mutable bindings or document capture-by-reference.
- [ ] `database/db/listen_test.go:109` **Panicking-listener test only asserts `NotPanics`** — unlike the utils test, it never verifies a later listener still runs after a panicking one at the `logger.Trace` level. Suggestion: register a second non-panicking listener and assert it still receives the event.
- [ ] `database/db/listen_test.go:245` **`Tx.Update` and `Tx.Statement` event emission is never tested** — only `Insert` and `Delete` (both via `Tx.exec`) are covered. Suggestion: add a case for the shared exec path.
- [ ] `database/utils/query_event_test.go:22` **`collectQueryEvents` appends without a mutex** — inconsistent with `listen_test.go`'s `collectEvents`; safe only because dispatch is synchronous. Suggestion: add a mutex for parity or a comment noting the synchronous-dispatch invariant.

## Automated Checks
- [ ] `go vet ./database/...` — clean
- [ ] `go test ./database/db/... ./database/utils/... -count=1` — pass (`ok` for both packages)

PR #1558, opened 2026-09-07
feat: [#880] make all make:* commands accept multiple names

### LINKED ISSUE #880: Make commands should accept multiple arguments
### Before feedback (在反馈之前)

### Describe Feature (描述功能)

Currently, each `make:*` command accepts only a single name. For example, `make:controller` allows creating only one controller at a time. If we need to create multiple controllers, we have to run the command multiple times.

It would be more convenient if the command could accept multiple names in a single execution, similar to how Linux commands like `mkdir` work.

**Example:**

```bash
./artisan make:controller AuthController TaskController
./artisan make:model User Task
```

This would improve DX and reduce repetitive command execution.

wafik:
## Problem

Follow-up to #1498, which converted only 4 of the 24 `make:*` commands. Maintainer asked to "cover other commands to implement this feature in this PR or another" — this is the other PR. Closes goravel/goravel#880.

## Approach

Applies the exact pattern merged in #1498 to the remaining 20 commands (agent, tool, policy, channel, factory, model, observer, seeder, migration, event, listener, package, provider, test, mail, notification, job, filter, rule, view):

- `Handle` loops over `supportconsole.MakeNames(ctx)` and calls a new `makeOne(ctx, name)`.
- Per-name failures are reported via `ctx.Error` and the loop continues.
- `make:package` / `make:view` keep their internal empty-name prompt/check inside `makeOne`.
- Single-name and no-arg interactive behavior unchanged. No contract changes, no mock regeneration needed.

## Testing

- `go build ./...` clean
- `go test` on all 16 touched packages — 590 passed
- Each command gained a `*_MultipleNames` test asserting multiple files are generated; `make:model` also covers partial failure.

### DIFF
--- a/ai/console/agent_make_command.go
+++ b/ai/console/agent_make_command.go
@@ -39,15 +39,23 @@ func (r *AgentMakeCommand) Extend() command.Extend {
 
 // Handle Execute the console command.
 func (r *AgentMakeCommand) Handle(ctx console.Context) error {
-	make, err := supportconsole.NewMake(ctx, "agent", ctx.Argument(0), support.Config.Paths.Agents)
+	for _, name := range supportconsole.MakeNames(ctx) {
+		if err := r.makeOne(ctx, name); err != nil {
+			ctx.Error(err.Error())
+		}
+	}
+
+	return nil
+}
+
+func (r *AgentMakeCommand) makeOne(ctx console.Context, name string) error {
+	make, err := supportconsole.NewMake(ctx, "agent", name, support.Config.Paths.Agents)
 	if err != nil {
-		ctx.Error(err.Error())
-		return nil
+		return err
 	}
 
 	if err := file.PutContent(make.GetFilePath(), r.populateStub(r.getStub(), make.GetPackageName(), make.GetStructName())); err != nil {
-		ctx.Error(err.Error())
-		return nil
+		return err
 	}
 
 	ctx.Success("Agent created successfully")
--- a/ai/console/agent_make_command_test.go
+++ b/ai/console/agent_make_command_test.go
@@ -38,23 +38,23 @@ func TestAgentMakeCommand(t *testing.T) {
 		}
 	}
 
-	mockContext.EXPECT().Argument(0).Return("").Once()
+	mockContext.EXPECT().Arguments().Return([]string{""}).Once()
 	mockContext.EXPECT().Ask("Enter the agent name", mock.Anything).Return("", errors.New("the agent name cannot be empty")).Once()
 	mockContext.EXPECT().Error("the agent name cannot be empty").Once()
 	assert.NoError(t, agentMakeCommand.Handle(mockContext))
 
-	mockContext.EXPECT().Argument(0).Return("UserAgent").Once()
+	mockContext.EXPECT().Arguments().Return([]string{"UserAgent"}).Once()
 	mockContext.EXPECT().OptionBool("force").Return(false).Once()
 	mockContext.EXPECT().Success("Agent created successfully").Once()
 	assert.NoError(t, agentMakeCommand.Handle(mockContext))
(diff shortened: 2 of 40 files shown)


goravel-coder:
# 🤖 Automated review

## Summary
PR #1558 faithfully applies the #1498 multi-name pattern (`Handle` → loop over `MakeNames` → `makeOne`) to the remaining 20 `make:*` commands. It is a clean, consistent, well-tested refactor with no blocking bugs — `go build` and `go vet` pass, and the conversion preserves all per-command validation (force flag, migration registration, package/view empty-name handling). The remaining concerns are an efficiency regression for `make:model --table` with multiple names, and a few test-hygiene gaps in the new `*_MultipleNames` tests.

## Verdict
- Must Fix: 0 · Should Fix: 3 · Nits: 4

## Findings

### Should Fix
- [ ] `database/console/model_make_command.go:97` **Repeated DB schema round-trips when `--table` is set** — `makeOne` calls `schema.HasTable(table)`, `schema.GetColumns(table)`, and `generateModelInfo(...)` per name, so `make:model User Post Comment --table=users` issues 3× the queries for identical schema data. Suggestion: when `--table` is set, hoist the schema fetch/model-definition out of the loop and pass it into `makeOne`.
- [ ] `database/console/model_make_command_test.go:587` **Non-deferred cleanup leaks files on assertion failure** — teardown is `assert.Nil(t, file.Remove("app"))` at the end, so an earlier failing `file.Exists` assertion leaves generated `app/models/*.go` in the source tree. Same pattern in `auth/console/policy_make_command_test.go:55`, `database/console/observer_make_command_test.go:56`, `database/console/seeder_make_command_test.go:180`, `event/console/event_make_command_test.go:89`, `foundation/console/provider_make_command_test.go:131`, `foundation/console/test_make_command_test.go:115`, `view/console/view_make_command_test.go:127`. Suggestion: use `t.Cleanup`/`defer` uniformly (agent/tool/channel/job/filter/rule tests already do).
- [ ] `database/console/model_make_command_test.go:565` **Partial-failure coverage only exists for make:model** — the "continue after an error" loop is identical across all commands, but only model tests it, and only via the "file already exists" path. Commands with a post-create registration step (job, filter, rule, seeder, provider, migration) have no test proving a mid-loop registration error still processes the remaining names. Suggestion: add one partial-failure test for a registration-step command (e.g. filter/rule).

### Nits
- [ ] `support/console/console.go:28` **24× copy-pasted loop-and-report block** — the `for ... MakeNames { if err := makeOne(...); err != nil { ctx.Error(...) } }` block is now duplicated across all 24 commands, so future changes to continuation/error semantics touch 24 places. Suggestion: add `supportconsole.HandleEach(ctx, fn)` next to `MakeNames` and have each `Handle` be a one-liner (retrofitting the 4 #1498 commands too).
- [ ] `foundation/console/provider_make_command_test.go:116` **gofmt trailing whitespace** — `gofmt -l` flags trailing tabs on blank lines. Pre-existing, not introduced by this PR, but the file is touched here so it's a cheap cleanup. Suggestion: `gofmt -w`.
- [ ] `view/console/view_make_command_test.go:117` **Inconsistent multi-name test naming** — `TestMultipleNames` vs `TestHandleMultipleNames` (provider/package) vs `TestXxxMakeCommand_MultipleNames` (non-suite files). Suggestion: standardize on one convention.
- [ ] `database/console/model_make_command_test.go:558` **Weak failure message** — `assert.True(t, file.Exists(...))` reports only "Should be true" and can't distinguish which of `user.go`/`post.go` is missing. Suggestion: use `assert.FileExists(t, "app/models/user.go")`.

## Automated Checks
- [ ] `gofmt -l` — flags `foundation/console/provider_make_command_test.go` (pre-existing trailing whitespace; all other changed files clean)
- [ ] `go vet ./...` (changed packages) — clean

PR #1562, opened 2026-09-19
feat: [#972] add When to the ORM query

### LINKED ISSUE #972: [Feature] Add Laravel-style When method for conditional ORM/database queries
### Before feedback (在反馈之前)

## 描述
- 希望能在数据操作中添加`when`方法

## 使用示例
```go
var users []models.User

err := facades.Orm().Query().
    When(keyword != "", func(query orm.Query) orm.Query {
        return query.Where("name", "like", "%"+keyword+"%")
    }).
    When(status != "", func(query orm.Query) orm.Query {
        return query.Where("status", status)
    }).
    When(startAt != "", func(query orm.Query) orm.Query {
        return query.Where("created_at", ">=", startAt)
    }).
    Find(&users)
```

darakanoit:
## 📑 Description

Closes https://github.com/goravel/goravel/issues/972

`db.Query` already has `When`; this adds the same method to `orm.Query`, so a query can be built conditionally without breaking the chain:

```go
err := facades.Orm().Query().
	When(keyword != "", func(query orm.Query) orm.Query {
		return query.Where("name LIKE ?", "%"+keyword+"%")
	}).
	When(sortByVotes, func(query orm.Query) orm.Query {
		return query.OrderByDesc("votes")
	}, func(query orm.Query) orm.Query {
		return query.OrderBy("name")
	}).
	Find(&users)
```

The signature and behaviour match `db.Query.When`: the callback runs when the condition is true, otherwise the optional second callback runs, and without one the query is returned unchanged. Unlike Laravel, the condition is a `bool` and isn't passed to the callback, since the closure can already use the value it checks.

The integration test covers the three cases on each driver.

## ✅ Checks


hwbrzzl (approved):
Thanks, LGTM

PR #1564, opened 2026-09-21
feat: [#1006] support multiple queues in the worker

### LINKED ISSUE #1006: ✨ [Feature] Support multiple queue names in queue workers
### Before feedback

- [x] There are no existing Issues or PRs covering multiple queue names for the queue worker.
- [x] The proposed behavior is based on Laravel's documented queue priorities.

### Describe Feature

Goravel queue workers currently consume one queue name at a time. Please add Laravel-compatible support for consuming multiple queue names in priority order.

Laravel supports:

```shell
php artisan queue:work --queue=high,default
```

The worker checks queues from left to right. It only checks `default` after `high` is empty. If `high` remains busy, `default` may be starved; this is the intended priority semantics.

For Goravel, the proposed API is:

```go
facades.Queue().Worker(queue.Args{
	Connection: "redis",
	Queue:      "high,default",
	Concurrent: 2,
	Tries:      3,
})
```

The existing `queue.Args.Queue` field remains a string and accepts a comma-separated list. `OnQueue` continues to dispatch a job to one queue.

The default queue runner should optionally support a separate worker-only configuration key so the dispatch default remains a single queue:

```go
"connections": map[string]any{
	"redis": map[string]any{
		"driver":     "custom",
		"queue":      "default",
		"queues":     []string{"high", "default"},
		"concurrent": 2,
	},
},
```

The implementation should keep `Driver.Pop(queue string)` and `DriverWithReceive` compatible. The worker can call `Pop` in priority order and pass the actual queue name through failure handling so `failed_jobs.queue` remains correct. For a multi-queue worker using `DriverWithReceive`, the first version can fall back to `Pop`; batch receiving can remain available for single-queue workers.

Expected coverage:

- high-priority jobs are selected before lower-priority jobs;
- an empty high-priority queue falls through to the next queue;
- driver errors do not silently consume lower-priority queues;
- failed jobs record the queue from which they were actually reserved;
- existing single-queue and batch-receive behavior remains unchanged.

### Implementation question

Please confirm whether the maintainers prefer the Laravel-compatible `Args.Queue` comma-separated string proposed above, or a new Go-specific `Args.Queues []string` field. The comma-separated form keeps the public API small and matches Laravel; the slice avoids ambiguity for queue names containing commas.

Reference:

- https://laravel.com/docs/12.x/queues#queue-priorities
- https://github.com/laravel/framework/blob/12.x/src/Illuminate/Queue/Worker.php

tinpont:
## Summary
- `queue.connections.<connection>.queue` now accepts several queue names (`"high,default"`), and the worker consumes them in that order.
- A job is taken from the first queue that has one, so an empty high-priority queue falls through to the next; driver errors stop the round instead of draining lower-priority queues.
- Failed jobs are recorded against the queue they were actually consumed from, keeping `queue:retry` and the failed-jobs table accurate.

Closes https://github.com/goravel/goravel/issues/1006

## Why
A connection could only name one queue, so keeping high-priority jobs from waiting behind low-priority ones meant running one worker process per queue. Listing the queues on the connection gives that ordering to a single worker.

```go
// config/queue.go
"database": map[string]any{
	"driver":     "database",
	"connection": "sqlite",
	"queue":      "high,default",
	"concurrent": 5,
},
```

The worker pops from the first queue that returns a job and only moves on when the driver reports no job found, so `high` jobs are handled before `default` ones. Drivers implementing `Receive` keep receiving in batches while one queue is configured and switch to `Pop` once several are listed. Multi-queue workers also pass the queue a job was consumed from into failure handling, so a job that fails on `high` is retried on `high`.


tinpont:
Review Ready

goravel-coder:
# 🤖 Automated review

## Summary
The worker now splits a comma-separated `queue.connections.<conn>.queue` value and pops queues in priority order, threading the consumed queue name into failure handling. The worker-side mechanics are sound and well covered, but the feature changes the meaning of a shared config key without updating the dispatch path, so enabling it silently strands every default-dispatched job.

## Verdict
- Must Fix: 1 · Should Fix: 7 · Nits: 9

## Findings

### Must Fix
1. [x] `queue/pending_job.go:38` **Multi-queue config strands default-dispatched jobs** — `Config.DefaultQueue()` (`queue/config.go:23`) still returns the raw config string, so with `"queue": "high,default"` a job dispatched without `OnQueue` is pushed via `driver.Push(r.task, r.queue)` (`pending_job.go:112`, `:121`) to a queue literally named `"high,default"`, while the worker only ever pops `"high"` and `"default"`; such jobs are never consumed (and `queue:retry` re-pushes them to the same unconsumable name). Suggestion: keep the dispatch default single-valued — split/trim in `queue.Config` and expose a worker consumption list separately (e.g. `DefaultQueues() []string`), and add an integration test asserting a default-dispatched job is popped.

### Should Fix
1. [ ] `queue/worker.go:446` **N `Pop` calls per poll round** — each round calls `driver.Pop` once per configured queue; the database driver takes a per-queue cache lock (`driver_database.go:66`) plus a `LockForUpdate` transaction per call, so draining a backlog from a lower-priority queue pays an extra lock + transaction per job and idle polling scales linearly with queue count. Suggestion: add a driver-level multi-queue pop (e.g. `PopMany([]string)` / `WHERE queue IN (...)`), or a short-TTL known-empty-queue cache, and benchmark 1 queue vs N.
2. [ ] `queue/worker.go:246` **`DriverWithReceive` + multi-queue silently degrades to `Pop`** — gating on `len(queueNames) == 1` forces every receive-based driver (Kafka-like, whose `Pop` may be a stub) onto `runWithPop`; with more than one queue it can spin logging failures without ever consuming and with no hint that the config caused it. Suggestion: extend `Receive` to accept a queue list, or reject/log explicitly when a `DriverWithReceive` driver is configured with multiple queues.
3. [x] `queue/worker.go:451` **A non-`NoJobFound` error on a higher-priority queue starves lower queues** — `pop` returns immediately on any error, and the database driver returns `QueuePopIsLocked` on lock contention (`driver_database.go:67`), so a contended `high` queue stops the worker from draining `default` and the `currentDelay *= 2` path can sleep up to 32s. Suggestion: treat lock-contention/transient errors as retryable and continue to the next queue, aborting only on fatal errors (or document the intentional tradeoff).
4. [x] `contracts/queue/config.go:11` **`DefaultQueue() string` contract and parsing layer not updated** — comma parsing now lives inside `Worker.queueNames()`, so the meaning of the config key is owned by one consumer while `PendingJob`, drivers, and `Application.Worker` see a different value. Suggestion: expose the parsed list from `queue.Config` (e.g. `Queues() []string`), have `NewWorker` accept `[]string`, and regenerate `mocks/queue`.
5. [x] `queue/console/queue_retry_command.go:45` **`queue:retry --queue` filtering not adapted to the list** — `Failer.Get` matches with an exact `Where("queue", queue)` (`queue/failer.go:43`), so users following the new comma syntax (`--queue=high,default`) match nothing now that failed jobs store a single resolved name. Suggestion: support the same comma list (`WhereIn`) or document that `--queue` takes one name.
6. [x] `queue/worker_multi_queue_test.go:160` **Failed-job queue attribution only asserted at the `call` seam** — `TestCallRecordsReservedQueue` invokes `r.call(task, nil, "high")` directly, so the `pop` → `processReservedJob(reservedQueue)` → `call` plumbing this PR changed is never exercised; a regression passing `r.queue` instead of the popped name would still pass. Suggestion: add a `run()` case where the job is popped from the lower-priority queue and assert the recorded `models.FailedJob.Queue`.
7. [x] `queue/worker.go:456` **`pop` panics on an empty queue slice** — `queueNames[len(queueNames)-1]` evaluates index `-1`; unreachable via `queueNames()` today, but `pop` is now a directly tested helper. Suggestion: guard `if len(queueNames) == 0 { return nil, "", errors.QueueDriverNoJobFound }`.

### Nits
1. [x] `queue/worker.go:456` **Bare `QueueDriverNoJobFound` returned without args** — the sentinel is `New("no job found in %s queue")` (`errors/list.go:283`) and the driver always supplies `.Args(queue)`; returning it bare leaves `%s` unexpanded if it is ever logged. Suggestion: return `errors.QueueDriverNoJobFound.Args(...)` or add a dedicated no-arg sentinel.
2. [x] `queue/worker.go:434` **All-whitespace config passed through as a queue name** — for `" , "` the method returns `[]string{" , "}` and the worker polls a nonsense queue; the comment "Nothing to split" is also inaccurate (the branch triggers only when every token is empty). Suggestion: fall back to `"default"` or return an error, and reword the comment.
3. [x] `queue/worker.go:425` **Queue names are trimmed only on the consume side** — `OnQueue`/`Push` do not normalize, so `" default "` targets a different row than `"default"`. Suggestion: normalize names consistently on both push and pop.
4. [x] `queue/worker.go:448` **`pop` treats `(nil, nil)` from a driver as success** — a nil `reservedJob` with a nil error makes `processReservedJob` panic on `reservedJob.Task()`; the new receive-to-`Pop` fallback widens the set of drivers reaching this path. Suggestion: treat a nil job as no-job and fall through to the next queue.
5. [x] `queue/worker.go:442` **Priority guarantee overstated in the doc comment** — with `concurrent > 1` each goroutine pops independently, so a lower-priority job can be reserved while a higher-priority one is pending. Suggestion: soften the comment or reserve through a single atomic driver operation.
6. [x] `queue/worker.go:360` **Naming: `reservedQueue` vs `queueName`** — `call`, `runWithReceive`, and `pop` all use `queueName` for the same concept; also the local `queueNames` at `worker.go:424` shadows the `queueNames()` method and `pop` is a vague name for a priority-list walk. Suggestion: use one name (`queueName`) and rename the local (`names`) and/or method (`popNextJob`).
7. [x] `queue/worker.go:234` **Debug log prints the raw comma list** — `errors.QueueProcessingJobs.Args(r.connection, r.queue)` logs `high,default` rather than the resolved names. Suggestion: hoist `queueNames()` and log the joined resolved list.
8. [x] `queue/setup/stubs.go:33` **Stub config doesn't document the new format** — the generated `queue` config still shows `"queue": "default"` with no mention of comma-separated consumption lists, so users won't discover the feature. Suggestion: update the stub once the contract is settled.
9. [x] `queue/worker_multi_queue_test.go:51` **Test hygiene gaps** — `TestQueueNames` mutates the shared `s.worker.queue` without `s.SetupTest()`, the `Receive` non-call assertion (`:130`/`:151`) is racy and made a dead no-op by `.Maybe()`, `TestRunWithMultipleQueuesPops` implicitly relies on the 1s backoff to avoid a second `Pop` (`:133`), and `TestCallRecordsReservedQueue` asserts only `Queue` rather than the full `models.FailedJob` (`:178`). Suggestion: reset state per subtest, fail the test inside the `Receive` callback, expect extra polls deterministically, and assert the whole failed-job struct.

## Automated Checks
- [x] `gofmt -d queue/worker.go queue/worker_multi_queue_test.go queue/worker_test.go` — clean
- [x] `go vet ./queue/...` — passed
- [x] `go test ./queue/ -run TestWorkerTestSuite -count=1` — passed
- [ ] `golangci-lint run ./queue/...` — not run (installed binary targets go1.26, module targets go1.27.0)

tinpont:
Thanks for the detailed review. I checked the findings against Laravel 12.x's queue worker implementation and addressed the actionable issues.

### Addressed

- Kept `queue.connections.<connection>.queue` single-valued for default dispatch. If a comma-separated value is supplied, the dispatch default resolves to the first non-empty queue, preventing jobs from being pushed to a literal `high,default` queue. Multiple queues remain available through an explicit worker argument such as `queue.Args{Queue: "high,default"}`.
- Added defensive handling for an empty queue list and for drivers returning `(nil, nil)`.
- Returned `QueueDriverNoJobFound` with the resolved queue name instead of leaving the `%s` placeholder unexpanded.
- Made empty or whitespace-only worker queue values fall back to `default`.
- Replaced the direct `call` test with coverage of the full `runWithPop -> pop -> processReservedJob -> call` path, verifying that a failed job records the queue from which it was reserved.
- Removed the ineffective optional `Receive` expectation and reset worker state in queue parsing subtests.
- Softened the priority comment so it does not imply a global ordering guarantee across concurrent workers.

### Not changed

- The worker still calls `Pop` once per queue. Laravel's `Worker::getNextJob` follows the same left-to-right polling model, so a driver-level multi-queue operation is an optional performance enhancement rather than a correctness requirement.
- A driver error still stops the current polling round. Laravel wraps the queue iteration in one exception boundary and also stops checking lower-priority queues after a pop error.
- Multi-queue `DriverWithReceive` workers still fall back to `Pop`, as explicitly allowed by the originating issue.
- `queue:retry --queue` remains a single-queue filter, matching Laravel's documented behavior.

Validation:
- `go test ./queue -count=1`
- focused race tests for the changed worker paths
- `go vet ./queue/...`
- `git diff --check`

tinpont:
I amended the PR commit as `97f04d4e` after reviewing the automated findings against Laravel's queue internals.

Key changes:
- Keep the dispatch target single-valued by normalizing the configured queue to the first valid queue name. The worker still reads the raw configured value and consumes comma-separated queues in priority order, so default-dispatched jobs are not stranded on a literal `high,default` queue.
- Preserve the actual reserved queue when recording failed jobs.
- Guard empty queue lists, `(nil, nil)` driver results, and no-job errors with the correct queue argument.
- Keep the existing single-queue `Receive` path and multi-queue `Pop` fallback without expanding the public driver contract.

This matches Laravel's worker behavior: the worker splits the queue list and checks queues left-to-right, while dispatch uses one queue target. I also kept driver-error handling unchanged because Laravel stops the current pop pass when a driver exception occurs.

goravel-coder:
# 🤖 Automated review


## Summary
The author's round-1 fix is correct and Laravel-compatible: `Config.DefaultQueue()` normalizes the connection's `queue` value to a single dispatch target while the worker reads the raw value and polls the comma list left-to-right, so default-dispatched jobs are consumed from the first queue instead of being stranded (which is what Laravel would do with a list in that key). The blocking correctness issue is resolved; what remains are three still-open round-1 items and documentation/test hygiene nits.

## Verdict
- [NEW] Must Fix: 0 · Should Fix: 0 · Nits: 11
- [ROUND 1](https://github.com/goravel/framework/pull/1564#issuecomment-5772921268): Must Fix: 0 · Should Fix: 3 · Nits: 5 · Resolved: 9

## Findings

### Nits
1. [x] `contracts/queue/driver.go:26` **Stale `DriverWithReceive` doc** — the comment says "the Worker uses Receive instead of Pop", which is now true only when exactly one queue is configured. Suggestion: state the single-queue condition in the interface comment.
2. [x] `queue/config.go:24` **First-name dispatch target is undocumented** — with `"queue": "high,low"`, `DefaultQueue()` returns `high`, so default `Dispatch()` jobs go to the priority lane (Laravel would use the raw `"high,low"` and strand them). It's a deliberate, safer choice, but users may expect the catch-all. Suggestion: document that the first valid name is the dispatch default.
3. [x] `queue/pending_job.go:131` **`OnQueue` single-name semantics undocumented** — Laravel's `onQueue` also takes one name and never splits, so `OnQueue("high,low")` strands the job; here the behavior is identical but undocumented while the config/`queue.Args` path accepts lists. Suggestion: document it takes a single queue name (or reject commas).
4. [x] `queue/application.go:71` **`Worker()` re-derives the queue config key and re-parses it inline, with an inconsistent default** — the `fmt.Sprintf("queue.connections.%s.queue", …)` fallback is repeated here and at `:83` (which still falls back to the literal `"default"`), so dispatch (`DefaultQueue()`) and consumption can drift; the raw-config path is also untested. Suggestion: add a `Config.QueueNames(connection string) []string` accessor and use it in both branches, with a test asserting the raw list reaches `NewWorker`.
5. [x] `queue/worker.go:430` **Unreachable empty-list guard emits a malformed error** — `splitQueueNames` never returns empty, so the `len(queueNames) == 0` branch is dead in production and `errors.QueueDriverNoJobFound.Args("")` renders `no job found in  queue`. Suggestion: drop the branch (the invariant lives in `splitQueueNames`) or pass a meaningful placeholder.
6. [x] `queue/config.go:75` **`splitQueueNames` lacks a doc comment and de-duplication** — the other new helpers are documented, and `"high,high"` yields a duplicate entry (and a duplicate `Pop` per round). Suggestion: add a comment and skip already-seen names.
7. [ ] `queue/worker.go:448` **`pop`'s returned queue name is undocumented, misnamed, and re-allocates the sentinel each empty round** — the `string` result is the reserved queue, but the all-empty branch returns an arbitrary `lastQueue` whose error is swallowed upstream. Suggestion: document the return, name it `lastCheckedQueue` (or return the first name), and return the driver's last error.
8. [x] `queue/worker.go:422` **`queueNames()` is a pure one-line wrapper** over `splitQueueNames(r.queue)`, adding indirection that duplicates the package helper. Suggestion: call `splitQueueNames(r.queue)` directly, or justify the wrapper in its comment.
9. [x] `queue/worker_multi_queue_test.go:174` **Test hygiene gaps** — the negative `Receive` assertion is a single non-blocking `select` (use `AssertNotCalled` after `Shutdown()`), `<-runErrChan` (`:181`) has no timeout so a stuck `run()` hangs the suite, `TestCallRecordsReservedQueue` (`:202`) reads the raw channel instead of the persisted record, and the `pop(nil)` case (`:100`) asserts a production-unreachable branch. Suggestion: make these deterministic or drop the dead-branch case.
10. [x] `queue/config_test.go:46` **`TestConfiguredQueueListKeepsDispatchQueueSingle` overpromises and misses the fallback** — it only asserts `DefaultQueue() == "high"` and never covers `configuredQueue == ""`/`" , "` (the `splitQueueNames` → `"default"` path in `NewConfig`). Suggestion: rename to reflect `DefaultQueue()` normalization and add an empty/whitespace case.
11. [ ] `queue/worker_test.go:431` **Pre-existing `-race` failure in a touched test file** (not introduced here) — test goroutines call `s.NoError(...)` while testify's `SetT` runs on the test goroutine, so `go test -race ./queue/` fails. Suggestion: capture the error and assert on the main goroutine (as `TestRunWithMultipleQueuesPops` already does via a channel).

## Automated Checks
- [x] `gofmt -d queue/application.go queue/config.go queue/worker.go queue/config_test.go queue/worker_multi_queue_test.go queue/worker_test.go` — clean
- [x] `go vet ./queue/...` — passed
- [x] `go test ./queue/...` — passed
- [ ] `golangci-lint run ./queue/...` — not run (installed binary targets go1.26, module targets go1.27.0)

tinpont:
Thanks for the review. Commit 7d5cb95c clarifies queue semantics, centralizes config lookup, simplifies polling, and adds end-to-end failed-job coverage plus more deterministic tests.

I kept per-queue Pop and duplicate names to match Laravel's priority iteration. Multi-queue fallback to Pop is explicitly allowed by #1006 and preserves the driver contract. I did not add a public Config accessor because the existing contract is sufficient. The reported worker_test.go race predates this PR and is best handled separately.

goravel-coder:
# 🤖 Automated review
This is an AI-generated code review. Please double-check each finding before acting. You don't need to address every issue if they are inaccurate, but please point them out if any exist.


## Summary
This PR adds comma-separated multi-queue consumption to the worker (priority polling), keeps the dispatch default single-valued so default jobs are not stranded, and threads the reserved queue name into failure records. Round 2's blocking items and the design questions (per-queue `Pop`, `Receive`→`Pop` fallback, config-accessor) were verified as resolved or intentional and are closed; what remains is one lock-contention edge case and test/doc polish.

## Verdict
- [NEW] Must Fix: 0 · Should Fix: 1 · Nits: 6
- [ROUND 2](https://github.com/goravel/framework/pull/1564#issuecomment-5777890253) Must Fix: 0 · Should Fix: 0 · Nits: 0 · Resolved: 11
- [ROUND 1](https://github.com/goravel/framework/pull/1564#issuecomment-5772921268) Must Fix: 0 · Should Fix: 0 · Nits: 0 · Resolved: 17

## Findings

### Should Fix
1. [ ] `queue/worker.go:436` **Lock contention on a high queue stalls lower queues**
   **Problem:** `popNextJob` returns immediately on any error, and the database driver returns `QueuePopIsLocked` after blocking up to a minute on its per-queue lock.
   **Impact:** a contended `high` queue stops the worker from ever checking `default`, and the retry path can sleep up to 32s.
   **Fix:** continue to the next queue for this retryable lock error, aborting only on fatal driver errors.
   ```go
   // before
   if !errors.Is(err, errors.QueueDriverNoJobFound) {
       return nil, queueName, err
   }
   lastErr = err
   // after
   if !errors.Is(err, errors.QueueDriverNoJobFound) && !errors.Is(err, errors.QueuePopIsLocked) {
       return nil, queueName, err
   }
   lastErr = err
   ```

### Nits
1. [ ] `queue/worker.go:426` **Shutdown waits on every queue's pop**
   **Problem:** `popNextJob` loops over all queues calling `Pop` without checking `isShutdown` in between.
   **Impact:** a worker with several queues can take one blocking driver call per queue to stop instead of one.
   **Fix:** return as soon as shutdown is requested.
   ```go
   // before
   for _, queueName := range queueNames {
       reservedJob, err := r.driver.Pop(queueName)
   // after
   for _, queueName := range queueNames {
       if r.isShutdown.Load() {
           return nil, queueName, errors.QueueDriverNoJobFound.Args(queueName)
       }
       reservedJob, err := r.driver.Pop(queueName)
   ```

2. [ ] `contracts/queue/queue.go:33` **Public `Args.Queue` doc omits comma lists**
   **Problem:** the field is still documented as "Specify queue", but `Worker(Args{Queue: "high,default"})` now accepts a comma-separated priority list.
   **Impact:** callers cannot discover multi-queue consumption through the public `Args` API.
   **Fix:** document the comma-separated, left-to-right priority semantics on the field.

3. [ ] `queue/config.go:78` **Non-empty return guarantee is undocumented**
   **Problem:** `splitQueueNames` always returns at least one name, but its doc comment never states that.
   **Impact:** callers index `[0]` and `[len-1]` directly, so an edit dropping the fallback would panic at runtime.
   **Fix:** state in the comment that the result is never empty and falls back to `"default"`.

4. [ ] `queue/application_test.go:12` **Plain test breaks the package suite convention**
   **Problem:** this is the only plain `func TestXxx(t *testing.T)` in the package; every other queue test uses a `testify` suite.
   **Impact:** two competing test styles now coexist in `queue/`.
   **Fix:** fold the cases into a suite with `SetupTest`, matching the rest of the package.

5. [ ] `queue/worker_multi_queue_test.go:18` **Queue-splitting test lives in the wrong file**
   **Problem:** `TestSplitQueueNames` covers `splitQueueNames` from `queue/config.go` but sits in the worker test file.
   **Impact:** the pure helper's tests drift away from their `config_test.go` neighbours.
   **Fix:** move the test next to `TestNewConfigNormalizesDefaultQueue`.

6. [ ] `queue/worker_multi_queue_test.go:128` **Error asserted by pointer identity**
   **Problem:** `s.Same(lastErr, err)` requires the exact same error object rather than an equivalent one.
   **Impact:** wrapping the returned error later would break a still-correct behavior.
   **Fix:** assert with `s.ErrorIs`.
   ```go
   // before
   s.Same(lastErr, err)
   // after
   s.ErrorIs(err, errors.QueueDriverNoJobFound)
   ```

tinpont:
Addressed the six cleanup items in `234bb7e7`; all CI checks now pass.

I kept `QueuePopIsLocked` fatal for the current polling round because #1006 requires driver errors not to consume lower-priority queues, and this error may require manual lock cleanup.

